PR #1, opened 2015-02-22
Week 3 challenge finished!

SebastienPires:
Here's my code for reviews guys. Thank you!
If you want to check the app: http://awesome-rock-paper-scissors.herokuapp.com/


features/support/start_game.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Starting the game
```
tansaku: ideally this would not be in the support directory, but in the features directory one lower down

features/support/start_game.feature
```diff
@@ -0,0 +1,19 @@
...
+  In order to play RPS
+  As a student of MA
+  I want to start a new game
+
+  Scenario: Starting game
+    Given I'm on the homepage
+    When I follow "New Game"
+    Then I should see What's your name?
+
+    Scenario: Registering
```
tansaku: careful of your indenting here, all scenarios should be on the same level

features/step_definitions/start_game_step.rb
```diff
@@ -0,0 +1,11 @@
+Given(/^I'm on the homepage$/) do
+  visit '/'
+end
+
+When(/^I follow "(.*?)"$/) do |button|
+  click_on(button)
```
tansaku: ideally this should be 'click_link'

features/step_definitions/game_step.rb
```diff
@@ -0,0 +1,11 @@
+Given(/^that I am on the game page$/) do
+  visit '/game'
+end
+
+When(/^I click on the "(.*?)" button$/) do |arg1|
+  click_on(arg1)
```
tansaku: ideally this should be click_button

features/support/start_game.feature
```diff
@@ -0,0 +1,19 @@
...
+      Given I am on the name page
+      When I enter "My name"
+      Then I should see the game
+
+      Scenario: Playing game
+      Given that I am on the game page
+      When I click on the "rock" button
+      Then I should see the result page
```
tansaku: overall good set of high level tests here

features/step_definitions/start_game_step.rb
```diff
@@ -0,0 +1,11 @@
...
+end
+
+Then(/^I should see What's your name\?$/) do
+  body.should match "What's your name?"
```
tansaku: this is old rspec syntax (version 2) - ideally you should prefer version 3, e.g. expect(page).to have_content "what's your name"

features/step_definitions/game_step.rb
```diff
@@ -0,0 +1,11 @@
...
+end
+
+Then(/^I should see the result page$/) do
+  body.should match "Sweet, here are the results:"
```
tansaku: this is old rspec syntax (version 2) - ideally you should prefer version 3, e.g. expect(page).to have_content "what's your name"

features/support/start_game.feature
tansaku: this could be DRYed out with "Then I should see What's your name?", but two repeats for me is usually borderline - a third similar step would make me want to create a generic step

spec/game_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'game.rb'
+
+describe Game do
+  let(:game){Game.new}
+
+  context 'Adding players' do
+
+  it 'should be able to add a player' do
```
tansaku: inside context should be indented

lib/game.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def result(player_choice, opponent_choice)
+    if opponent_choice == combinations[player_choice]
+      value = "Win!"
+    elsif player_choice == combinations[opponent_choice]
+      value = "Lost!"
+    elsif player_choice == opponent_choice
+      value = "Draw!"
+    end
+    return value
```
tansaku: value is unecessary here - could be removed completely, e.g. 

```ruby
  def result(player_choice, opponent_choice)
    if opponent_choice == combinations[player_choice]
      "Win!"
    elsif player_choice == combinations[opponent_choice]
      "Lost!"
    elsif player_choice == opponent_choice
      "Draw!"
    end
  end
```

also note that ideally we don't return display strings from our internal objects - you should prefer objects or symbols.  You are not effectively separating the display of the interface from the game logic.

Also note that this long elsif could be condensed much further with appropriate use of early returns and ternary operators - have a go condensing this to 3 lines or less

lib/game.rb
```diff
@@ -0,0 +1,36 @@
+class Game
+
+  attr_reader :combinations, :randomness, :name
+
+  def initialize
+    @combinations = { rock: :scissors, paper: :rock, scissors: :paper }
+    @randomness = combinations.keys
+    @player_one = nil
```
tansaku: this is redundant - in ruby all vars are initialized to nil anyway

views/game.erb
```diff
@@ -0,0 +1,12 @@
+<h3 class="text-center"> Hello <%= @name %> pick your flavour ! </h3>
+<div class="text-center"></div>
```
tansaku: in semantic HTML we prefer not to use divs

views/index.erb
```diff
@@ -0,0 +1,12 @@
+
+  <h3 class="text-center">You don't know who's gonna pay for the drinks tonight?</h3>
+  <h4 class="text-center">Resolve this with Rock-Paper-Scissors!</h4>
+  <br>
+  <img class="img-responsive center-block" src="http://i.imgur.com/8PMuVQF.png">
+  <br>
+  <br>
+  <div class="text-center">
```
tansaku: in semantic HTML we prefer not to use divs

views/index.erb
```diff
@@ -0,0 +1,12 @@
...
+  <form action='/player' method='get'>
```
tansaku: this form is redundant when we are just using a link

views/layout.erb
```diff
@@ -0,0 +1,20 @@
...
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
+  <script src="http://maxcdn.bootstrapcdn.com/bootstrap/3.2.0/js/bootstrap.min.js"></script>
+</head>
+<body>
+<div class="container">
+  <div class="jumbotron">
+
+<%= yield %>
+
+</div>
```
tansaku: please ensure html is also indented correctly

app/server.rb
```diff
@@ -0,0 +1,50 @@
...
+  enable :sessions
+  game = Game.new
+  player = Player.new
+
+  get '/' do
+    erb :index
+  end
+
+  get '/player' do
+    @name = params[:name]
```
tansaku: this is not being used - should be removed

app/server.rb
```diff
@@ -0,0 +1,50 @@
...
+    erb :player
+  end
+
+  post '/player' do
+    @name = params[:name]
```
tansaku: great that you are grabbing the name from the form, but note that you are then not assigning this to anything - so all that work of collecting the name is for nothing - perhaps you meant to assign this to the player object?

app/server.rb
```diff
@@ -0,0 +1,50 @@
...
+    session[:player] = player
```
tansaku: since you have player as global, it's redundant to add it to the session, and generally you should avoid adding objects to the session since it is limited in size.  You should prefer to add ids to the session where possible

views/game.erb
```diff
@@ -0,0 +1,12 @@
...
+
+<% 2.times do %><br><% end%>
```
tansaku: you're not really saving yourself any characters here and also you should be avoiding <br>.  We suggest <p> instead - or use css to extend margin-bottom

views/game.erb
```diff
@@ -0,0 +1,12 @@
...
+
+<form action="/result" method="post" class="text-center">
+    <button type="submit" value="rock" class="btn" name="game"><img src="http://i.imgur.com/Wu4kvGl.png"></button>
```
tansaku: any reason you are using external images here rather than the ones in the public directory

app/server.rb
```diff
@@ -0,0 +1,50 @@
...
+    erb :game
+  end
+
+  get '/game' do
+    @name = params[:name]
```
tansaku: seems like this is not being used at all

app/server.rb
```diff
@@ -0,0 +1,50 @@
...
+    session[:player] = player
+    erb :game
+  end
+
+  post '/game' do
+    @name = params[:name]
```
tansaku: seems like this is not being used at all

app/server.rb
```diff
@@ -0,0 +1,50 @@
...
+    session[:player] = player
+    erb :result
+  end
+
+  post '/result' do
+    player = session[:player]
+    @name = params[:name]
```
tansaku: seems like this is not being used at all

app/server.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  post '/game' do
+    @name = params[:name]
+    session[:player] = player
+    erb :result
+  end
+
+  post '/result' do
+    player = session[:player]
```
tansaku: again this doesn't need to be grabbed from session as you have it already declared globally

app/server.rb
```diff
@@ -0,0 +1,50 @@
...
+    @name = params[:name]
+    @computer = game.random
```
tansaku: this line and below feel like we have too much logic in the controller - much of this could be moved into the game class

lib/player.rb
```diff
@@ -0,0 +1,7 @@
+class Player
```
tansaku: this class should be doing more - the change from string to symbol doesn't really need it's own method - player class should be holding name of player, perhaps checking option selected are correct.

tansaku:
Overall good work here Sebastien.  Watch out though for redundant code that's not being used for anything, and for HTML semantics.  Would be nice to see a revision with your player class taking a bit more responsibility.

spike01:
:checkered_flag:

PR #2, opened 2015-02-22
rock paper scissors

meads58:
https://warm-inlet-3430.herokuapp.com/

The Sinatra side of things isn't that pretty, pretty new to HTML and CSS so any advice on how I've done things would be great.

If you were given my lib and spec files and asked to add in a new feature could you easily do this just from my files or do I need more work on my naming of methods?


app/server.rb
```diff
@@ -0,0 +1,59 @@
...
+    erb :play_game
+  end
+
+  post '/begin' do
+    if params[:player].empty?
+      @enter_name = "You Must Enter Your Name"
+      erb :new_player
+    else
+      session[:player] = Player.new(params[:player])
+      session[:computer] = Computer.new(Gem::Platform.local.os)
```
pitchinvasion: This is pretty nifty, I wasn't sure exactly what you were doing at first but I like the idea of giving the AI an identity!

app/server.rb
```diff
@@ -0,0 +1,59 @@
...
+  get '/begin' do
+    erb :play_game
+  end
+
+  post '/begin' do
+    if params[:player].empty?
+      @enter_name = "You Must Enter Your Name"
+      erb :new_player
+    else
+      session[:player] = Player.new(params[:player])
```
pitchinvasion: Don't set objects directly into the session, it causes all kinds of problems and often errors out with weird errors. Instead think of another thing you could store in the session (perhaps using the #object_id method), and then retrieve that from the session when you need it and use it to look up your object which you can store in a hash perhaps inside the sinatra app.

It's a strange approach you'll see this week how using databases a lot of these problems will fortunately disappear!

app/server.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  post '/result' do
+    begin
+      session[:player].select_player_weapon(params[:weapon], weapons)
+      session[:computer].random_weapon(weapons)
+      session[:result] = GameResult.new(umpire, session[:player], session[:computer])
+      redirect '/result'
+    rescue
+      @error_message = "weapons selection was wrong go back and try again"
```
pitchinvasion: You shouldn't be rescuing like this as this rescues all errors - only rescue specific errors. Generally you don't want your app raising errors unless something exceptional happens, I think maybe you are looking to do some validation instead. You could do this in a simple check maybe

```
 if player.weapon_valid?(weapon)
```

features/rock_paper_scissor.feature
```diff
@@ -0,0 +1,39 @@
...
+    And press the "Submit" button
+    Then I will see "You Must Enter Your Name"
+
+  Scenario: Start game with players name showing
+    Given I am on the "new_player" page
+    When I enter "player" as "Maverick"
+    And press the "Submit" button
+    Then I will see "Maverick"
+
+  Scenario: Choose a weapon
```
pitchinvasion: Choosing a weapon is a different feature no? 

As a registered player
I want to be able to choose a weapon
So I can try and win at the game

features/rock_paper_scissor.feature
```diff
@@ -0,0 +1,39 @@
+Feature: Start game
+  As a new player of rps
+  I want to create myself as a player
+  So I can play the game
+
+  Scenario: Go to new player page
```
pitchinvasion: These are good! :)

lib/computer.rb
```diff
@@ -0,0 +1,14 @@
+class Computer
+
+  attr_reader :weapon, :name
+
+  def initialize(name = 'computer')
+    @weapon = {}
```
pitchinvasion: You don't want to initialize this as a hash, you in fact don't even need to initialize this as anything until it's needed, just remove this line

lib/game_result.rb
```diff
@@ -0,0 +1,37 @@
...
+    attr_reader :weapons_users
+
+  def initialize(umpire, player, computer)
+    @umpire = umpire
+    @player = player
+    @computer = computer
+    @weapons_users = {}
+  end
+
+  def get_weapon_choices
```
pitchinvasion: I would rename this method we avoid using `get_` in ruby method names as it is unnecessary extra words (we like to keep things short!) and also in this case it's misleading, you're not getting here you're actually setting things to weapon_users.

Instead of doing this, why not have weapon_users as a method that returns the correct values when you need it. You could remove your attr_reader and replace it with 

```
def weapon_users
  { 
    @player.weapon => @player.name,
    @computer.weapon => @computer.name
  }
end
```

lib/game_result.rb
```diff
@@ -0,0 +1,37 @@
...
+    weapons_users.merge!({ @computer.weapon => @computer.name })
+  end
+
+  def announcement
+    get_weapon_choices
+    winning_weapon = check_weapons(weapons_users.keys)
+    result_headliner(winning_weapon)
+  end
+
+  private def check_weapons(weapons)
```
pitchinvasion: This is fine but we tend to use the private keyword once, and then all methods below that keyword (which always appears below all our public methods) ruby marks as private. Again it just makes things cleaner

lib/game_umpire.rb
```diff
@@ -0,0 +1,24 @@
+class GameUmpire
```
pitchinvasion: Nice thinking to have a GameUmpire class, it shows you've been thinking about the Single Responsibility Principle and OOD

lib/game_umpire.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  ROCK_BEATS_SCISSORS   = [:rock, :scissors]
+  PAPER_BEATS_ROCK      = [:paper, :rock]
+  SCISSORS_BEATS_PAPER  = [:paper, :scissors]
+
+
+  def rule_logic(weapons)
+    played_weapons = weapons.sort
+
+    case
```
pitchinvasion: We tend to try and avoid case/when statements as they are so long. Can you think of a way of re-writing this method not using case/when? Think about how if you changed things around to use a hash you could reduce this method to one line!

lib/game_weapons.rb
```diff
@@ -0,0 +1,22 @@
+class GameWeapons
+
+  attr_reader :weapons
+
+  def initialize()
```
pitchinvasion: you don't need brackets if there are no args

lib/game_weapons.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def random_weapon
+    weapons.sample
+  end
+
+  def select_weapon(weapon)
+    if weapons.include?(weapon)
+      return weapon
+    else
+      raise "Can't select that weapon"
```
pitchinvasion: As I said above I tend to avoid raising exceptions, as if this is not caught it will crash your whole app - I prefer to make sure in my tests that the wrong weapon won't be chosen, and if there is a danger I would use validations. 

But if you are going to raise here you should raise a custom exception so you can catch that specific error, the reason being otherwise if there were other genuine errors in this method your app would never catch them properly and you wouldn't realise you had a bug in your game

lib/player.rb
```diff
@@ -0,0 +1,15 @@
+class Player
+
+  attr_reader :name, :weapon
+
+  def initialize(name)
+    @name = name
+    @weapon = String.new
```
pitchinvasion: Again if you don't need a value for this to start with leave it blank

spec/computer_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'computer'
+
+describe Computer do
+
+  let(:game_weapons) {double :GameWeapons}
+
+  it "should return weapon paper when assigned this weapon" do
```
pitchinvasion: in terms of behaviour this isn't realy what you are defining what your computer wants to do, you need to think about it more by asking the question "what do I want this class to do?"

I think I would say it "selects a random weapon"

spec/game_umpire_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'game_umpire'
+
+describe GameUmpire do
+
```
pitchinvasion: Really good spec! Reads really nicely too in natural english

spec/game_result_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+describe GameResult do
+
+  let(:umpire) {double :umpire }
+  let(:jim) { double :player }
+  let(:imac) {double :computer }
+  let(:result) { GameResult.new(umpire, jim, imac) }
+
+
+  it "can say 'player wins' when the player wins" do
```
pitchinvasion: Great! Not saying "should return" makes this read really nicely

spec/game_weapons_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+describe GameWeapons do
+
+  let(:weapons) {GameWeapons.new}
+
+
+  it "is created with 3 weapons rock, paper and scissors" do
+    expect(weapons.weapons).to match_array [:rock, :scissors, :paper]
+  end
+
+  it "will rasie and error if player selects unavalible weapon" do
```
pitchinvasion: oops, does your text editor come with a spell checker?

views/index.erb
```diff
@@ -0,0 +1,4 @@
+
+<h1>Welcome to Rock Paper Scissors</h1>
+<h2>Time to start a new game</h2>
+<a href="/new_player">Click To Begin</a>
```
pitchinvasion: Wrap <a> tags in <p> tags if they're not in another tag

views/layout.erb
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <title>Rock Paper Scissors</title>
+    <link rel="stylesheet" type="text/css" href="css/style.css">
```
pitchinvasion: Indentation matters in HTML too!

views/new_player.erb
```diff
@@ -0,0 +1,5 @@
+<form action="/begin" method="post">
+ Enter players name <input type="text" name="player"><br>
```
pitchinvasion: Avoid using <br> have a google about ways of semantically marking up forms so you don't need them

views/play_game.erb
```diff
@@ -0,0 +1,30 @@
+<audio controls autoplay>
+  <source src="/sounds/zza.mp3" type="audio/mpeg">
```
pitchinvasion: Ha the audio was fun although I warn you most users hate audio on websites!

views/play_game.erb
```diff
@@ -0,0 +1,30 @@
...
+</audio>
+
+<h1>Begin game</h1>
+<h2>Today <%=session[:player].name%> you are playing the mighty <%=session[:computer].name%> machine</h2>
+
+<h1>
```
pitchinvasion: You normally only have one <h1> tag per page (lots of reasons but it matters for SEO), I'd make this an <h2> probably

views/play_game.erb
```diff
@@ -0,0 +1,30 @@
...
+  <%=session[:player].name%>
+  take your time and choose your weapon
+</h1>
+<br>
```
pitchinvasion: Don't use <br> tags here, if you need more space below your <h1> use margin-bottom

views/play_game.erb
```diff
@@ -0,0 +1,30 @@
...
+  <input type="submit" name="weapon" value="Paper" />
+  <input type="submit" name="weapon" value="Scissors" />
+</form>
+
+<form action="/" >
+  <input type="submit" value="New Player" />
+</form>
+
+<h1>
+  <font color="red">
```
pitchinvasion: `<font>` tags are evil! Use css instead to add colour, also don't use `<u>` again you can use css for this required effect, and again you don't want to use `<h1>`

pitchinvasion:
:checkered_flag:  I added lots of comments but generally this was really good, well-tested and good object design. I think in terms of method names just see if you can make them shorter if possible - we always aim for less underscores where possible.

Your HTML needs to be a bit more semantic, particularly avoid `<font>` type tags - all styling goes in CSS. And obviously design-wise it could look a bit cleaner but that's not really my remit!

PR #3, opened 2015-02-22
Finished challenge

robertpulson:


features/step_definitions/starting_a_game.rb
```diff
@@ -0,0 +1,11 @@
...
+  visit '/'
+end
+
+When(/^I follow "(.*?)"$/) do |arg1|
+  pending # express the regexp above with the code you wish you had
+end
+
+Then(/^I should see "(.*?)"$/) do |arg1|
+  pending # express the regexp above with the code you wish you had
+end
```
spike01: Did you decide against using Cucumber?

features/beginning_gameplay.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Entering your name
+  In order to set game up
+  As a enthusiastic battleshipper
+  I want to begin playing
+
+  Scenario: Beginning Gameplay
+    Given I am on enter_name page
+    When I provide "name" and click "Begin"
+    Then I should see the board
```
spike01: This is copied over from Battleships

lib/game.rb
```diff
@@ -0,0 +1,28 @@
+class Game
+
+  attr_accessor :results, :choice
```
spike01: Unless you have to change the value of the results, don't use an `attr_accessor` - use a reader instead

lib/game.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def initialize
+    @results = { :Rock     => [:Scissors, :Lizard],
+                 :Paper    => [:Rock, :Spock],
+                 :Scissors => [:Paper, :Lizard],
+                 :Lizard   => [:Spock, :Paper],
+                 :Spock    => [:Scissors, :Rock] }
+    @choice
```
spike01: You don't need to initialize `@choice` if it doesn't have a value when the object is created

lib/game.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def generate_choice
+    @choice = results.keys.sample
+  end
+
+  def play_with(player)
+    return 'Ties With' if self.choice == player.choice
```
spike01: `self` is implicit, you don't need it here or on line 20

lib/player.rb
```diff
@@ -0,0 +1,22 @@
...
+    @choice
+  end
+
+  def name_player(name)
+    @name = name
+  end
+
+  def choose(choice)
+    @choice = choice
+  end
```
spike01: While these are good in terms of expressive naming, they are effectively `attr_writers` for `@name` and `@choice` - so in this case, an `attr_accessor` would be appropriate

lib/player.rb
```diff
@@ -0,0 +1,22 @@
+class Player
+
+  attr_reader :name, :choice
+
+  def initialize
+    @name = name
```
spike01: You aren't passing anything in to the constructor, so this assignment doesn't make sense

lib/game.rb
```diff
@@ -0,0 +1,28 @@
+class Game
+
+  attr_accessor :results, :choice
+
+  def initialize
+    @results = { :Rock     => [:Scissors, :Lizard],
+                 :Paper    => [:Rock, :Spock],
+                 :Scissors => [:Paper, :Lizard],
+                 :Lizard   => [:Spock, :Paper],
+                 :Spock    => [:Scissors, :Rock] }
```
spike01: Generally, prefer lower case for symbols

app/server.rb
```diff
@@ -0,0 +1,40 @@
...
+  get('/') do
+    erb :index
+  end
+
+  get('/enter_name') do
+    erb :enter_name
+  end
+
+  post('/play') do
+    player.name_player(params[:player_name].capitalize.to_s)
```
spike01: You could encapsulate the conversion in your `name_player` method, instead of having it happen in the controller

spec/game_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+describe Game do
+
+  let (:game)   { Game.new                          }
+  let (:player) { double :player, choice: :Scissors }
+
+  it 'can generate a random option' do
+    game.generate_choice
+    expect(game.choice).to satisfy{ :Rock || :Paper || :Scissors || :Lizard || :Spock }
+  end
```
spike01: This test is brittle - you can change those symbols to anything and it'll still pass. You can mock `game.generate_choice` by using `expect(game).to_receive(:generate_choice).and_return(:scissors)` and test for that instead

spec/game_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  it 'can play the game with a player' do
+    expect(game).to respond_to(:play_with)
+  end
```
spike01: Try to avoid testing only if something responds to a method - run the method properly as you do later

spec/player_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'player'
+
+describe Player do
+
+  let (:player) { Player.new }
+
+  it 'can have a name' do
+    expect(player).to respond_to(:name)
+  end
```
spike01: See above

spec/player_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  it 'can be given a name' do
+    player.name_player('Phil')
+    expect(player.name).to eq('Phil')
+  end
+
+  it 'can make a choice' do
+    expect(player).to respond_to(:choose)
+  end
```
spike01: See above above

views/enter_name.erb
```diff
@@ -0,0 +1,23 @@
...
+  </div>
+  
+  <form action='/play' method='post'>
+    <input type='text' name='player_name' class='button' id='enter_name'>
+    <input type='submit' value='START GAME' class='button' id='button'>
+  </form>
+
+</body>
+
+</html>
```
spike01: Indentation on nested tags

views/index.erb
```diff
@@ -0,0 +1,22 @@
...
+    <img src="css/rockpaperscissors.jpg" alt="Rock Paper Scissors Lizard Spock" width="800px" height="500px">
+  </div>
+  
+  <form action='/enter_name' method='get'>
+    <input type='submit' value='PLAY' class='button' id='button'>
+  </form>
+
+</body>
+
+</html>
```
spike01: Consider using a layout.erb?

spike01:
Overall a good submission - only a few small snags that could do with improvement. Bit of a shame you didn't get Cucumber working as the rest of it is strong.

:checkered_flag:

features/beginning_gameplay.feature
tansaku: Yeah, force yourself to type out from memory where possible - it really helps - trust us!

PR #4, opened 2015-02-22
challenge submission

GabeMaker:


features/playing.feature
```diff
@@ -0,0 +1,19 @@
...
+
+	Scenario: Losing with Rock
+		Given opponent randomly picks "Paper"
+		When I see "Pick one:"
+		And I push the "Rock" button
+		Then I should see "Sorry Gabe, you lose"
+
+	Scenario: Winning with Rock 
+		Given opponent randomly picks "Scissors"
+		When I see "Pick one:"
```
stephenlloyd: Whats the purpose of this 'when', Could 'and' be used instead?

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,219 @@
+# # Taken from the cucumber-rails project.
```
stephenlloyd: comments are the devil. This would annoy potential code reviewers. I know why you did it though and I mean it merely for future reference.

features/support/paths.rb
```diff
@@ -0,0 +1,29 @@
+# Taken from the cucumber-rails project.
+
+module NavigationHelpers
+  # Maps a name to a path. Used by the
+  #
+  #   When /^I go to (.+)$/ do |page_name|
+  #
+  # step definition in web_steps.rb
+  #
```
stephenlloyd: just delete this file :)

lib/game.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def initialize
+    @players = []
+  end
+
+  def add(player)
+    raise "Sorry, you can't have more than two players" if players.count == 2
+
+    @player_one = player if players.count == 0
+    @player_two = player if players.count == 1 && players
```
stephenlloyd: This really isn't that bad, but there is a slightly more efficient way of doing it with a ternary. Check battleships.

lib/game.rb
```diff
@@ -0,0 +1,51 @@
...
+    players << player
+  end
+
+  def winner
+    if player_one.shape == :rock && player_two.shape == :scissors
```
stephenlloyd: This is way too long. Essentially a rock beats scissors... this is a perfect key value pair. If you had a WINNERS hash constant you could easily programe this. 
```ruby 
WINNERS = {rock: :scissors, scissors: :paper, paper: :rock}
```

```ruby
@winner = ( WINNERS[player_one.shape] == player_two.shape ? player_one : player_two
```

lib/game.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def add(player)
+    raise "Sorry, you can't have more than two players" if players.count == 2
+
+    @player_one = player if players.count == 0
+    @player_two = player if players.count == 1 && players
+    players << player
+  end
+
+  def winner
```
stephenlloyd: the else at the end you could probably do at the top simply like this 
```ruby
return :tie if player_one.shape == player_two.shape
```

lib/player.rb
```diff
@@ -0,0 +1,21 @@
+class Player
+  
+  attr_reader :name, :shape
```
stephenlloyd: nice simple class,  I like it. Good to see you're taking a lot from poodr with your initializing the hash.

lib/game.rb
```diff
@@ -0,0 +1,51 @@
...
+      @winner = player_one
+    elsif player_two.shape == :scissors && player_one.shape == :paper
+      @winner = player_two
+    elsif player_one.shape == :rock && player_two.shape == :paper
+      @winner = player_two
+    elsif player_two.shape == :rock && player_one.shape == :paper
+      @winner = player_one
+    else
+      @winner = :tie
+    end
```
stephenlloyd: my biggest issue with this message is assigning the instance variable @winner. It's not being used. Nowhere else in the class are you using @winner so it's pointless. All you want is to return the winner from this method. Sorry for harsh words.

server.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    # computer.pick [:rock, :paper, :scissors].sample
+    GAME.add computer
+
+    erb :name
+  end
+
+  post '/play' do
+    GAME.player_one.pick [:rock, :paper, :scissors].sample
+    GAME.players.pop if GAME.players.count == 2
```
stephenlloyd: this is quite puzzling. I would probably reset the game somehow and wrap it in a method to do this. Even wrapping this line(19) in a descriptive method would be better.

server.rb
```diff
@@ -0,0 +1,46 @@
...
+    erb :play
+  end
+
+  post '/result' do
+
+    @user_shape     = GAME.player_two.pick params.values.first.to_sym
+    @computer_shape = GAME.player_one.shape
+    @name = GAME.player_two.name
+
+    if GAME.winner == GAME.player_one
```
stephenlloyd: you could easily set this in one line

server.rb
```diff
@@ -0,0 +1,46 @@
...
+    @name     = params[:name]
+    user      = Player.new(name: @name)
+    GAME.add user
+
+    erb :play
+  end
+
+  post '/result' do
+
+    @user_shape     = GAME.player_two.pick params.values.first.to_sym
```
stephenlloyd: params.values.first.to_sym this scares me. The point of key value pairs is to use the key. What happens if this post has additional params, you can't guarantee the order.

views/result.erb
```diff
@@ -0,0 +1,15 @@
+<p>You chose <%= @user_shape %></p>
+<p>Your opponent chose <%=	@computer_shape %></p>
+
+<% if @winner == :computer %>
+	Sorry <%= @name %>, you lose
```
stephenlloyd: whilst this is quite small there might be a better way to write this more simply. It's ok in this instance but pretty soon logic like this can make your code unreadable.

stephenlloyd:
Summary: It's my job to be hard on you and in reality you did really well and it's one of the top ones I've seen. You're right that the long elseif statement was too long and you should definitely refactor. I'm really happy with your tests and you definitely get the badge for this. Good job.  :checkered_flag:

PR #5, opened 2015-02-22
Erik's Rock Paper Scissors

ErikAGriffin:


app/lib/computer.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  def weapon
+    @weapon ||= :no_weapon
+  end
+
+  def choose_weapon
+    @weapon = [:ROCK,:PAPER,:SCISSORS].sample
+  end
+
```
roidriscoll: Symbols should be lowercase so they aren't confused with constants, but this is more of a stylistic note.

app/lib/computer.rb
```diff
@@ -0,0 +1,16 @@
+class Computer
+
+  def name
+    "the Computer"
+  end
+
+  def weapon
+    @weapon ||= :no_weapon
+  end
+
```
roidriscoll: In which case would you need `:no_weapon`

app/server.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  post '/play' do
+    @browser_id = session[:session_id]
+    @game = games.find {|game| !game.full?}
+    @player = Player.new(params[:name],@browser_id)
+    @game.add_player(@player)
+    @opponent = @game.opponent_of(@browser_id)
+    session.store(:game_id, @game.uuid)
+
```
roidriscoll: You could also use `:object_id` here, instead of going to the trouble of creating a unique reference

pitchinvasion:
This is excellent, well-written tests and generally well-designed objects. However I think you could be pulling more logic out of your controllers - you want to design your apps so ideally there is an absolute minimum logic there.

Also with the uuid thing you should be aware that although it is almost certainly unique it isn't 100% certain of that, in this case using object_id (the simpler solution) is also the more reliable one.

But generally this is really good :checkered_flag:

app/views/layout.erb
```diff
@@ -0,0 +1,39 @@
...
+<html>
+    <head>
+        <title>RPS</title>
+
+<style type="text/css">
+
+    body {
+    text-align:center;
+    }
+
```
roidriscoll: Your back end code is really clean and showcases a lot of care and attention. However, this file almost seems to have been authored by a different person. If this was an interview situation this would be a red flag. As developers we should be just as careful throughout the stack.

ErikAGriffin: Can you explain what you are referencing by "controllers"?  What logic should I be pulling out and to where?

tansaku:
some additional follow up 

Your spec folder and lib folder are in incorrect places, preventing specs from running.

It looks like you moved them as part of getting the web app working, which is something you want to avoid if at all possible.

You have cucumber pending tests, but I believe we cleared up those issues up in the cucumber and stubbing breakout:

https://www.youtube.com/watch?v=m-pWxBEqpz0

PR #6, opened 2015-02-23
Challenge submission

sphaughton:
Unfinished but close. Stuck on an error message I'm receiving where after entering a name into the form I get ```#<NoMethodError: undefined method `pick=' for #<Player:0x007fc1848fe768 @name=nil>>``` and can't see where this is in my code, even using sublime text's find function.


sphaughton:
I've solved this error since submitting by adding ```:pick``` to the accessor in my ```Player``` class, also removed where I was calling on ```weapon``` which was a method I hadn't defined and supposed to be changed to ```pick```

features/starting_a_game.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Starting the game
+  In order to play Rock Paper Scissors
+  As a bored marketeer
+  I want to start a new game
+  
+  Scenario: Starting a game
+  Given I am on the homepage
```
silvabox: Indent Given/When/Then under Scenario

lib/computer.rb
```diff
@@ -0,0 +1,16 @@
+class Computer
+
+  attr_reader :pick
+
+  def pick
```
silvabox: This method is not tested!  It fails when running the game - which should have been caught in your unit tests

lib/player.rb
```diff
@@ -0,0 +1,16 @@
+class Player
+
+  attr_accessor :name
+
+  def initialize(name)
+    @name = name
+  end
+
+  def name
```
silvabox: This method is not required as you are using `attr_accessor`.  Note as name is passed on initialization and presumably doesn't change, use `attr_reader` instead.

lib/player.rb
```diff
@@ -0,0 +1,16 @@
...
+    @name
+  end
+
+  def pick(value)
```
silvabox: You created a `pick` method to set the pick, but there is no method to read the value of pick back from the player.  Maybe use `attr_accessor :pick` or couple your `pick` mutator method with a `picked` reader method.

lib/rockpaper.rb
```diff
@@ -0,0 +1,41 @@
+require 'sinatra/base'
+require_relative 'player'
+require_relative 'computer'
+require_relative 'game'
+
+class RockPaper < Sinatra::Base
+
+  enable :sessions
+
+  player = Player.new(:player_name)
```
silvabox: Here, you are instantiating a new player with the name `:player_name` - is this intended?  If the state of player is not known at the time of initializing the game, then defer creating the player until it is known and then add the player to the game using a method `add_player` or similar in the game.

lib/rockpaper.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  post '/' do
+    if params[:player_name].empty?
+      @name = "Please tell me your name!"
+      erb :index
+    else
+      @name = params[:name]
+      @weapon = params[:weapon]
+      player.name = @name
+      player.pick = @pick
```
silvabox: You can't use `player.pick = @pick` unless you have a method called `pick=` on player.  You would need to say `player.pick(@pick)` as it is now.  But what is `@pick` at this point?

spec/computer_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'computer'
+
+describe Computer do
+
+  let(:computer){Computer.new}
+
+  it "can take a turn" do
+  allow(computer).to receive(:pick).and_return("Paper")
```
silvabox: You have not indented these lines correctly

spec/computer_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'computer'
+
+describe Computer do
+
+  let(:computer){Computer.new}
+
+  it "can take a turn" do
```
silvabox: This test is vacuous as you are not testing the `Computer` class at all.  You are creating a stub saying receive `pick` and return `"Paper"`, which means the actual `pick` method of Computer is not being called at all - your testing that the stub returns `'"Paper"` when you call `pick` - which is exactly what you just defined it to do.

spec/game_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    allow(computer).to receive(:pick).and_return "scissors"
+    expect(game.winner).to eq "You win!"
+  end
+
+  it "should know if there is a draw" do
+    allow(player).to receive(:pick).and_return "rock"
+    allow(computer).to receive(:pick).and_return "rock"
+    expect(game.winner).to eq "Draw! Try again"
+  end
+
```
silvabox: You haven't tested the computer winning

spec/player_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  let(:player){Player.new("Alex Kidd")}
+  
+  it 'has a name' do
+    player.name
+    expect(player.name).to eq "Alex Kidd"
+  end
+
+  it 'can take a turn' do
+    player.pick("paper")
```
silvabox: This line does nothing as you call the same method again in the follow expectation

spec/player_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    expect(player.pick("paper")).to eq "paper"
```
silvabox: You are not testing the state of the player in this test - just that the `pick` method returns `"paper"` which was passed into it - which it does simply because this is the last expression evaluated in the method.

spec/player_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'player.rb'
+
+describe Player do
+
+  let(:player){Player.new("Alex Kidd")}
+  
+  it 'has a name' do
+    player.name
```
silvabox: This line is redundant.  Calling `player.name` in isolation like this does nothing.

lib/computer.rb
```diff
@@ -0,0 +1,16 @@
+class Computer
+
+  attr_reader :pick
```
silvabox: `attr_reader` creates a method `pick` which reads a `@pick` instance variable.  However, you have defined a `pick` method yourself, so the `attr_reader` is rendered obsolete.

lib/computer.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def pick
+    number == rand(1..9)
```
silvabox: Why `rand(1..9)`?  Why not `rand(1..3)` and just test for 1, 2 and 3?

lib/computer.rb
```diff
@@ -0,0 +1,16 @@
...
+    if number <= 3
+      @pick = "rock"
```
silvabox: Using an instance variable `@pick` is not necessary as every call to `pick` generates a new pick and the value is not accessed anywhere else in the instance.

lib/game.rb
```diff
@@ -0,0 +1,30 @@
...
+    @computer = computer
+    @players = [player, computer]
+    @beats = {
+      "rock" => ["paper"],
+      "paper" => ["scissors"],
+      "scissors" => ["rock"], 
+    }
+  end
+
+  def player
```
silvabox: You've user `attr_reader :players` so why not `attr_reader` for `player` and 'computer` also?

lib/game.rb
```diff
@@ -0,0 +1,30 @@
...
+    @player
+  end
+
+  def computer
+    @computer
+  end
+
+  def winner
```
silvabox: As a general rule, don't include presentation logic in your business logic.  Instead of returning presentation strings here, return codes to represent win/lose/draw (e.g. symbols), then transform those codes in the view layer.

silvabox:
Good attempt Sean, but you're game isn't playable as there are a number of breaking issues that should have been caught in unit tests.  There's quite a bit of confusion over when to use `attr_reader` etc. and how to test you methods and resulting instance state.  Grab me if you have any questions - I'll happily go through it with you.

spike01:
:checkered_flag:

PR #7, opened 2015-02-23
My version of RPS

vvirgitti:
The challenge is incomplete as the result page doesn't display correctly the result. I struggled mostly with creating the rules on ruby and also understanding how to use the params correctly.


app/server.rb
```diff
@@ -0,0 +1,53 @@
...
+    erb :index
+  end
+
+  get '/register' do
+    erb :register
+  end
+
+  post '/register' do
+    if params[:name].empty?
+      @message = "Please enter your name"
```
henrygarner: Your message should ideally be in the view ... Your controller only needs to determine if there's an error, not create the actually content of your page.

app/server.rb
```diff
@@ -0,0 +1,53 @@
...
+      erb :register
+    else
+      player = Player.new(params[:name])
+      @welcome = "Welcome #{params[:name]}"
```
henrygarner: Your welcome should also be in the view. If player was an instance variable you could refer to `@player.name` in the view.

app/server.rb
```diff
@@ -0,0 +1,53 @@
...
+      erb :choose
+    end
+  end
+
+  get '/choose' do
+    game = Game.new
```
henrygarner: Likewise, `game` won't be accessible in the view since only `@`variables are available. Did you want to use this in the view at some point and then change your mind?

app/server.rb
```diff
@@ -0,0 +1,53 @@
...
+    erb :choose
+  end
+
+  get '/result' do
+    computer = Computer.new
```
henrygarner: Same as with `game`/`@game`. Not currently being used so ought to be removed.

app/server.rb
```diff
@@ -0,0 +1,53 @@
...
+    erb :result
+  end
+
+  post '/result' do
+    game = Game.new
+    @choice = "You chose: #{params[:choice]}"
+    @weapon = %w(Rock Paper Scissors).sample
```
henrygarner: You have written `computer.weapon` to encapsulate `%w(Rock Paper Scissors).sample` but you're not using it here?

app/server.rb
```diff
@@ -0,0 +1,53 @@
...
+    @opponent = "Your opponent has chosen: #{@weapon}"
+    @outcome = game.result(:choice,:weapon)
+    @results = "The outcome of the fight is #{params[:outcome]} has won"
```
henrygarner: There's too much going on in the controller here. Ideally your messages (`@choice`,`@opponent`,`@results`) will be created in the view. If `game` was `@game` you could use it to determine which player won in the view too and output the appropriate message directly.

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,66 @@
+#Scenario Getting to the registration form
+
+  Given(/^I am on the homepage$/) do
+    visit ('/')
+  end
+
+  When(/^I follow "(.*?)"$/) do |register|
+    click_on('New Game')
```
henrygarner: You ought to use the `register` local variable here. At the moment if your feature said 'I follow "Quit"', the test would still click on "New Game".

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  Then(/^I should see "(.*?)"$/) do |form|
+    expect(page).to have_content("What's your name?")
```
henrygarner: As above!

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,66 @@
...
+#Scenario Registration process
+
+  Given(/^I am on the register page$/) do
+    visit ('/register')
+  end
+
+  When(/^I type in my name$/) do
+    fill_in('name', :with => 'Joe')
+    click_on('Submit')
+    expect(page).not_to have_content("Please choose your mass destruction weapon")
```
henrygarner: You should maybe leave your expectations to the `Then` block. You have other expectations in the feature where this is used. More than one expectation is usually a bad thing.

This probably should be called "I type in my name and click submit".

Also I notice that this test fails because it can't find the Submit button. If you gave your button a `name="Submit"` and removed these expectations, your test would pass. :)

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,66 @@
...
+    expect(page).to have_content("Please enter your name")
+    # Trying to make the test pass with a stub or various expressions but to no avail yet
+  end
+
+  Then(/^I should be successfully registered$/) do
+    visit ('/choose')
```
henrygarner: You ought to have expectations in a Then... this is just going back to the `/choose` page and losing all your game state :(

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,66 @@
...
+
+
+#Scenario Choose a weapon
+
+  Given(/^I am on the game page$/) do
+  visit ('/choose')
+end
+
+When(/^I select an option and submit$/) do
+  select('submit', :from => 'choice')
```
henrygarner: You actually want to `select('Rock', :from => 'choice')` (or Paper, Scissors), and then `click_on("Submit")` (assuming you have given this submit button a `name="Submit"` too)

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,66 @@
...
+end
+
+Then(/^I should see the title "(.*?)"$/) do |choice|
+  expect(page).to have_content("You chose")
```
henrygarner: As above. You ought to be using the `choice` local variable.

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,66 @@
...
+end
+
+#Scenario Results
+
+Given(/^I am on the result page$/) do
+  visit ('/result')
```
henrygarner: This should probably be something like Given "I have played a game"
```ruby
  visit ('/register')
  fill_in('name', :with => 'Joe')
  click_on('submit')
  select('Rock', :from => 'choice')
  click_on('submit')
```

Then you'll be able to make assertions about what happens once you've played a game. Just visiting `/result` gives you essentially a blank page and loses all your game state.

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,66 @@
...
+end
+
+Then(/^I should see a mention "(.*?)"$/) do |opponent|
+  expect(page).to have_content("Your opponent has chosen")
```
henrygarner: As above. You ought to be using the `opponent` local variable.

features/support/starting_a_game.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Starting a game
```
henrygarner: This is good! I suspect if you hadn't run into difficulty you would have tests for if you win and lose as well.

features/support/starting_a_game.feature
```diff
@@ -0,0 +1,26 @@
...
+
+
+  Scenario: Getting to the registration form
+    Given I am on the homepage
+    When I follow "New Game"
+    Then I should see "What's your name?"
+
+  Scenario: Registration process
+    Given I am on the register page
+    When I type in my name
```
henrygarner: ..."and click submit"...

lib/computer.rb
```diff
@@ -0,0 +1,9 @@
+class Computer
+
+  attr_accessor :weapon
+
+  def weapon
+    ['Rock','Paper','Scissors'].sample
```
henrygarner: Good!

lib/game.rb
```diff
@@ -0,0 +1,13 @@
+class Game
+
+  attr_accessor :result
+
+  def result(p1, p2)
+    winning_moves = {"Rock" => "Paper", "Paper" => "Scissors", "Scissors" => "Rock"}
+    ([p1, p2] & winning_moves.values_at(p1, p2)).first
```
henrygarner: This is a novel approach. How would you deal with draws?

lib/paper.rb
```diff
@@ -0,0 +1,11 @@
+class Paper
+
+  def chosen(player)
+    true
```
henrygarner: I take it you were intending to do something else at some point? :)

lib/player.rb
```diff
@@ -0,0 +1,13 @@
+class Player
+
+  attr_reader :name
+
+  def initialize(name)
+    @name = name
```
henrygarner: This is great, though you don't appear to use it... It might be useful on the register page?

views/choose.erb
```diff
@@ -0,0 +1,16 @@
...
+<center><img src='/images/fight.jpeg'></center>
+
+
+<form action ='/result' method='post' align='center'>
+  Please choose your weapon of mass destructions
+  <select name='choice' size='1'>
+    <option> Rock
+    <option> Paper
+    <option> Scissors
+  <input type='submit' id='submit'>
```
henrygarner: `name="Submit"` would help here!

views/result.erb
```diff
@@ -0,0 +1,11 @@
+<h1 align="center"> <%= @choice %> </h1>
```
henrygarner: As above, prefer not to pass strings directly into the view (what about when you build an amazingly popular game and need to support lots of languages? :). Instead, if you built a little more logic into your game, you could say something like:

```erb
<% if @game.player_won? %>
  <p>Congratulations, you won!</p>
<% end %>
```

Your view templates are then the only thing which is language-specific

views/result.erb
```diff
@@ -0,0 +1,11 @@
...
+
+</br>
+</br>
+
+<h1 align="center"> <%= @opponent %> </h1>
+
+</br>
+</br>
+
+<h3 align="center"> <%= @results %></a></h3>
```
henrygarner: :checkered_flag:

henrygarner:
Overall good work! I've picked you up on lots fiddly small bits, but I think you would do well to read up on MVC and make sure you keep view logic out of the controller.

I sense you would have finished the challenge in a TDD way if you'd not got tripped up by tests not clicking the submit button...

Full marks for persevering. Well done :checkered_flag:

PR #8, opened 2015-02-23
Challenge submission

tomcoakes:
There's a more recent version of this HERE that combines singleplayer and multiplayer into the same app: https://github.com/tomcoakes/rps-multiplayer

And it's live on Heroku HERE: https://rockpaperscissors-tomcoakes.herokuapp.com


server.rb
```diff
@@ -0,0 +1,48 @@
...
+require_relative './lib/computer'
+require_relative './lib/player'
+
+class RPS < Sinatra::Base
+  
+  enable :sessions
+
+  game = Game.new
+  computer = Computer.new
+
```
danldb: Preferable to use constants GAME and COMPUTER

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,25 @@
...
+  click_link(link)
+end
+
+Then(/^I should see "(.*?)"$/) do |displayed|
+  expect(page).to have_content(displayed)
+end
+
+Given(/^the computer has chosen "(.*?)"$/) do |weapon|
+  # Need to find a way of stubbing this... may need to use sessions?
+end
```
danldb: You don't ordinarily use stubs in feature tests, as you are testing the user experience rather than the back end code. In this case, I would suggest unit testing / stubbing the random elements. Once you know that the method has been tested at the unit level, you do not need to test it in feature tests...

HOWEVER... this is a unique case! Ordinarily we would expect state to be held in a database, making stubbing unnecessary. In order to test your results pages, you do need to stub the game.

You can stub out your computer class using something like this:
``` 
computer = Capybara.app::COMPUTER 
allow(computer).to receive(choose_weapon).and_return(:rock)
```

features/finishing_the_game.feature
```diff
@@ -0,0 +1,24 @@
...
+
+  Scenario: The computer beats me
+    Given the computer has chosen "Paper"
+    When I click "Rock"
+    Then I should see "You lost!"
+
+  Scenario: I draw with the computer
+    Given the computer has chosen "Rock"
+    When I click "Rock"
+    Then I should see "It's a draw."
```
danldb: Good feature tests

lib/computer.rb
```diff
@@ -0,0 +1,9 @@
+class Computer
+
+attr_accessor :weapon
+
```
danldb: You don't need to use an instance variable here - calling the method choose_weapon should be enough.

lib/game.rb
```diff
@@ -0,0 +1,32 @@
+class Game
+  
+  attr_accessor :players
+
```
danldb: Do not use attr_accessor unless you absolutely have to. In this case you have a method to add a plaer, and you have a method players, so you can just delete line 3.

lib/game.rb
```diff
@@ -0,0 +1,32 @@
...
+      return @players[1]
+    when @players[0].weapon == :Rock && @players[1].weapon == :Paper
+      return @players[1]
+    when @players[0].weapon == :Rock && @players[1].weapon == :Scissors
+      return @players[0]
+    when @players[0].weapon == @players[1].weapon
+      return nil
+    end
+  end
+
```
danldb: Long case statements are a code smell - consider a hash lookup instead. 
Methods should be no longer than 3 lines.

lib/player.rb
```diff
@@ -0,0 +1,9 @@
+class Player
+
+  attr_accessor :weapon, :name
+
```
danldb: attr_accessor is too powerful. 

attr_reader would be appropriate, and then write a method to add or amend the weapon - this way you can control that it will only respond to a rock, paper, or scissors.

server.rb
```diff
@@ -0,0 +1,48 @@
...
+      erb :index
+    else
+      game.players = []
+      @player = Player.new(name)
+      game.add_player(computer)
+      game.add_player(@player)
+      erb :game
+    end
+  end
+
```
danldb: controllers should not contain logic - consider extracting this functionality to the game class.

server.rb
```diff
@@ -0,0 +1,48 @@
...
+    game.players[1].weapon = params[:weapon].to_sym
+    game.players[0].choose_weapon
+    @computer_weapon = game.players[0].weapon
+    @selected_weapon = game.players[1].weapon
+    @player = game.players[1]
+    @computer = game.players[0]
+    @winner = game.winner
+    erb :outcome
+  end
+
```
danldb: Controller is too fat - controllers should be skinny. Extract logic into relevant classes.

spec/computer_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'computer'
+
+describe Computer do
+
+  it "chooses a random weapon" do
+    computer = Computer.new
+    computer.choose_weapon
+    expect(computer.weapon).not_to eq nil
+  end
```
danldb: Instead of testing that it is not nil, you could give an array of possible values and check that it responds with one of those values.

spec/game_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    expect(game.winner).to eq computer
+  end
+
+  it "declares the game to be a draw when the player chooses rock and the computer chooses rock" do
+    game.add_player(computer)
+    game.add_player(player)
+    allow(computer).to receive(:weapon).and_return(:Rock)
+    expect(game.winner).to eq nil
+  end
+
```
danldb: Good use of doubles and stubs.

Use a 'before' block to DRY your code out - lots of repetition here.

Could make use of contexts for 'player chooses rock'. Again, this would dry out your code - also it would make your test names more concise without losing any information.

spec/player_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'player'
+
+describe Player do
+  it "has a weapon" do
+    player = Player.new
+    expect(player.weapon = :rock).to eq :rock
+  end
```
danldb: You are not really testing anything here. 

Consider:

```
player.weapon= :rock
expect(player.weapon).to eq :rock
```

danldb:
Hi Tom, Thanks for your submission!

Generally this was a strong submission. You have demonstrated a solid understanding of mocks and stubs, and your test coverage is good. I have proposed a solution to stubbing out your feature tests, but please remember that this is a very rare case and in 'the real world' it is usually best to avoid situations where you have to do this. 

Main areas of improvement for you:

1. controllers should be skinny - methods should be extracted to your classes and models.
2. ``` if else ``` and ``` case ``` are code smells - avoid anything more complicated than a ternary, preferring instead to use a hash lookup or some other data structure.
3. Remember to dry out your tests using before and let. 
4. Avoid attr_accessor - use attr_reader as a preference.
5. Your methods are WAY too long. Work on extracting logic out into small, concise, easily managed methods that each do one thing.

I really liked that you chose a common interface between Player and Computer - allowing you to substitute a computer for a player without any other area of your application become confused.

I hope that helps - let us know if you have any questions!

spike01:
:checkered_flag:

PR #9, opened 2015-02-23
Rock Paper Scissors

braunsnow:
Negatives: I started off using cucumber and then felt like I was being hindered a bit so prioritise completing the game and getting a good understanding of sessions. I tried working on a generic style sheet but wasn't happy with the result and a bit confused overall. My website code needs to be neater and should make better use of GET POST conventions.

Pros: I fully understand sessions, and feel comfortable with the display process in erb.


spec/player_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'player'
+
+describe Player do
+  let(:player){Player.new('Yannick')}
+
+
```
tansaku: indenting and spacing here is off - please get that fixed up

spec/player_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'player'
+
+describe Player do
```
tansaku: note that you are not checking anything about the choice attribute you added

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
+require 'sinatra/base'
```
tansaku: feels a little odd to have this class in lib, although I guess it's not unreasonable.  We tend to follow a pattern of having the sinatra code in a server.rb file in the app directory, although Sinatra is flexible about this.

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
...
+require 'active_support/all'
+require 'sinatra'
+require_relative './game'
+require_relative './player'
+
+class Rps < Sinatra::Base
+
```
tansaku: please ensure to remove excess empty lines from places like this

views/layout.erb
```diff
@@ -0,0 +1,13 @@
+<head>
+  <title>Rock, Paper, Scissors</title>
+  <link rel="stylesheet" type="text/css" href="/public/stylesheet.css">
+  <link href='http://fonts.googleapis.com/css?family=Quicksand' rel='stylesheet' type='text/css'>
+</head>
+
+<body>
+  <h1> Rock, Paper, Scissors </h1>
+  <img src="./public/Images/rps.png" alt="picture">
```
tansaku: generally we'd have this directory be in lower case, i.e. 'images'

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
...
+    session[:game] = Game.new
+    session[:player] = Player.new("Yannick")
+    session[:game].add(session[:player])
+    erb :index
+  end
+
+  get '/game' do
+    choice = params[:choice]
+
+
```
tansaku: don't like the empty space here

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
...
+    if choice.nil?
```
tansaku: overall this method feels rather long, i.e. too much logic in your controller - and also I see too much of the view in here too - see comments below

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
...
+      @outcome = "To get started #{session[:player].name} pick a weapon"
+      erb :play_game, locals: {outcome: @outcome, player_score: @player_score, computer_score: @computer_score}
+    else
+      computer_choice = session[:game].get_random_choice
+      win_lose_draw = session[:game].outcome?(choice.to_sym,computer_choice)
+      session[:game].assign_score(win_lose_draw)
+      @outcome = "The computer played the #{computer_choice}, you #{win_lose_draw.to_s}"
```
tansaku: rather than formatting these strings in your controller you should have them in the view and be passing the relevant variables to the view to have the details interpolated there

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
...
+    session[:game].add(session[:player])
+    erb :index
+  end
+
+  get '/game' do
+    choice = params[:choice]
+
+
+    if choice.nil?
+      @outcome = "To get started #{session[:player].name} pick a weapon"
```
tansaku: don't have interface strings in your controller - move them to view

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
...
+      erb :play_game, locals: {outcome: @outcome, player_score: @player_score, computer_score: @computer_score}
```
tansaku: I don't think you need this locals - instance variables will be automatically available under those names in your view

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
...
+    else
+      computer_choice = session[:game].get_random_choice
```
tansaku: this functionality could be pushed  into the player or game

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
...
+      win_lose_draw = session[:game].outcome?(choice.to_sym,computer_choice)
+      session[:game].assign_score(win_lose_draw)
+      @outcome = "The computer played the #{computer_choice}, you #{win_lose_draw.to_s}"
+      @computer_score = "Computer score: #{session[:game].computer_score}" 
+      @player_score = "Player score: #{session[:game].player_score}"
+      erb :play_game, locals: {outcome: @outcome, player_score: @player_score, computer_score: @computer_score}
```
tansaku: again I don't think you need this locals declaration

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+  end
```
tansaku: overall this method is much too long - you need to pull out the game logic into the other classes you have, get the view stuff in the view etc.

lib/game.rb
```diff
@@ -0,0 +1,47 @@
+require_relative './player'
+
+class Game
+
+  attr_accessor :player,:computer_score, :player_score
```
tansaku: do you really need attr_accessor for all these?  Seems like overkill

lib/game.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def ready?
+    has_a_player?
+  end
+
+  def assign_score(outcome)
+    outcome == :win ? @player_score += 1 : (outcome == :lose ? @computer_score += 1 : nil)
+  end
+
+  def outcome?(player_choice,rand_choice)
```
tansaku: this could be shortened with quick returns or ternary arguments, or even drop these symbols into the main hash and just return based on a double lookup

lib/player.rb
```diff
@@ -0,0 +1,9 @@
+class Player
+  attr_reader :name 
+  attr_accessor :choice
```
tansaku: do you really need accessor here?

tansaku:
Thanks, shame that you didn't finish the cucumber features.  Have another go at them, making sure that you see them fail appropriately by breaking them in retrospect.  Also note that the empty stylesheet.css should be removed if you are not using it, and that we would usually prefer lower case for the directories, i.e. public/images - you'll run into problems deploying on linux mixing up the case like that.  Great that you have a working game, but you need to clear up your controller and get view logic in views, game logic in the game, and ensure that you aren't exposing any more object state than is really necessary. :finished:

spike01:
:checkered_flag:

PR #10, opened 2015-02-23
C Christenson RPS-Challenge

c-christenson:


Gemfile
```diff
@@ -0,0 +1,16 @@
+# A sample Gemfile
+source "https://rubygems.org"
+
+# gem "rails"
```
MihaiLiviuCojocar: delete unnecessary comments...

README.md
```diff
@@ -1,31 +1,11 @@
...
-
-## Basic Rules
-
-- Rock beats Scissors
-- Scissors beats Paper
-- Paper beats Rock
+Further functionality to implement:
+- Only allow the user to enter the three choices (Rock, Paper, or Scissors). Right now the user can type anything into the submit field.
+- Add styling and images to the pages.
+- Add a play again button on the results page.
```
MihaiLiviuCojocar: If you deployed, put the link to the live app ...

features/starting_a_game.feature
```diff
@@ -0,0 +1,29 @@
+Feature: Starting the game
```
MihaiLiviuCojocar: INDENTATION is very important!!
e.g.:
Feature: Starting the game
  In order to play rock paper scissors
  As a marketeer at Makers Academy
  I want to start a new game

  Scenario: Enter Name
    Given I am on the homepage
    And I enter a name into the form
    When I click "submit"
    Then I should be taken to a new page "What would you like to play?"

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,47 @@
...
+
+Given(/^I do not enter a name into the form$/) do
+  fill_in('player_name', :with => '')
+end
+
+Then(/^I should stay on the homepage "(.*?)"$/) do |text|
+  expect(page).to have_content text
+end
+
+Given(/^I am on the game page$/) do
```
MihaiLiviuCojocar: INDENTATION, again! Make sure you use two spaces for one tab!

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+			when @players[1].weapon == :Rock && @players[0].weapon == :Scissors
+				return @players[1]
+			when @players[1].weapon == :Paper && @players[0].weapon == :Rock
+				return @players[1]
+			when @players[0].weapon == @players[1].weapon
+				return nil
+		end
+	end
+
+
```
MihaiLiviuCojocar: INDENTATION and spaces between lines! Try to be consistent! :)

spec/player_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+		expect(player.name).to eq "Carrie"
+	end
+
+	it 'can pick a weapon' do 
+		player.pick(:paper)
+		expect(player.weapon).to eq :paper
+	end
+
+
+end
```
roidriscoll: As Mihai has flagged, your indentation needs attention. If you are using Sublime Text, go to Preferences/Settings - User and put the following in that hash:
```
	"tab_size": 2,
	"translate_tabs_to_spaces": true,
```

lib/player.rb
```diff
@@ -0,0 +1,14 @@
+class Player
+
+attr_accessor :name
+attr_accessor :weapon
```
MihaiLiviuCojocar: There is no need for an accessor for @weapon, because the pick() method sets the value of @weapon. Try not to write any code that doesn't get used or is unnecessary!

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+				return @players[1]
+			when @players[1].weapon == :Rock && @players[0].weapon == :Scissors
+				return @players[1]
+			when @players[1].weapon == :Paper && @players[0].weapon == :Rock
+				return @players[1]
+			when @players[0].weapon == @players[1].weapon
+				return nil
+		end
+	end
+
```
roidriscoll: While it makes sense why you have chosen to use a case statement, there is usually a better way. This is the case in this instance also. If we were following SOLID principles, we might make an object for each weapon, and each object might know which other object it will defeat, and be defeated by. As another example, you could alternatively make a hash which has corresponding key/value pairs which also state their weakness, and strengths.

lib/game.rb
```diff
@@ -0,0 +1,35 @@
+class Game
+
+attr_accessor :players
```
roidriscoll: Here you should really use `attr_reader` as your `add_player` method takes care of the writing to `@players`

server.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  post '/' do
+    if params[:player_name].empty?
+      @name = "Please provide your name to play the game."
+      erb :index
+    else
+      player.name=(params[:player_name])
+      game.add_player(player)
+      game.add_player(computer)
+      session[:me] = player
```
MihaiLiviuCojocar: Carefull when you add objects to session... it might become too big!

server.rb
```diff
@@ -0,0 +1,58 @@
...
+    else
+      player.pick(params[:player_choice])
+      session[:me] = player
+      @player_weapon = session[:me].weapon
+      computer.pick_weapon
+      @computer_weapon = computer.weapon
+      @result = "You lose!" if game.winner == computer
+      @result = "You win!" if game.winner == player
+      @result = "It's a draw!" if game.winner == nil
+      erb :result
```
MihaiLiviuCojocar: You have logic in the controller. This can be and should be moved in the Game class.
Imagine you go to a restaurant. Let's say the view is the customer, the waiter is the controller and the model(Game class) is the cook. If we have this kind of logic in the controller is like the waiter(controller) would have to cook(make some logic) the food the the customer(view) needs. You could have a method in the Game class that could provide a message, and you could use it in the controller like this: @result = game.result_message().

spec/player_spec.rb
c-christenson: Ah thank you! My code looks fine from my end, but I guess github has been using tabs instead of 2 spaces. I promise I'm not indenting it like crazy from my end!

views/result.erb
```diff
@@ -0,0 +1,4 @@
+<div>You picked <%= @player_weapon %></div>
+<div ="/game" method="post"></div>
+<div>The computer picked <%= @computer_weapon %></div>
+<div><%= @result %></div>
```
MihaiLiviuCojocar: :checkered_flag:`

Gemfile
c-christenson: Done

README.md
c-christenson: Link added

features/starting_a_game.feature
c-christenson: Fixed!

lib/game.rb
c-christenson: fixed.

server.rb
c-christenson: fixed!

spike01:
:checkered_flag:

PR #11, opened 2015-02-23
Not particularly beautiful, but there is a nice little functionality

TStrothjohann:
If someone found out, why I have to restart the server after every round in order to keep playing... I would pretty much appreciate. Added some more comment to README.md


app/server.rb
```diff
@@ -0,0 +1,61 @@
...
+    @@playertwo = Player.new("Computer")
+    RPS.add_player(@@playertwo)
+    erb :game
+  end
+end
+
+post '/game' do
+  @gameover = RPS.over?
+  @weapon = params[:weapon]
+  @@playerone.choose_weapon(@weapon)
```
pitchinvasion: You shouldn't be using class variables here, use instance variables instead, this might be why you have to restart the server every time, although I can't be sure

app/server.rb
```diff
@@ -0,0 +1,61 @@
...
+  @@playertwo.choose_random_weapon
+  #Decides who won or tell if it was a tie:
```
pitchinvasion: There's two much logic in the controller here you could definitely move a lot of this logic into objects where you can unit test them

app/server.rb
```diff
@@ -0,0 +1,61 @@
...
+  @beater = RPS.beater?(@@playertwo, @@playerone)
+  if @beater == "Tie. Choose again"
+    @tie = "Tie. Choose again"
```
pitchinvasion: You shouldn't be putting strings in your controllers, anything to do with presentation should go into your views

app/views/game.erb
```diff
@@ -0,0 +1,28 @@
+<h1>Rock, Paper, Scissors</h1>
+
+<!-- <p><i>Staring <b><%= @players %></b> against <b><%= @players %></b> </i></p>  -->
```
pitchinvasion: Delete comments before committing

app/views/game.erb
```diff
@@ -0,0 +1,28 @@
...
+
+<% if !@gameover %>
```
pitchinvasion: space words with an underscore in variables

app/views/game.erb
```diff
@@ -0,0 +1,28 @@
...
+  
+  <p><%= @countmessage %></p>
+  <p style="color: red" ><%= @tie %></p>
+  
+
+  <p>Hello <%= session[:playername] %>, please choose one of the following weapons:</p>
+
+  <form action="/game" method="POST" id="rps">
+    <input type="radio" name="weapon" value="rock" id="rock" >Rock
+    <br>
```
pitchinvasion: See if you can work out a more semantic way of marking up your form without using `<br>` 

https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Forms/How_to_structure_an_HTML_form

features/rps-start.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Registration
```
pitchinvasion: Nice work splitting these features into two, definitely makes sense to have two seperate features like you've done here

features/step_definitions/rps_steps.rb
```diff
@@ -0,0 +1,64 @@
...
+  visit('/register')
+  fill_in('playername', :with=>'Thomas') 
+  click_button('Start Game')
+  
+  find('form#rps')
+  choose('rock')
+  click_button('Choose')
+
+  # find('form#rps')
+  # choose('scissor')
```
pitchinvasion: Delete comments!

lib/game.rb
```diff
@@ -0,0 +1,37 @@
+class Game
+  attr_accessor :players
+  attr_reader :count 
+  
+  def initialize
+  @players = []
+  @beats = { paper: :scissor, rock: :paper, scissor: :rock}
```
pitchinvasion: I would put this as a constant, as you shouldn't be planning to change this variable

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+  @weapons = @beats.keys
```
pitchinvasion: Rather than have this as an instance variables you could create a method that does this

```
def weapons
  @beats.keys
end
```

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def add_player(name)
+       @players << name 
+  end
+
+  def beater?(player2, player1)
+
+    if player1.weapon == player2.weapon
+      "Tie. Choose again"
```
pitchinvasion: It's better to return symbols here rather than strings, and then add strings in the views with the actual text you want

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+    elsif player1.weapon == @beats[player2.weapon]
+      @turns << player1.name
```
pitchinvasion: Is there a way you could do this on fewer lines?

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+      player1
+    else
+      @turns << player2.name
+      player2
+    end
+  end
+
+  def count
+    @count = @turns.each_with_object(Hash.new(0)) { |player,beats| beats[player] += 1 }
```
pitchinvasion: This is nice that you've got it on one line. But you don't need to set it to an instance variables, and I would look at how complex this method is and think that there is probably a simpler way of working out the number of turns for each player without having to do this. Maybe if you had a method that could take a player as an argument, and then instead of doing this you could just do

```
def count(player)
  @turns.count { |turn| turn == player }
end
```

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def over?
+    @turns.length < 2 ? false : true
```
pitchinvasion: A ternary that returns a boolean value is a code smell, you should do something like 

```
@turns.length >= 2
```

lib/player.rb
```diff
@@ -0,0 +1,16 @@
+class Player
+
+  attr_reader :name, :weapon
+  def initialize(name)
+    @name=name
+  end
+
```
pitchinvasion: Nice although if you wanted to you could even shorten the method names to `weapon` `random_weapon`

spec/game_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    expect(game.beater?(thomas, thomas_alias)).to eq("Tie. Choose again")
+  end
+
+  it "should keep track of the turns" do
+    game.beater?(thomas, computer)
+    game.beater?(thomas, computer)
+    expect(game.count[:thomas]).to eq(2)
+  end
+
+  it "should know that the geme is over after three beats" do
```
pitchinvasion: spelling!

spec/game_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    game.beater?(thomas, computer)
+    game.beater?(thomas, computer)
+    game.beater?(thomas_alias, thomas)
+    game.beater?(thomas, computer)
+    expect(game).to be_over
+  end
+
+  it "shouldn't break when asked over? and no turn has happend" do
```
pitchinvasion: you don't normally need this sort of test, you could say "when there are still games left the game is not over"

spec/player_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'player'
+
+describe Player do
+
+  let(:player){Player.new("Thomas")}
+  let(:weapon){double :weapon, choose_weapon: :scissor}
+
+  it "should have a name" do
```
pitchinvasion: try and avoid saying should over and over, you can just say "has a name"

pitchinvasion:
Good to see you've got everything tested and pretty much working, in terms of improvement you could possibly extract more of the logic that is currently in your controller into your objects or views (remember that views are where your presentation should be, your controllers should be as thin as possible and everything else nicely encapsulated inside your objects) and make your methods as small and tidy as possible (while retaining readability). Also remember sometimes the simpler approach is the cleaner approach, but this is a good approach. 

Also if you're needing to restart the server, you should look at using sessions to keep track of your objects - although don't put your actual objects in the session as this is considered bad practice for numerous reasons including that there is a limit to the memory size of sessions. Consider using object_id to keep track of your objects and storing them inside a hash in Sinatra. :checkered_flag:

TStrothjohann:
Hi Leo,
thank you for the detailed code review. I really appreciate. I have tried
to refactor everything you pointed out. I think it looks better now.

I resolved the server-breakdown-issue by putting the constant that contains
the game into the post route. It is reset when the player registers now.
Cucumber still doesn't like it very much though.

I haven't found a way of making the beater?-method much slimmer as you
suggested. It's maybe one line shorter now.

Cheers,
Thomas

2015-02-23 14:37 GMT+00:00 Leo Allen <notifications@github.com>:

> Good to see you've got everything tested and pretty much working, in terms
> of improvement you could possibly extract more of the logic that is
> currently in your controller into your objects or views (remember that
> views are where your presentation should be, your controllers should be as
> thin as possible and everything else nicely encapsulated inside your
> objects) and make your methods as small and tidy as possible (while
> retaining readability). Also remember sometimes the simpler approach is the
> cleaner approach, but this is a good approach.
>
> Also if you're needing to restart the server, you should look at using
> sessions to keep track of your objects - although don't put your actual
> objects in the session as this is considered bad practice for numerous
> reasons including that there is a limit to the memory size of sessions.
> Consider using object_id to keep track of your objects and storing them
> inside a hash in Sinatra. [image: :checkered_flag:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/makersacademy/rps-challenge/pull/11#issuecomment-75550900>
> .
>



-- 
Thomas Strothjohann
Journalist + Fotograf

Sachsenhäuser Landwehrweg 225A
60598 Frankfurt am Main
+49-6959606923
+44-7542902955
www.strothjohann.net

PR #12, opened 2015-02-23
Eddy B's attempt at the RPS challenge. 

eddbrown:
I haven't got to Heroku yet. The code seems to work in IRB, but trying to get the flow of information right using Sinatra is difficult.


features/step_definitions/rps_steps.rb
```diff
@@ -0,0 +1,26 @@
...
+  pending # express the regexp above with the code you wish you had
+end
+
+When(/^I choose the correct answer$/) do
+  pending # express the regexp above with the code you wish you had
+end
+
+Then(/^I should either win or lose$/) do
+  pending # express the regexp above with the code you wish you had
+end
```
spike01: If you aren't using the steps, don't leave them in. Also, your Cucumber tests don't run

lib/game.rb
```diff
@@ -0,0 +1,39 @@
+class Game
+
```
spike01: INDENTATION

lib/game.rb
```diff
@@ -0,0 +1,39 @@
...
+	attr_reader :players
+
+	def initialize
+		@players = []
+		# @turn = false
```
spike01: Don't leave comments in

lib/server.rb
```diff
@@ -0,0 +1,56 @@
+require 'sinatra/base'
+require '~/Projects/rps-challenge/lib/game.rb'
+require '~/Projects/rps-challenge/lib/player.rb'
+require '~/Projects/rps-challenge/lib/rps.rb'
```
spike01: These `require` statements are not gonna work out - they should read more like `require './lib/game'`

lib/server.rb
```diff
@@ -0,0 +1,56 @@
...
+		player1.choose_rock if session[:p1] == 'Rock'
+
+		player2.choose_scissors if session[:p2] == 'Scissors'
+		player2.choose_paper if session[:p2] == 'Paper'
+		player2.choose_rock if session[:p2] == 'Rock'
+
+		@output = game.winner
+
+		erb :game_winner
+	end
```
spike01: All of this logic should be encapsulated in your objects - not in the controller

spec/game_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  	game.add(player_loser)
+  	game.add(player_winner)
+  	expect(game.winner).to eq('player 2 wins!')
+  end
+
+  it 'should recognise a draw' do
+  	game.add(player_draw)
+  	game.add(player_draw)
+  	expect(game.winner).to eq('draw!')
+  end
```
spike01: These two test fail because you haven't mocked `rps_holder` on your players

lib/player.rb
```diff
@@ -0,0 +1,25 @@
...
+
+	def choose_scissors
+		@rps_holder.change_to_scissors
+	end
+
+	def choose_rock
+		@rps_holder.change_to_rock
+	end
+
+end
```
spike01: Would it not be nicer to have a method that takes an argument (the rock, paper or scissors selection) and makes the change based on what argument is passed in?

lib/rps.rb
```diff
@@ -0,0 +1,21 @@
...
+
+	def change_to_scissors
+		@content = 'scissors'
+	end
+
+	def change_to_rock
+		@content = 'rock'
+	end
+
+end
```
spike01: These three methods could be reduced to one, as mentioned above

views/game_winner.erb
```diff
@@ -0,0 +1,3 @@
+<h1><%= @output %></h1>
+
+<a href="http://localhost:9292/player1choose">New Game?</a>
```
spike01: You shouldn't be hard-coding the start of the link - this will fail when deployed. The link should go to `/player1choose`

views/new_game2.erb
```diff
@@ -0,0 +1,16 @@
+<h1>Player 2 choose</h1>
+
+<form action="/player2choose" method= 'post'>
+    <input type="submit" value="Rock" name="Rock">
+    <% p2choice = 'Rock' %>
```
spike01: This doesn't do anything - also, this file isn't being used by your server, so should be deleted

views/p1choose.erb
```diff
@@ -0,0 +1,14 @@
...
+    <br>
+    Scissors : <input type="radio" name= "weapon" value= "Scissors">
+    <br>
+    <input type="submit">
+</form>
+
+
+
+
+
```
spike01: Indentation, trailing whitespace

app/lib/rps.rb
```diff
@@ -0,0 +1,21 @@
+class Rps
+
+	attr_accessor :content
+
+	def initialize
+		@content = 'lizard'
+	end
```
spike01: Would you want this to always start as lizard? You haven't implemented the other parts of the rock/paper/scissors/spock/lizard

spike01:
The code works (once those requires are sorted out) which is good, but start to think about how you can refactor:

- take as much logic out of your controllers
- consider using a hash to lookup the results
- if you have methods that are extremely similar except for one changing part (your `change_to_[something]` methods, for example) you could condense them into one:

```ruby
def change_to(choice)
  @content = choice
end
```

:checkered_flag:

PR #13, opened 2015-02-23
single player game works with option to play game again

wardymate:
Cucumber test failing and lots of refactoring needed will continue to work on and expand the features further.

https://pure-brook-4821.herokuapp.com/


spec/game_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  it 'should not be able to add more than 2 players' do
+    2.times{game.add_player(player1)}
+    expect{game.add_player(player1)}.to raise_error(RuntimeError, 'Only 2 players allowed')
+  end
+
+  it 'should know who has won' do
+    game.add_player(player1)
+    game.add_player(player2)
+    expect(game.result).to eq player2
```
stephenlloyd: This is failing. It's returning draw rather than player2.

lib/weapon.rb
```diff
@@ -0,0 +1,21 @@
+class Weapon
+
+  attr_accessor :type, :paper, :scissors, :rock
+
```
stephenlloyd: interesting approach. Do rock, paper and scissors have to be accessors. I think you only really need one attr_reader of type. 

Also in the this class you're doing essesntially the same thing twice. '@type = __method__' would also work. Can you think of a way to refactor this down to one method??? It's a bit of a challenge for you but it is possible to use one method, one constant and one attr_reader. :) If you need a hand let me know.

lib/game.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def add_player(player)
+    raise 'Only 2 players allowed' if @players.length == 2
+    @players << player
+  end
+
+#neeed to find a way to refactor the method below as it is far too big
+
+  def result
```
stephenlloyd: You're right this does need to be shorter. Maybe the first line should be   
```ruby
return "draw" if players.first.weapon_choice == players.last.weapon_choice
```

lib/game.rb
```diff
@@ -0,0 +1,44 @@
...
+    if @players.first.weapon_choice == "scissors"
+      if @players.last.weapon_choice == "rock"
```
stephenlloyd: so you're saying rock beats scissors, papers beats... that could look like this 
```ruby 
winners = {"rock" => "scissors" }
```
this way the you can do something like 
```ruby
winners[@players.first.weapon_choice] == @players.last.weapon_of_choice
```
this would mean that the first player beat the second player
```

File: lib/player.rb [hunk 5]
```diff
@@ -0,0 +1,9 @@
+class Player
+
+  attr_accessor :name, :weapon_choice
+
+  def initialize(arguments_hash)
+    @name = arguments_hash[:name]
```
stephenlloyd: nice, good job on taking tips from poodr/jakes talk

server.rb
```diff
@@ -0,0 +1,70 @@
...
+    if params[:weapon_choice] == "no_weapon"
+      erb :game
+    else
+      @player_weapon = params[:weapon_choice]
+      @player_name = params[:player_name]
+
+      human = session[:human]
+
+      if @player_weapon == "rock"
+        human.weapon_choice = Weapon.new.rock
```
stephenlloyd: Theres a lot of logic here. I would want to see this in the game class

server.rb
```diff
@@ -0,0 +1,70 @@
...
+         human.weapon_choice = Weapon.new.scissors
+      end
+
+      @game = session[:game]
+      computer = @game.players.select{|player| player.name == "Computer"}.first
+      @computer_weapon = computer.weapon_choice
+
+      winner = @game.result
+      if winner == "draw"
+        @winner = "It's a " + winner +"!"
```
stephenlloyd: winner is used a lot here I would have a variable of @winning_message to make it a bit more clear. You could probably use a ternary to condense this to one line making the initial winner variable redundant.

server.rb
```diff
@@ -0,0 +1,70 @@
...
+    else
+      @player_weapon = params[:weapon_choice]
+      @player_name = params[:player_name]
+
+      human = session[:human]
+
+      if @player_weapon == "rock"
+        human.weapon_choice = Weapon.new.rock
+      elsif @player_weapon = "paper"
+         human.weapon_choice = Weapon.new.paper
```
stephenlloyd: as it stands with type being a accessor you could just do human.weapon_choice = @player_weapon. Not only would this save about 10 lines it will also make you rethink the purpose of your weapon class.

spec/player_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'player'
+
+describe Player do
+
+  player = Player.new({:name => "Chris"})
+  let(:paper) {double :weapon}
+
+  it 'should be bale to set its name' do
+    expect(player.name).to be_a String
```
stephenlloyd: checking to see if it was equal to "Chris" would have been a better test

views/result.erb
```diff
@@ -0,0 +1,15 @@
+<h1>The result is ........<h1>
+
+<weapon> You chose <%= @player_weapon %> </weapon><br>
```
stephenlloyd: Whilst we want semantic html. This will get converted to just a div. You should look through the html 5 tags to see what would have been most relevant.

views/result.erb
```diff
@@ -0,0 +1,15 @@
...
+<weapon> The computer chose <%= @computer_weapon %></weapon><br>
+
+<result> <%= @winner %></result><br>
+
+<h3>
+  <form action="/game" method="post">
+    <question>Would you like to play again <%= @player_name %>?</question><br>
+    <input style="display: none" type="hiddden" value="<%= @player_name %>" name="player_name">
```
stephenlloyd: as it stands when you play again the computer will keep choosing it's first choice. Can you make it different each time?

stephenlloyd:
All and all it's pretty good. I've listed some improvements to make it better and neaten it up which I think you should definitely try to take on board. Especially the results method in game. You'll get the sticker for this and well done for getting it live. :checkered_flag:

PR #14, opened 2015-02-23
nearly got a two player game working

veliancreate:
sorry its late...


features/play_game.feature
```diff
@@ -0,0 +1,28 @@
...
+		And I choose the two player button
+		And I click on the submit button
+		When I enter "Julian" in one of the fields 
+		When I enter "John" in the other field
+		And I click on the player choice page
+		Then I should see "Player 1 will use" 
+
+	Scenario: Playing a two player game
+		Given I am on the one player game page
+
```
roidriscoll: Just a small note here, but an important one - your indentation __pattern__ is correct but the values are too much. If you are using Sublime Text, go to Preferences/Settings - User and put the following in that hash:
```
"tab_size": 2,
"translate_tabs_to_spaces": true,
```

lib/game.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def determine_winner(player)
+    if selections_equal?(player) 
+      self.selection
+    else
+       check_selections(player)           
+    end   
+  end
+
```
roidriscoll: You could condense `determine_winner` nicely by using a [ternary operator](http://stackoverflow.com/questions/4252936/how-do-i-use-the-conditional-operator-in-ruby)

lib/game.rb
```diff
@@ -0,0 +1,49 @@
...
+      when :scissors
+        if player.selection==:rock
+          player
+        elsif player.selection==:paper
+          self
+        end       
+    end   
+  end  
+end  
+
```
roidriscoll: While it makes sense why you have chosen to use a case statement, there is usually a better way. This is the case in this instance also. If we were following SOLID principles, we might make an object for each weapon, and each object might know which other object it will defeat, and be defeated by. As another example, you could alternatively make a hash which has corresponding key/value pairs which also state their weakness, and strengths.

lib/player.rb
```diff
@@ -0,0 +1,18 @@
+class Player
+
+  attr_accessor :selection, :name
+
+  def initialize
+    @selection=nil
+    @name = nil
+  end
```
roidriscoll: Instance variables are automatically assigned `nil`, so line 7 is unnecessary

lib/player.rb
```diff
@@ -0,0 +1,18 @@
+class Player
+
+  attr_accessor :selection, :name
+
```
roidriscoll: You are assigning @selection on line 11, so do you need to use `attr_accessor`? If no write access is required elsewhere you should use `attr_reader`. This will ensure that you are not creating unncessary access levels.

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,160 @@
...
+
+  enable :sessions
+
+  PLAYER=Player.new
+  PLAYER_1=Player.new
+  PLAYER_2=Player.new
+  TWO_PLAYER_GAME=TwoPlayerGame.new
+  TWO_PLAYER_GAME.player_one=PLAYER_1
+  TWO_PLAYER_GAME.player_two=PLAYER_2
+
```
roidriscoll: Be consistent with your whitespace. You should always have a space before and after your `=`. So `PLAYER=Player.new` should be `PLAYER = Player.new`.

lib/game.rb
```diff
@@ -0,0 +1,49 @@
...
+  attr_reader :selection
+
+  def initialize
+    @selection = [:rock,:paper,:scissors].sample
+  end  
+
+  def selection
+    @selection
+  end  
+
```
roidriscoll: The `selection` method is redundant, due to the `attr_reader` on line 3

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,160 @@
...
+    @player = PLAYER_1
+    @opponent = PLAYER_2
+  end  
+
+  def player_two_setup
+    @form_direct_string = 'player_2_game'
+    @player = PLAYER_2
+    @opponent = PLAYER_1
+  end  
+
```
roidriscoll: We want to keep as much logic as possible out of controllers. In order to do this, try and think about redesigning this, so that all of the logic is encapsulated inside of objects, each of which have a single responsibility for handling the logic of the game.

views/two_player_game.erb
```diff
@@ -0,0 +1,30 @@
...
+		    Go Rock!<input id="rock" type="radio" name="selection" value="rock">
+		    Go Paper!<input id="paper" type="radio" name="selection" value="paper">
+		    Go Scissors!<input id="scissors" type="radio" name="selection" value="scissors">
+			<input type="submit" name="go" value="Go!">
+		</form>
+	 
+	 	<form action='/' method='get'>
+	 		<input type="submit" name="quit" value="Quit current game">	
+	 	</form>
+	 <% end %>
```
roidriscoll: It seems unnecessary to have a form here. There's nothing clearing out evidence of the current game, so surely a simple link to the homepage would suffice? That way you wouldn't end up with unsightly query strings in the address bar.

roidriscoll:
I think the most useful overall feedback here would be to avoid putting logic in your controllers as much as possible. It would be worth you revisiting this in future and having another go at the challenge with this in mind. :checkered_flag:

PR #15, opened 2015-02-23
rock, paper, scissors

loris-fo:


app/server.rb
```diff
@@ -0,0 +1,46 @@
...
+    erb :new_game
+  end
+
+  get '/player' do
+    @name = params[:name]
+    erb :player
+  end
+
+
+    post '/player' do
```
silvabox: indentation incorrect.

app/server.rb
```diff
@@ -0,0 +1,46 @@
...
+class RockPaperScissors < Sinatra::Base
+  set :views, Proc.new { File.join(root, "..", "views") }
+  enable :sessions
+
+  get '/' do
+    erb :new_game
+  end
+
+  get '/player' do
+    @name = params[:name]
```
silvabox: What is `params[:name]` on a get request?

app/server.rb
```diff
@@ -0,0 +1,46 @@
...
+    erb :arena
+  end
+
+  get '/arena' do
+    @name = params[:name]
+    session[:player] = @player
+    erb :arena
+  end
+
+  post '/game' do
```
silvabox: incorrect indentation

app/server.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  get '/player' do
+    @name = params[:name]
+    erb :player
+  end
+
+
+    post '/player' do
+    @name = params[:name]
+    session[:player] = @player
```
silvabox: What is @player in this context?  Has it been initialized?

features/step_definitions/choosing_weapon_step.rb
```diff
@@ -0,0 +1,11 @@
+Given(/^I am on the arena page$/) do
+  visit '/arena'
+end
+
+When(/^I type "(.*?)"$/) do |weapon|
```
silvabox: This will match any given/then/when statement `When I type "something"` but you are filling in a specific field with a specific value.  You should not use a parameterized matcher here, or if you do, parameterize the chose weapon.  For example `When I choose "rock" as the weapon` with the matcher `When(/^I choose "(.*?)" as the weapon$/) do |weapon|`

features/step_definitions/choosing_weapon_step.rb
```diff
@@ -0,0 +1,11 @@
...
+  fill_in('weapon', :with => 'rock')
+end
+
+Then(/^I should see the result$/) do
+  visit '/result'
```
silvabox: There is no expectation here - you are just redirecting to the result page.  The `Then` section is the test - so you must include an expectation otherwise you are not testing anything.

features/step_definitions/registering_step.rb
```diff
@@ -0,0 +1,11 @@
+Given(/^I am on the player page$/) do
+  visit '/player'
+end
+
+When(/^I enter "(.*?)"$/) do |name|
+  fill_in('name', :with => 'makers')
+end
+
+Then(/^I should see the arena$/) do
+  visit '/arena'
```
silvabox: As with another comment, there is no expectation here - you are just redirecting to the arena page. The `Then` section is the test - so you must include an expectation otherwise you are not testing anything.

lib/game.rb
```diff
@@ -0,0 +1,34 @@
+class Game
+
+attr_reader :players, :winning, :possibilities
+
+  def initialize
+    @players = []
+    @winning = {"rock" => "scissors",
```
silvabox: `@winning` should be a constant rather than an instance variable as it never needs to change and it is the same for all instances of game.

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+      "scissors" => "paper",
+      "paper" => "rock"
+    }
+    @possibilities = possibilities
```
silvabox: This line does nothing.  `possibilities` is a reader method which is included due to `attr_reader :possibilities`, which returns the value of `@possibilities`.  So this line is the equivalent of setting `@possibilities` to itself.

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def add_player(player)
+    @players << player
+  end
+
+
+  def winner?(players)
```
silvabox: Why are you passing `players` as a parameter to `winner?`?  `players` is already accessible to the game instance as it is an attribute, so passing it in here is misleading.

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+      return "win"
+    elsif players[0].choice == winning[players[1].choice]
+      return "lost"
+    else
+      return "draw"
+    end
+  end
+
+  def random_answer
+    possibilities = ["rock", "paper", "scissors"].shuffle
```
silvabox: Every call to `random_answer` will cause an 2 array objects and three string objects to be instantiated.  Overcome this by defining `["rock", "paper", "scissors"]` as a constant on the class.

lib/marketer.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def initialize
+    @name = []
+  end
+
+  def add_name(name)
+    @name << name
+  end
+
+  def chosen_option(weapon)
```
silvabox: You have defined an `attr_reader :choice` but created a `chosen_option` method which sets the instance variable used by `choice`.  This is inconsistent and unnecessary.  Simply define `attr_accessor :choice` then you can call `choice = "rock"` for example.

lib/marketer.rb
```diff
@@ -0,0 +1,18 @@
+class Marketer
+
+  attr_reader :name, :choice,
+
+  def initialize
+    @name = []
```
silvabox: Why is name an array?  You have an `attr_reader` on `:name` which is confusing.  Name is singular so only one name is expected.  If you intent to return an array, then it should be called `names`

lib/marketer.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def add_name(name)
```
silvabox: It's unclear why you are adding a name instead of setting a name - see other comment about the use of an array.

spec/marketer_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    expect(marketer.name).to eq []
+  end
+
+  it 'should be able to add a name to a new player' do
+    marketer.add_name("makers")
+    expect(marketer.name).to eq ["makers"]
+  end
+
+  it 'should have a choice between rock, paper and scissors' do
+    expect(marketer.chosen_option("scissors")).to eq "scissors"
```
silvabox: You are not testing the resulting state of the object here, just the return value of the `chosen_option` method, which is simply the last expression executed within that method.

spec/marketer_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  it 'should have a choice between rock, paper and scissors' do
+    expect(marketer.chosen_option("paper")).to eq "paper"
```
silvabox: As above.

spec/marketer_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  it 'should have a choice between rock, paper and scissors' do
+    expect(marketer.chosen_option("rock")).to eq "rock"
```
silvabox: As above.

views/player.erb
```diff
@@ -0,0 +1,7 @@
+    <h4 class="text-center">What's your name?</h4>
+
+    <form action="/arena" method="get">
+    <input class="center-block" type="text" name="name" placeholder="Required" required>
```
silvabox: Incorrect indentation

views/result.erb
```diff
@@ -0,0 +1,10 @@
+<h3 class="text-center"> Here is your result, <%= @name %> ! </h3>
+<div class="text-center"></div>
+
+<% 3.times do %><br><% end%>
```
silvabox: Don't do this - it doesn't properly show intent.  If the number of breaks is dependent on your application state, then that's OK, but here your just using it to generate 3 `<br>` elements, where actually typing `<br>` 3 times would have been more readable and less typing in any case!
You shouldn't use `<br>` tags to achieve layout - this is the job of CSS.

silvabox:
Good attempt Loris, I can sort of see some of your thinking, but there are clearly misunderstandings on what to test and how to use attributes and methods.  I ran your tests and features and there were a number of failures.  You should not submit code with failing tests and note that if you continue to make changes and push to your repository, these will be included on the pull request.  Grab me if you have any questions.  :checkered_flag:

features/support/new_game.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Starting the game
```
silvabox: Note this file is in the wrong place - it should be under `features` not `features/support`

PR #16, opened 2015-02-23
Not complete...No Heroku...

katebeavis:
This is by no means complete, can add player and make choice.

I was feeling unwell on Sunday so I didn't get much done then, poor excuse I should have worked more on it on Friday. Spent most of my time on weekend on Ruby code and Rspec. Struggling to get cucumber/capybara tests passing.

I'm going to spend today trying to complete it, I will get that Ronin badge!!! (eventually....)


features/starting_a_game.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Starting the game
```
henrygarner: Have you run these recently? I get an ambiguous match for "I click on Submit"

features/starting_a_game.feature
```diff
@@ -0,0 +1,16 @@
...
+    Given I am on the homepage
+    And I type a "name"
+    When I click on "submit"
+    Then I should see "Welcome"
+
+ Scenario: No name given
+ 	Given I am on the player page
+ 	And I don't type a "name"
+ 	When I click on "submit"
+ 	Then I should see "You MUST enter a name!"
```
henrygarner: You don't test the game logic at all?

features/step_definitions/starting_a_game.rb
```diff
@@ -0,0 +1,27 @@
+Given(/^I am on the homepage$/) do
+  visit('/')
+end
+
+And(/^I type a "(.*?)"$/) do |name|
+  fill_in("name", with: "kate")
```
henrygarner: The matcher isn't very desciptive. 'I type a "x"' could mean anything?

features/step_definitions/starting_a_game.rb
```diff
@@ -0,0 +1,27 @@
...
+end
+
+When(/^I click on "(.*?)"$/) do |link|
+  click_on(link)
+end
+
+Then(/^I should see "(.*?)"$/) do |arg1|
+  page.should have_content("Welcome")
```
henrygarner: You don't use `arg1` so the test will always look for `"Welcome"`

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def player_count
+    players.count
+  end
+
+  def winner?(player1, player2)
+  player1.choice == player2.choice ? "It's a draw!" :
+  player1.choice == @rules[player2.choice] ? "Player One has won!" :
+  "Player Two has won!"
```
henrygarner: This nested ternary operator is very hard to read. I can see you tried using if statements below and refactored. I would propose:

```ruby
return "It's a draw!" if player1.choice == player2.choice
return "Player One has won!" if player1.choice == @rules[player2.choice]
"Player Two has won!"
```

In addition you should generally not be returning strings from method calls like this (the game could be played in other languages). Maybe you could return the actual `player` that has won?

That way your view could contain something like the following (pseudocode):
```ruby
<% if game.winner(current_player, opponent) == current_player %>
  <p>Congratulations! You won!</p>
<% end %>
```

lib/player.rb
```diff
@@ -0,0 +1,17 @@
+class Player
+
+  attr_reader :name, :choice
+
+  def initialize(name)
+    @name = name
+  end
+
+  def make_choice(choice)
+    @choice = choice
```
henrygarner: Choice doesn't appear to need to be an instance variable. Do you ever query `player.choice`?

lib/server.rb
```diff
@@ -0,0 +1,50 @@
+require 'sinatra/base'
+require_relative 'player'
+require_relative 'game'
+
+class Rps < Sinatra::Base
+
+	game = Game.new
```
henrygarner: Your game is a local variable so you won't be able to refer to it in your view. `@game` would make it accessible.

lib/server.rb
```diff
@@ -0,0 +1,50 @@
...
+
+	enable :sessions
```
henrygarner: You've enabled sessions but you don't appear to use them?

lib/server.rb
```diff
@@ -0,0 +1,50 @@
...
+
+	get '/' do
```
henrygarner: Indentation... :)

lib/server.rb
```diff
@@ -0,0 +1,50 @@
...
+  	erb :index
+  end
+
+  post '/' do
+  	@name = params[:name]
+    @game = game
+    if params[:name].empty?
+    	@message = "You MUST enter a name!"
```
henrygarner: You create a message in the server, but really this ought to be in the view. The server only needs to determine if there is an error to be reported, your view can decide how to report it. Don't forget the possibility of multiple languages. Also the extra variables in `server.rb` make the logic harder to follow.

lib/server.rb
```diff
@@ -0,0 +1,50 @@
...
+    	erb :index
+    else
+    	@@player1 = Player.new(@name)
```
henrygarner: This could be an opportunity to use the session store. At the moment there can be only 1 `player1` shared by all users of the site, whereas the session could have a different `player1` for every unique user of the site. We'll be showing how to use databases for this in the next week or so, too...

lib/server.rb
```diff
@@ -0,0 +1,50 @@
...
+      game.add_player(@@player1)
+      @@player2 = Player.new("AI")
+      game.add_player(@@player2)
+      puts @game.inspect
```
henrygarner: I can see you've been experimenting here... Did you discover what you wanted to know?

lib/server.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  get '/player' do
+    erb :player
+  end
+
+  post '/player' do
+  	@choice = params[:choice]
+  	@@player1.make_choice(@choice)
+  	@@player2.computer_choice
```
henrygarner: Nicely encapsulated computer_choice!

henrygarner:
I'm sorry to hear you were feeling unwell over the weekend.

This is a good start. I hope my comments are useful to help you continue. I've been picky about lots of the details, but without a working game it's easy to get caught up in the detail.

Do come and find me before Thursday this week if you'd like me to take a look again once you've finished.

:checkered_flag:

PR #18, opened 2015-02-23
Diego Romero

jdiegoromero:
I've done almost all the back end, now working on pushing to heroku and doing the html and css. 

the only thing that I havent done is to write the draws in the winning method in the back end.


lib/Game.rb
```diff
@@ -0,0 +1,27 @@
...
+		elsif computer == 'paper' && player == 'rock'
+			'Computer wins!'
+		elsif computer == 'scissors' && player == 'paper'
+			'Computer wins!'	
+		else
+			'It is a Draw'
+		end
+	end
+
+end
```
MihaiLiviuCojocar: INDENTATION! Set the tabs to 2 spaces! :)

lib/Game.rb
```diff
@@ -0,0 +1,27 @@
+class Game 
+
+	attr_accessor :options
```
MihaiLiviuCojocar: This attr_accessor is not used ... And it gets overwritten when you define options() method, anyway.

lib/Game.rb
```diff
@@ -0,0 +1,27 @@
...
+
+	def options
+		@options = %w(scissors paper rock)
+	end
+
```
MihaiLiviuCojocar: This method is not used.

lib/Game.rb
```diff
@@ -0,0 +1,27 @@
...
+			'Computer wins!'
+		elsif computer == 'paper' && player == 'rock'
+			'Computer wins!'
+		elsif computer == 'scissors' && player == 'paper'
+			'Computer wins!'	
+		else
+			'It is a Draw'
+		end
+	end
+
```
MihaiLiviuCojocar: This is actually the only logic that you have in the game. You have no players in the game, there are compared just some strings. If so, this class could be called Rules or something like this.

MihaiLiviuCojocar: This logic is not properly tested. If something works it doesn't mean that you don't have to do it the right way. 
Let's think about this like a doctor not washing his hands before surgery. We have to think that untested code is considered invalid code.

lib/Rps.rb
```diff
@@ -0,0 +1,36 @@
...
+  	game = Game.new
+  	@name = params[:name]
+  	@weapon = params[:weapon]
+  	player_weapon = @weapon
+  	computer_pick = computer.pick
+  	@computer_weapon = computer_pick
+  	@winner = game.winner(player_weapon.to_s.downcase, computer_pick.to_s.downcase)
+  	erb :winner
+  end
+
```
MihaiLiviuCojocar: I like that you didn't do the logic(declaring a winner) here and you passed the responsibility to the model.

spec/computer_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'computer'
+
+describe Computer do
+	
+	it 'should select a random option' do
+		computer = Computer.new
+		expect(computer.pick).to eq computer.answer
+	end
```
MihaiLiviuCojocar: This doesn't make to much sense. I guess that you wanted to test that the computer has a random pick. You could do something like this: 
options = ['rock', 'paper', 'scissors']
expect(options).to include(computer.pick)

spec/game_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+describe Game do
+
+	let(:game){Game.new}
+	let(:player){double :player, name: 'Diego', pick: 'rock'}
+	let(:computer){double :computer, pick: 'scissors'}
+
+	it 'should have a player' do
+		expect(player.name).to eq 'Diego'
+	end
```
MihaiLiviuCojocar: Here you say that the game should have a player, the described class is Game but what is tested is the functionality of the player, which is a double, anyway.

spec/game_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+	it 'should show the options' do
+		expect(game.options).to eq ['scissors', 'paper', 'rock']
+	end
+
+	it 'should have a winner' do
+		expect(game.winner(player.pick, computer.pick)).to eq 'Player wins!'
+	end
```
MihaiLiviuCojocar: "The game should have a winner" it tells me that the game has a player as a winner but game.winner() actually returns just a message and it compares just two strings and it should compare two players.

spec/player_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+	it 'should have player 2' do
+		player.player_1 = 'Jimmy Page'
+		expect(player.player_1).to eq 'Jimmy Page'
+	end
+
+	it 'should be able to select an option' do
+		player.set_pick(game.options[1])
+		expect(player.pick).to eq 'paper'
+	end
+
```
MihaiLiviuCojocar: "Player should have player 1" ?

views/index.erb
```diff
@@ -0,0 +1,18 @@
...
+<link href='http://fonts.googleapis.com/css?family=Fontdiner+Swanky' rel='stylesheet' type='text/css'>
+ <h1>Rock Paper Scissors</h1>
+ <h2>Welcome to the British royal marines</h2>
+     <h2>You ready to fight the pirate machine??</h2>
+     	<br>
+     	<br>
+     	<br>
+     	<br>
+     	<br>
+     	<br>
```
MihaiLiviuCojocar: Does this look DRY? We could set some margin bottom if we want to have some more space between two elements.

views/winner.erb
```diff
@@ -0,0 +1,17 @@
...
+	  	<br>
+	  	<br>
+		  <h3>You have chosen <%= @weapon %></h3>
+		  <h3>the computer chose <%= @computer_weapon %></h3>
+		  <form action='/game'>
+		  	<input type="submit" value="New Game?" id="submit">
+		  </form>
+		 	<audio src="101137__cgeffex__play-ball.wav"  autoplay></audio>
+		 	<audio src="tumblr_lsgp97uZOk1qkn96bo1.mp3"  autoplay></audio>
+	  </body>
```
MihaiLiviuCojocar: There are no acceptance (cucumber) tests. Like I said above, there are two ways to develop software. We try to teach how to craft software. If you choose to do it "cowboy style" you encounter problems maintaining the application and it might be a bit difficult in finding a job because.

:checkered_flag:`

spike01:
:checkered_flag:

PR #24, opened 2015-02-23
rps

Pau1fitz:


features/step_definitions/starting_game.rb
```diff
@@ -0,0 +1,12 @@
+Given(/^I am on the homepage$/) do
+  visit('/')
+end
+
```
danldb: You have the same step definition twice (once in game_setups_steps and again in starting_game), leading to ambiguous matching. ``` Given I am on the homepage ``` always has the definition visit('/') so there is no reason to include it twice.

features/step_definitions/entering_name.rb
```diff
@@ -0,0 +1,11 @@
...
+    visit('./enter_name')
+  end
+
+When(/^I enter my "(.*?)"$/) do |arg1|
+    fill_in('name', :with => 'Paul')
+  end
+
+Then(/^I should go to the "(.*?)" page$/) do |arg1|
+    visit('new_game')
+  end
```
danldb: You are not referring to arg1 at all in your definition, which makes it hard to reuse this step.

lib/computer.rb
```diff
@@ -0,0 +1,16 @@
...
+    number = rand(1..9)
+      if number <= 3
+        @weapon = "Rock"
+      elsif number <=6
+        @weapon = "Paper"
+      else
+        @weapon = "Scissors"
+      end
+  end
+
```
danldb: Pick method could be reduced by using ['rock', 'paper', 'scissors'].sample

lib/game.rb
```diff
@@ -0,0 +1,30 @@
...
+       "#{@player.name} wins"
+     elsif @player.weapon == "Paper" && @computer.weapon == "Rock"
+       "#{@player.name} wins"
+     elsif @player.weapon == "Scissors" && @computer.weapon == "Paper"
+       "#{@player.name} wins"
+     else
+       "Draw"
+     end
+   end
+
```
danldb: Avoid if else. Complex logic like this should be extracted to a data structure such as a hash, that could be a constant at the top of the class.

danldb: Method is WAY too long - keep methods to a maximum of 3 lines and with a single responsibility.

spec/computer_spec.rb
```diff
@@ -0,0 +1,11 @@
+require './lib/computer'
+
+describe Computer do
+
+  it 'can pick a weapon' do
+    computer = Computer.new
+    allow(computer).to receive(:weapon).and_return("Paper")
+    expect(computer.weapon).to eq "Paper"
+  end
```
danldb: This is a vacuous test - you are just testing your stub

spec/game_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  it 'enables two players to join the game' do
+    paul = double(:player)
+    computer = double(:computer)
+    game = Game.new(paul, computer)
+    expect(game.players.count).to eq 2
+  end
+
+  it 'knows if there is a winner' do
+    paul = double(:player, name: "Paul")
+    computer = double(:computer)
```
danldb: use a let statement to avoid repetition

spec/game_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    expect(game.winner).to eq "Paul wins"
+  end
+
+  it 'knows when its a draw' do
+    paul = double(:player, name: "Paul")
+    computer = double(:computer)
+    game = Game.new(paul, computer)
+    allow(paul).to receive(:weapon).and_return "Rock"
+    allow(computer).to receive(:weapon).and_return "Rock"
+    expect(game.winner).to eq "Draw"
```
danldb: use a before block to avoid repetition

app/server.rb
```diff
@@ -0,0 +1,35 @@
...
+  post '/new_game' do
+    @name = params[:name]
+    @weapon = params[:weapon]
+    player.name = @name
+    player.weapon = @weapon
+    computer.pick
+    @winner = game.winner
+    erb :new_game
+  end
+
```
danldb: This controller is too fat - logic should be extracted to the models and classes.

spike01:
:checkered_flag:

PR #27, opened 2015-04-04
RPS Pull Request

sanjsanj:
Barebones have been deployed in Sinatra and Heroku, CSS styling to come.


lib/decide.rb
```diff
@@ -0,0 +1,10 @@
+class Decide
+  def initialize
+    @beats = { "Rock" => "Paper", "Paper" => "Scissors", "Scissors" => "Rock" }
+  end
+  def make p1, p2
+    return "tie" if p1 == p2
+    return "win" if p1 == @beats[p2]
+    return "lose"
```
sanjsanj: Fixed

lib/decide.rb
```diff
@@ -0,0 +1,10 @@
+class Decide
+  def initialize
+    @beats = { "Rock" => "Paper", "Paper" => "Scissors", "Scissors" => "Rock" }
+  end
+  def make p1, p2
```
sanjsanj: Fixed

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,38 @@
...
+Given(/^I am on the homepage$/) do
+  visit '/'
+end
+
+Then(/^I should see "([^"]*)"$/) do |arg1|
+  expect(page).to have_content arg1
+end
+
+Given(/^I fill in "([^"]*)" with "([^"]*)"$/) do |arg1, arg2|
+  fill_in arg1, :with => arg2
```
sanjsanj: Fixed

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,38 @@
...
+
+Then(/^I should be on the "([^"]*)" page$/) do |arg1|
+  expect(current_path).to eq arg1
+end
+
+Then(/^the page should have "([^"]*)" button$/) do |arg1|
+  expect(page).to have_button arg1
+end
+
+
```
sanjsanj: Fixed

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,37 @@
...
+Then(/^the page should have "([^"]*)" button$/) do |arg1|
+  expect(page).to have_button arg1
+end
+
+Given(/^I follow link "([^"]*)"$/) do |arg1|
+  click_link arg1
+end
+
+Given(/^my opponent chooses "([^"]*)"$/) do |arg1|
+  CPU.any_instance.stub(:choice) { arg1 }
```
tansaku: great that you are stubbing this, but as discussed, you should prefer the more modern:

```rb
allow_any_instance_of(CPU).to receive(:choice).and_return arg1
```
or the stronger

```rb
expect_any_instance_of(CPU).to receive(:choice).and_return arg1
```

views/game.erb
```diff
@@ -0,0 +1,27 @@
...
+
+<h1>
+  Welcome <%= session[:Name] %>
+</h1>
+
+<h2>
+  You are playing RPS against the computer
+</h2>
+
+<div class="left" style="background-color:#FCD57A; background-image: url(https://openclipart.org/image/125px/svg_to_png/213382/rock-paper-scissors.png);"><%= session[:Name] %></div>
```
tansaku: great use of background images, although we could replace div with a more semantic tag?

views/game.erb
```diff
@@ -0,0 +1,27 @@
...
+
+<div class="right" style="background-color:#FCD57A; background-image: url(https://openclipart.org/image/125px/svg_to_png/213382/rock-paper-scissors.png);">Computer</div>
+
+<p>
+  <form action="/game" method="post">
+    Make your choice <br>
+    <button type="submit" name="Choice" value="Rock">Rock!</button>
```
tansaku: trivial, but I'd prefer lower case for these names --> i.e. 'choice'

views/game.erb
```diff
@@ -0,0 +1,27 @@
+<link href='http://fonts.googleapis.com/css?family=Bangers' rel='stylesheet' type='text/css'>
+<link rel="stylesheet" href="<%= url('/styles.css') %>">
+
+<h1>
+  Welcome <%= session[:Name] %>
```
tansaku: I'd prefer lower case keys here - at least partly because in Ruby we use capital letters to denote classes ... not directly applicable here, but camelcase feels odd to me for hash keys

views/game_rpsls.erb
```diff
@@ -0,0 +1,29 @@
+<link href='http://fonts.googleapis.com/css?family=Bangers' rel='stylesheet' type='text/css'>
+<link rel="stylesheet" href="<%= url('/styles.css') %>">
+
+<h1>
+  Welcome <%= session[:Name] %>
```
tansaku: you should avoid accessing the session directly in your view in order to make the view as generic as possible

views/game_rpsls.erb
```diff
@@ -0,0 +1,29 @@
...
+</h1>
+
+<h2>
+  You are playing RPSLS against the computer
+</h2>
+
+<div class="left" style="background-color:#FCD57A; background-image: url(https://openclipart.org/image/125px/svg_to_png/213382/rock-paper-scissors.png);"><%= session[:Name] %></div>
```
tansaku: you should avoid accessing the session directly in your view in order to make the view as generic as possible

spec/decide_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    it 'knows P1 Lizard beats P2 Spock' do
+      expect(decide.make "Lizard", "Spock").to eq "win"
+    end
+
+    it 'knows P1 Lizard beats P2 Paper' do
+      expect(decide.make "Lizard", "Paper").to eq "win"
+    end
+
+    # etc etc etc
```
tansaku: hmmm ...

spec/cpu_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'cpu'
+
+describe CPU do
+  let(:cpu) { described_class.new }
+
+  it 'can choose Rock, Paper or Scissors' do
+    choice_array = []
+    100.times { choice_array << cpu.choice }
```
tansaku: I think this is a little dangerous - you're relying on a certainly level of randomness from 'sample'.

I think you'd be better off testing the contents of the array that is being drawn on, and that sample was being called on it.  Here you are effectively testing the sample method, which you should avoid ...

server.rb
```diff
@@ -0,0 +1,48 @@
...
+  get '/' do
+    erb :index
+  end
+
+  get '/game' do
+    redirect '/'
+  end
+
+  post '/game' do
+    if params[:Choice]
```
tansaku: this is the logic in your view that I was talking about.  You have indeed pulled your game logic into your models/objects, but this switch on the params[:Choice] content is logic that we should avoid.

You should prefer two different post methods here, e.g. 

```ruby
post '/game' do 
  @name = session[:name]
  erb :game
end

post '/result' do
  @decision = decide.make params[:choice], cpu.choice
  erb :result
````

to cleanly separate the concerns and reduce the need for logic in the server (controller).  I would also be tempted to pull the cpu choice logic into the model, and be able to have the view rely on a single result or decision object for all its data.

Also, I think we might want to think about changing the method names "decide.make" doesn't read very well ...

**Comment by tansaku** [comment]
Hi Sanj, 

This is great work.  Don't take my comments as indication of anything else.  You've done really really well here and mastered a number of very sophisticated aspects of programming such as stubbing random behaviour.

It's partly because you've done so well that I'm focusing in on the areas where there is still room for improvement.  

I'm going to give you a sticker and mark you completed because you've got a great product here - and the interface looks great, but I do hope you'll review my comments and think about making some of the changes I suggest, or at least bear them in mind going forward.

:checkered_flag:

File: views/index.erb [hunk 14]
```diff
@@ -0,0 +1,15 @@
...
+<h1>
+  Play Rock-Paper-Scissors!
+</h1>
+
+<p>
+  <form action="/game" method="post">
+    Enter your name:<br />
+    <input style="margin:15px; padding:5px; border-radius:6px;" type="text" name="Name" value=<%= session[:Name] %>><br />
+    <button type="submit" value="Play">Play RPS</button>
+    <button type="submit" value="PlayRPSLS" onclick="this.form.action='/game_rpsls'">Play RPSLS</button>
```
tansaku: this is kind of ugly - I think I would prefer to switch in the controller based on which button was pressed.  That might seem to go against the "no logic in the controller" motif that I'm presenting, but I think we could actually pull that logic through into the model objects as well and collapse the rather similar post '/game_rpsls' and post 'game' methods into one.

tansaku:
Also alot of similarities between game.erb and game_rpsls.erb - could be DRYed out, same for result and results_rpsls

PR #33, opened 2015-04-06
RPS Challenge

rjlynch:


features/play_game.feature
```diff
@@ -0,0 +1,58 @@
+Feature:
+  As a marketeer
+  So that I can see my name in lights
+  I would like to register my name before playing an online game
+
+  As a marketeer
```
tansaku: ideally you should have only a single feature per feature file

views/layout.erb
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <style>
+    .center {
```
tansaku: great to see you using css, but ideally you'll move this to a separate css file

views/result.erb
```diff
@@ -0,0 +1,26 @@
+<%display = {paper: '<img src="paper.png" alt="paper" height="42" width="42">',
```
tansaku: this is an ingenious solution - great work - however going forward you'll want to use partials for this :-)

http://www.sinatrarb.com/faq.html#partials

views/submit_p2_name.erb
```diff
@@ -0,0 +1,12 @@
+<form action="/submit_player2_name" method="post">
```
tansaku: this file looks rather similar to submit_p1_name.erb --> could DRY out using partials

lib/2_player_holder.rb
```diff
@@ -0,0 +1,3 @@
+class PlayerHolder
+  attr_accessor :player1_name, :player2_name, :player1_move, :player2_move
```
tansaku: avoid attr_accessor wherever possible - it prevents proper encapsulation of state - it's a holder with its internals hanging out

lib/rps_engine.rb
```diff
@@ -0,0 +1,27 @@
...
+    [:paper, :rock] => :paper,
+    [:scissors, :rock] => :rock,
+    [:paper, :scissors] => :scissors
+  }
+
+  def check_move move
+    MOVES.include? move
+  end
+
+  def compare player1_move, player2_move
```
tansaku: this is a rather long method that should be broken up

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,126 @@
...
+
+Then(/^I should not see "([^"]*)"$/) do |arg1|
+  in_browser :firefox do
+    expect(page).not_to have_content arg1
+  end
+end
+
+Given(/^A player has already started a game$/) do
+  in_browser :chrome do
+    step 'I am on the hompage'
```
tansaku: good use of sub-steps here, although these are sometimes frowned upon for obscuring what's happening under the hood - use with care

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,126 @@
+def in_browser name
```
tansaku: great work getting this sorted

server.rb
```diff
@@ -0,0 +1,81 @@
...
+require_relative 'lib/2_player_holder.rb'
+
+class RPSWeb < Sinatra::Base
+  enable :sessions
+
+  RPS = RpsGame.new
+  comp = ComputerPlayer.new
+  PLAYERS = PlayerHolder.new
+
+  def find_winner player1, player2, p1_move, p2_move
```
tansaku: this method is a bit long - consider passing a single object with the necessary data to the view, rather than lots of separate variables

server.rb
```diff
@@ -0,0 +1,81 @@
...
+    PLAYERS.player1_move = params[:move].to_sym
+    redirect '/two_players_ready'
+  end
+
+  get '/join_two_player_game' do
+    erb :submit_p2_name
+  end
+
+  post '/submit_player2_name' do
+    @player_two = params[:name]
```
tansaku: you've got some repetition here regarding player 1 and player 2 that could be DRYed out using the named parameters we've been looking at this week:

http://www.sinatrarb.com/intro.html#Routes

tansaku:
Hi Rich,

Great work getting this all working, particularly the two player version.  See my various embedded comments, and a few more general ones:

1) you have a couple of outstanding rubocop issues that could be easily fixed - overall your code is very neat - please do fix those last couple
2) You haven't really developed an OO solution.  You've got a single class with a long method, and two 'helper' classes that don't really do anything and don't have their own unit tests --> we should look at how to distribute the functionality across multiple objects
3) I worry that your tests are not covering all possible combinations of game play ...

But still great work.  Do you think you can find time to take another pass, particularly the OO issue?

:construction:

server.rb
```diff
@@ -0,0 +1,81 @@
+require 'sinatra/base'
+require_relative 'lib/rps_engine.rb'
+require_relative 'lib/computer_player.rb'
+require_relative 'lib/2_player_holder.rb'
+
+class RPSWeb < Sinatra::Base
+  enable :sessions
+
+  RPS = RpsGame.new
```
tansaku: so this a reasonable 'hacky' solution in the absence of a database, but it would be slightly cleaner to pull these constants into the classes themselves

PR #40, opened 2015-04-07
Basic challenge Plus a stab at the two players and sessions

Nadav-Rosenberg:


server.rb
```diff
@@ -0,0 +1,42 @@
...
+  
+  get '/' do
+    erb :homepage
+  end
+
+  get '/homepage' do
+    erb :play_again_page
+  end
+
+  post '/game_page' do
```
tansaku: this method has got a bit long, and I think you've got the wrong end of the stick about what we were suggesting in terms of a 2 player game

views/results_page.erb
```diff
@@ -0,0 +1,9 @@
+<center>
+
+<p>
+<%=session[:result]%>
```
tansaku: ideally we shouldn't access the session directly in views - instead prefer to set a variable and access that in the view

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def player_one_has_tool?
+    player_one[:tool] != nil
+  end  
+
+  def player_two_has_tool?
+    player_two[:tool] != nil
+  end  
+
+  def declare_winner
```
tansaku: this method is very long, and ideally you need the functionality of this class spread over multiple classes

tansaku:
Hi Nadav,

Good work getting the basic engine working.

My main concerns here are a lack of OO design and that you got the wrong end of the stick regarding the two player game.

I think it might have made more sense to get a simple single player game working - submit that as safety and then work on the two player game in a branch?

:construction:

PR #47, opened 2015-04-07
RPS wk3 Challenge

PaulWallis42:


tansaku:
Great work getting the game working Paul - I'm able to play it in the browser, which is important.  You've also got your game logic in your models (Game object and Player object)

Here are some general comments on the code:

* README should be updated
* pay attention to general hound/rubocop issues - code is really messy in places
* use html label tags http://www.w3schools.com/tags/tag_label.asp
* avoid using br tags for spacing - prefer css margin adjustment - although see http://stackoverflow.com/questions/462619/should-br-and-hr-be-avoided-at-all-costs-in-web-design and http://stackoverflow.com/questions/1726073/is-it-sometimes-bad-to-use-br
* vacuous test in player_spec - no point in testing your doubles
* game spec doesn't actually find out who the winner is
* feature tests similarly don't actually test the game

The biggest issue here is that the tests are not really helping your development at the moment.  We should sit down together and look at how to use the tests to help you create your code.

:construction:

PR #52, opened 2015-04-07
Working game, but still need to do work on integrating with the website

phoebehugh:


views/layout.erb
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+<link href="/css/styles.css" rel="stylesheet" type="text/css" />
+  <body>
+    <center>
+    <img src="/images/rps.png" alt="Battleships" align="middle">
```
tansaku: alt tag out of date :-)

phoebehugh: Ahhh copied my battleships form!!

spec/computer_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'computer'
+
+describe Computer do
+
+  let(:computer) { Computer.new }
+
+  it 'can pick a random move' do
+    expect(computer).to respond_to :random_move
```
tansaku: you don't really need this expect statement when you have the 'computer.random_move' called on the next line.

spec/computer_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+    computer.random_move
+    expect(computer).not_to be_nil
```
tansaku: to test the random behaviour here we should probably prefer to check that sample is called on the correct array - perhaps we can look at this together ...

lib/player.rb
```diff
@@ -0,0 +1,16 @@
+require_relative 'game'
+
+class Player
+
+  attr_accessor :name
```
tansaku: avoid attr_accessor wherever possible

lib/game.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def winner
+    case
+      when @players[0].move == :scissors && players[1].move == :paper
+        return @players[0]
+      when @players[0].move == :rock && players[1].move == :scissors
+        return @players[0]
+      when @players[0].move == :paper && players[1].move == :rock
+        return @players[0]
+      when @players[1].move == :scissors && players[0].move == :paper
```
tansaku: This section of code is not being tested - see https://coveralls.io/builds/2281883/source?filename=lib%2Fgame.rb

lib/game.rb
```diff
@@ -0,0 +1,40 @@
+class Game
+
+  attr_reader :players
+
+  def initialize(players = [])
```
tansaku: you should probably avoid making players an array until you know you are going to need more than two ...

lib/game.rb
```diff
@@ -0,0 +1,40 @@
...
+    has_a_player?
+  end
+
+  def has_a_player?
+    !players.nil?
+  end
+
+  def winner
+    case
+      when @players[0].move == :scissors && players[1].move == :paper
```
tansaku: It's great that you have the overall game logic working here, however there's a lot of room to DRY things out.  One step would be to call the following at the start:

```ruby
move0 = @players[0].move
move1 = @players[1].move
```

and then refer to those subsequently.  

Even better would be to pull the pattern of wins and losses our of the boolean statments and into a hash that one could use to look up the result ...

tansaku:
Hi Phoebe,

Great work getting the basic game working.  I can see that you got stuck trying to get it to work through the web interface, but you are going in the right direction.

I can see that you had an issue trying to make the stylesheets and images accessible - which relates to the location of your public directory, and that ultimately your web game doesn't quite work, and the cucumber tests are not complete.

I'd strongly encourage you to work through the various hound/rubocop issues that your code has in order to clean it up.

You've got a good base here, but we should probably chat about how to move forward ... 

:construction:

PR #57, opened 2015-05-15
Web RPS one player version

DanielJohnston:


features/start_game.feature
```diff
@@ -0,0 +1,16 @@
...
+
+  Scenario: Navigating to registration page
+    Given I am on the homepage
+    Then I should see "What's your name?"
+
+  Scenario: Entering name and going to game
+    Given I am on the homepage
+    When I fill in "name" with "Daniel"
+    When I press "Let's go!"
+    Then I should be on the Play RPS page
```
ptolemybarnes: Having three expectations about the Play RPS page creates quite a tight coupling between your tests and production code. What if you come to add Spock etc? The expectation from line 16 will fail. It's better to have a single expectation that tests something about the page that is unlikely to change (and relates to the scenario of course). So for this example, "Entering name and going to game", testing that you're now on the Play RPS game would probably be sufficient.

lib/game.rb
```diff
@@ -0,0 +1,55 @@
+class Game
+attr_accessor :player_1
+attr_accessor :player_2
+attr_reader :players
+
+  def initialize player_class, winning_score = 5, multiplayer = false
+    @player_1 = player_class.new
+    @player_2 = player_class.new
+    @winning_score = winning_score
+    if multiplayer == true
```
ptolemybarnes: You needn't have an equality checker here. Just "if multiplayer" would suffice.

lib/game.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+    @plays = [:Rock, :Paper, :Scissors]
+    @game_over = false
+  end
+
+  def turn_over?
+    @player_1.has_played? and @player_2.has_played?
+  end
+
+  # Yup yup, this could totally use refactoring, I agree
```
ptolemybarnes: Agreed :smile_cat:

lib/player.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def play what_to_play
+    @played = what_to_play
+    @has_played = true
+  end
+
+  # This method is stubbed in feature tests to control randomness
+  def autoplay
+    @played = [:Rock, :Paper, :Scissors].select
```
ptolemybarnes: #select without a block returns an enumerator. Should this be #sample?

lib/player.rb
```diff
@@ -0,0 +1,61 @@
+class Player
```
ptolemybarnes: Generally this class is great :+1: , nice clean methods. Think about how you could refactor it to access instance variables via a public interface, however.

lib/rps_web.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  post '/play' do
+    # Create game if it doesn't exist yet
+    # Unimplemented: Spock and Lizard
+    multiplayer = true if params.has_key?('players')
+    @@game ||= Game.new Player, params[:score].to_i, multiplayer
+
+    # Check if a player name's been entered, store if so
+    if params.has_key?('name')
```
ptolemybarnes: Both of these checks on params could be refactored to helper methods.

lib/game.rb
```diff
@@ -0,0 +1,55 @@
...
+  def end_turn
+    # Figure out who won, add a point accordingly
+    if @player_1.played == @player_2.played
+      outcome = :draw
+    elsif @plays.index(@player_1.played) == ((@plays.index(@player_2.played)) -1) % 3
```
ptolemybarnes: Tip on refactoring: the logic here is rather confusing. Try to come up with descriptively named methods and extract logic out to those.

ptolemybarnes:
great attempt. There is a lot of refactoring that could be done. Let me know if you want to 1 to 1 and we can start on it together.
:checkered_flag:

PR #58, opened 2015-05-16
RPS-Challenge

chidumaga:


tansaku:
Hey Chidu,  just played your RPS - good work getting a working game - functionality looks good.  

Looking at the code I see that your feature test is not really a feature test.  A feature test for an RPS web game should be using capybara.  The feature test you have is actually a unit test, so needs to be moved into the root of the spec folder.

You also have a lot of replication through this test - e.g.     cpu = CPU.new should be pulled out into a `let(:cpu){CPU.new}`, although since this test is really a unit test of the GameLogic class, that cpu should be a double.

Line 7 of your GameLogic Class should be formatted like this:

```
    @game_rules = { "Rock" => ["Scissors", "Lizard"], 
                    "Paper" => ["Rock", "Spock"], 
                    "Scissors" => ["Paper", "Lizard"], 
                    "Lizard" => ["Paper", "Spock"], 
                    "Spock" => ["Scissors", "Rock"] }
```

and you should remove the commented code from rps_server.rb.

So there's gome good work here, but before I give this the all clear I'd like to see you turn that test into the proper unit test (with appropriate doubles) and take a stab at a real feature test using capybara (you can use rspec if you don't like cucumber). Here's an example of doing capybara feature tests with rspec if you prefer that:

https://github.com/makersacademy/course/blob/master/battle_ships_web/battle_ships_web_rspec.md

Ping me when you make updates and I'll take another look

:construction:

chidumaga:
Thanks for the comments @tansaku, I have made changes to my tests, my feature test is written in Cucumber and is in the features folder

PR #59, opened 2015-05-16
rps-challenge

andygout:


lib/player.rb
```diff
@@ -0,0 +1,23 @@
+class Player
+
+  attr_accessor :name
```
silvabox: Should this be an `attr_reader`?

lib/player.rb
```diff
@@ -0,0 +1,23 @@
...
+  attr_reader :wins
+
+  def initialize name
+    @name = name
+    @wins = 0
+  end
+
+  def cpu_choice rules
```
silvabox: CPU rules do not belong in the player class.

lib/rps.rb
```diff
@@ -0,0 +1,69 @@
...
+  get '/game/new' do
+    @@game = Game.new Player
+    erb :register
+  end
+
+  post '/game/get_ready' do
+    @@game.player_1.name = params[:name] if params[:name] =~ /\S+/
+    @@game.goal = params[:goal].to_i if params[:goal] =~ /\d/
+    @@game.rules = params[:rules]
+    @rules = @@game.rules
```
silvabox: try to avoid more than 2 instance variables in your route.  If you set `@game = @@game` instead, your view will have access to `@game.rules` and `@game.goal`

lib/game.rb
```diff
@@ -0,0 +1,66 @@
+class Game
+
+  GOAL = 1
+
+  attr_accessor :rules
```
silvabox: Should these all be `attr_reader`

lib/game.rb
```diff
@@ -0,0 +1,66 @@
+class Game
+
+  GOAL = 1
```
silvabox: I would call this `DEFAULT_GOAL`

lib/game.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  attr_accessor :rules
+  attr_accessor :player_1
+  attr_accessor :player_2
+  attr_accessor :goal
+
+  def initialize playerClass, goal = GOAL
+    @player_1 = initialize_player playerClass, "Player 1"
```
silvabox: By giving `Game` the responsibility of creating the players, you have painted yourself into a corner.  You cannot have 2 different types of player.  Ideally, you should inject the `player_1` and `player_2` so that they can be different types.  Then you player does not need to have the cpu behaviour too - it can be a different class.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+    @rules = 'rps'
+    @goal = goal
+    @rounds_played = 0
+  end
+
+  def initialize_player playerClass, name
+    player = playerClass.new name
+  end
+
+  def result p1_choice, p2_choice
```
silvabox: This long `if` statement is very hard to read.  Consider other ways to encapsulate the rules.

lib/game.rb
```diff
@@ -0,0 +1,66 @@
...
+      self.player_1
+    else
+      self.player_2.add_win
+      self.player_2
+    end
+  end
+
+  def reason p1_choice, p2_choice
+    key = []
+    combs = { 'lizard-paper': 'Lizard eats paper',
```
silvabox: This is quite a nice idea, but you should extract this hash into a constant.

lib/player.rb
```diff
@@ -0,0 +1,23 @@
...
+    if rules == 'rps'
+      ['rock', 'paper', 'scissors'].sample
```
silvabox: extract these arrays into constants

spec/player_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'player'
+
+describe Player do
+
+  context 'Play a game' do
+
+    let(:game) { Game.new Player }
```
silvabox: You should not be using `Game` in your `Player` unit tests.

silvabox:
Great work Andy - you have a fully playable game and all tests are passing.  Have a look through my comments; hopefully they'll be useful for the next challenge!

:checkered_flag:

PR #61, opened 2015-05-16
Sinatra + bootstrap #april2015

rodcul:
https://rps-rodcul.herokuapp.com


features/enter_name.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Enter my name
+
+  As a marketeer
+  So that I can see my name in lights
+  I would like to register my name before playing an online game
+
+  Scenario: Enter my name and see it on game page
+
+  Given I am on the homepage
```
silvabox: Be careful with your indentation - these lines should be indented under `Scenario:`

features/play_game.feature
```diff
@@ -0,0 +1,22 @@
+Feature: Play a game
```
silvabox: You have no feature test for lose.

features/play_game.feature
```diff
@@ -0,0 +1,22 @@
...
+
+  As a marketeer
+  So that I can enjoy myself away from the daily grind
+  I would like to be able to play rock/paper/scissors
+
+  Scenario: TIE rps mode: Player 1 & CPU both choose paper
+    Given I am on the homepage
+    And I seed with 2
```
silvabox: Cucumber steps should be high-level and non-technical.  `And I seed with 2` is meaningless within the domain.  Instead, define your step as `And the computer will choose Rock` or something similar.

lib/rps.rb
```diff
@@ -0,0 +1,32 @@
+class RPS
+  attr_reader :rules
+
+  def initialize
+    @rules = {
```
silvabox: Consider using a constant instead of an instance variable.  But otherwise this is an excellent strategy for defining the rules

lib/rps.rb
```diff
@@ -0,0 +1,32 @@
...
+    elsif rules[p1].include? p2
+      result = :win
+    else
+      result = :lose
+    end
+    { cpu_choice: p2, result: result }
+  end
+
+  def random(mode)
+    n = 3 if mode == :rps
```
silvabox: This is a good approach computationally, but it's slightly too cryptic.  Think about splitting this into 2 methods - one that handles the random selection for rps and the other for rpsls.  There are a number of alternate OO design patterns for this scenario, for example the Strategy pattern - but I'll leave that to you to look up if you want to.

lib/rps.rb
```diff
@@ -0,0 +1,32 @@
...
+    @rules = {
+      rock: [:scissors, :lizard],
+      paper: [:spock, :rock],
+      scissors: [:paper, :lizard],
+      lizard: [:spock, :paper],
+      spock: [:scissors, :rock]
+    }
+  end
+
+  def results(p1, mode = :rps)
```
silvabox: You could initialize the RPS instance with `mode` to avoid the need to pass it in here.

Avoid parameter names like `p1` and prefer descriptive names like `player`.

lib/rps.rb
```diff
@@ -0,0 +1,32 @@
...
+    p2 = random(mode)
+    if p1 == p2
+      result = :tie
+    elsif rules[p1].include? p2
+      result = :win
+    else
+      result = :lose
+    end
+    { cpu_choice: p2, result: result }
```
silvabox: A method should be responsible for only one thing.  The need to resort to a hash to return 2 objects from this method is a design smell.  It clearly indicates that you need to refactor.  `results` is currently responsible for getting and returning the cpu choice AND comparing this with the player's choice, which violates SRP.

rps_web.rb
```diff
@@ -0,0 +1,39 @@
...
+    @name = session[:name]
+    @mode = session[:mode]
+    erb :game
+  end
+
+  get '/game' do
+    @name = session[:name]
+    @mode = session[:mode].to_sym
+    unless params[:choice].nil?
+      game = RPS.new
```
silvabox: This action relies on 6 instance variables:

* `@name`
* `@mode`
* `@p1_choice`
* `@results`
* `@cpu_choice`
* `@result`

This is generally frowned upon and suggests you need some sort of class structure

spec/rps_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'rps'
+
+describe RPS do
+  context 'Classic rock, paper, scissors' do
+    before :each do
+      srand 2 # makes CPU always return :rock
+    end
+
+    it 'TIE: Player & CPU choose rock' do
```
silvabox: Try to make your descriptions read well including the `it`.  "it TIE: Player & CPU choose rock" does not make sense.  The `it` structure helps to structure your tests in terms of the thing being described

spec/rps_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'rps'
+
+describe RPS do
```
silvabox: Your unit tests do not test all possible routes through the application.  You should try to test every possible combination  and exceptional behaviour too.

silvabox:
Great work Rodney.  You have a fully playable and nice-looking game.  There are no major issues in your code, but I've made some comments for you to think about longer-term.
:checkered_flag:

PR #62, opened 2015-05-17
RPS Challenge Submission [Alex H]: One Player Version with Lizard and Spock

AlexHandy1:


features/take_turns.feature
```diff
@@ -0,0 +1,19 @@
...
+    Given I am on the Take Turn page
+    And I choose radio button "Rock"
+    When I press "Submit"
+    Then I should see "Rock"
+
+  Scenario: Player wins a game
+    Given I am on the Take Turn page
+    And I choose radio button "Rock"
+    Given my opponent chooses "Scissors" next
+    And I choose radio button "Rock"
```
silvabox: Did you mean to duplicate these two lines?

features/take_turns.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Take A Turn
```
silvabox: You are missing feature tests for losing and drawing

lib/game.rb
```diff
@@ -0,0 +1,75 @@
+require_relative 'player'
+
+class Game
+  attr_reader :player_1, :player_2
+
+  def initialize playerClass
```
silvabox: There is a subtle design issue here.  By making the game responsible for creating player 1 and player 2, you cannot use different classes for each player.  This has led to a number of design smells in the rest of your code.  It would have been better to pass player 1 and player 2 into game as parameters, rather than the class.

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    @player_1 = playerClass.new
+    @player_2 = playerClass.new
+  end
+
+  def process_turn
```
silvabox: method names like `process_turn` are a red flag.  They often indicate a problem with the design and indicate procedural code patterns rather than OO.

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    check_round_wins
```
silvabox: As above, `check_round_wins` is a red flag

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    return "Game already won" if winner?
```
silvabox: Avoid display logic in your business logic classes.  Return codes to (e.g. symbols) instead and then translate these in your view/controller layer

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    if match?
+      "Go again"
+    else
+      if rock_wins? || scissor_wins? || paper_wins? || lizard_wins? || spock_wins?
```
silvabox: This should be refactored into another method.  It's also not clear that `rock_wins?` etc. really mean s "player 1's rock wins"

lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    player_1.wins
+    player_2.wins
+  end
+
+  def match?
+    player_1.check_turn == player_2.check_turn
+  end
+
+  #RESULT EVALUATIONS
+  def rock_wins?
```
silvabox: There is an awful lot of structural duplication here.  Is there a better way of capturing the rules of the game?

lib/player.rb
```diff
@@ -0,0 +1,36 @@
+class Player
+  def initialize
+    @turn = []
```
silvabox: Prefer plural names for arrays and hashes i.e. `@turns`

Is this array necessary?  Do you ever utilize it as an array?

lib/player.rb
```diff
@@ -0,0 +1,36 @@
...
+    @winner = false
+    @round_wins = 0
+  end
+
+  def take_a_turn turn
```
silvabox: `take_a_turn` is a bit clumsy when you use it.  Think of it in terms of "(Player) take a turn Rock" - it's not quite right.  `choose` or `play` might be better: "(Player) play Rock".

lib/player.rb
```diff
@@ -0,0 +1,36 @@
...
+    @turn << turn
+  end
+
+  def random_selection
```
silvabox: `random_selection` is too abstract.  Random selection of what?  Is this an attribute or an interaction?

Because of your game is restricted to types of `Player` you have player and computer behaviour mixed together in this class, which has made it problematic.

lib/player.rb
```diff
@@ -0,0 +1,36 @@
...
+    @turn << ["Rock", "Paper", "Scissors", "Lizard", "Spock"].sample
```
silvabox: Use a constant instead of creating a new array every time.

lib/player.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def winner?
+    @winner
+  end
+
+  def wins
```
silvabox: `wins` is a tricky method name.  Again, is it a noun or a verb?  The method is actually unnecessary - you are using 2 variables to track whether the user is a winner: `@winner` and `@round_wins`.  You don't need `@winner` - in `winner?` just return `@round_wins >= 2`

lib/player.rb
```diff
@@ -0,0 +1,36 @@
...
+    @winner = true if @round_wins >=2
+  end
+
+  def add_round_win
+    @round_wins += 1
+  end
+
+  #ACCESSORS
+  def check_turn
```
silvabox: Use nouns for accessors and verbs for interactions.  `check_turn` implies that your are performing some sort of validation.  It should be `turn`

lib/player.rb
```diff
@@ -0,0 +1,36 @@
...
+    @turn.last
+  end
+
+  def check_round_wins
```
silvabox: As above, should be `round_wins`

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,77 @@
...
+      erb :game
+    end
+  end
+
+  get '/game/turn' do
+    erb :game_turn
+  end
+
+  post '/game/turn' do
+    @player_turn = params[:choice]
```
silvabox: There are too many lines in here.  You have presentation logic in your route, which belongs in the view.  The route (i.e. controller) is responsible for interpreting the request, interacting with the business logic layer (but not performing any business logic) then deciding the response (and accordingly handing off to the presentation layer).

spec/game_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'game'
```
silvabox: I can't find any tests for the actual game logic!  How do you know that the rules of the game are satisfied for every possible play?

silvabox:
Well done Alex - you have a great looking and playable game.  There is one failing scenario though.

There are quite a few problems with your design and I've made a number of comments which will hopefully make sense and help you to think of a way to structure it differently.

:checkered_flag:

PR #63, opened 2015-05-17
rps working against computer. test coverage ad maiora

bagolol:


features/rps_cucumber_test.feature
```diff
@@ -0,0 +1,18 @@
+Feature: Starting the game
+  In order to play rps
+  As a bored office worker
+  I want to start a "New Game"
+
+  Scenario: Signing in
+    Given I am on the homepage "/"
+    When I click "New Game"
+    Then I should see "What's your name?"
+    When I enter "Rocco"
```
silvabox: Whens should not follow Thens.  This is a sign that you need 2 tests

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,55 @@
...
+  fill_in("value", :with => name)
+end
+
+When(/^I hit "([^"]*)"$/) do |submit|
+  click_button(submit)
+end
+
+Then(/^I see "([^"]*)" and "([^"]*)"$/) do |text, start|
+  expect(page).to have_content text
+  expect(find_link(start).visible?).to be true
```
silvabox: this expectation should not presume start to be a link

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,55 @@
...
+end
+
+Given(/^I have submitted "([^"]*)" as my name$/) do|name|
+  visit("game/new")
+  fill_in("value", :with => name)
```
silvabox: 'value' is too general a target, your field should be called `name`.

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,55 @@
...
+  click_button("Submit")
+end
+
+When(/^I select "([^"]*)"$/) do |move|
+  select(move, :from => "move")
```
silvabox: Again, this step is too specific for the matcher, which could be selecting from anything

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,55 @@
...
+end
+
+Then(/^I should see a result$/) do
+
+end
+
+Given(/^I choose "([^"]*)"$/) do |move|
+  visit("game/new")
```
silvabox: This step is too specific for the matcher, which could be choosing from anything

lib/computer.rb
```diff
@@ -0,0 +1,12 @@
+class Computer
+
+  attr_accessor :moves
```
silvabox: You have defined an `attr_accessor` and a `play` method.  Use one or the other as the move can be overwritten using the setter method.

lib/computer.rb
```diff
@@ -0,0 +1,12 @@
...
+  def initialize
+    @moves = nil
+  end
+
+  def play
+    @moves = ["rock", "paper", "scissors"].sample
```
silvabox: Use a constant to store the options instead of repeatedly creating this array.

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+class Game
+
+  attr_reader :player1, :player2
+
+  def initialize player1, player2
+    @player1 = player1
+    @player2 = player2
+  end
+
+  def winner?
```
silvabox: methods with a question mark should only be considered as true or false.

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+    if a == b
+      "draw"
+    elsif a == "rock" && b == "scissors"
+      "Player1 wins"
+    elsif a == "paper" && b == "scissors"
+     "Player2 wins"
+    elsif a == "scissors" && b == "paper"
+     "Player1 wins"
+    elsif a == "paper" && b == "rock"
+      "Player1 wins"
```
silvabox: Avoid using display strings as return values from components.  Return coded responses instead - e.g. symbols:  `:draw`, `:win`, `:lose` and translate these in the ui layer.

lib/player1.rb
```diff
@@ -0,0 +1,12 @@
+class Player
+
+  attr_accessor :moves
```
silvabox: See earlier comment on `Computer` regarding `attr_accessor`

lib/rps.rb
```diff
@@ -0,0 +1,52 @@
...
+      erb :game_new
+    end
+  end
+
+  get '/game' do
+    erb :play
+  end
+
+  post '/result' do
+    session["move"] = session["game"].player1.play params[:move]
```
silvabox: Why are you using `session["move"]` here?  you immediately overwrite it on the next line then don't use it again.

lib/rps.rb
```diff
@@ -0,0 +1,52 @@
...
+    session["move"] = session["game"].player2.play
```
silvabox: Avoid repeated code.  You have `session["game"] three times in this route.  Use a local variable  - i.e. `game = session[:game]`

views/result.erb
```diff
@@ -0,0 +1,20 @@
+<div align="center">
+  <% if @result == "draw" %>
+    <p style="color:red">DRAW!</p>
+    </br>
+    <a href='/game'>Replay?</a>
+  <% elsif @result == "Player1 wins" %>
+    <p><span style="color:red"><%= session["user"] %>, you have won!</span></p>
```
silvabox: avoid access to `session["user"]` and similar from within your views.  Sessions belong in the controller - the view should not know about them.  Use controller instance variables or helper methods to wrap access to the session.

silvabox:
Hi Rocco, thanks for submitting your RPS challenge.  I ought not to pass it as you have failing tests, but you do have a playable game, so I'll let you off :wink: 

Have a look at my comments - I hope they all make sense.

:checkered_flag:

PR #64, opened 2015-05-17
Rock Paper Scissors Challenge

MollieS:


features/play_game.feature
```diff
@@ -0,0 +1,35 @@
+Feature: Playing a game of Rock Paper Scissors
```
ptolemybarnes: Indentation!??

features/step_definitions/playing_a_game_steps.rb
```diff
@@ -0,0 +1,8 @@
+When(/^I see "([^"]*)"$/) do |text|
+  page.has_content?(text)
```
ptolemybarnes: Should this be an expectation? Like so:
    expect(page).to have_content(text)

lib/rps_web.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  post '/game' do
+    @choice = params[:choice]
+    @computerchoice = ["Rock", "Paper", "Scissors"].sample
+    self.winner
+    erb :game
+  end
+
+  def winner
```
ptolemybarnes: You've implemented the logic inline within your Sinatra class, which is not a bad idea and, thinking about what our MVP is, might be the first step. You could then refactor this out to a separate class for reuse elsewhere. :+1:

features/play_game.feature
```diff
@@ -0,0 +1,35 @@
...
+In order to play a game of rock paper scissors
+I would like to start a game
+And play against my computer
+
+Scenario: Registering
+Given I am on the homepage
+When I follow "New Game"
+And I see "What's your name?"
+When I type in "name"
```
ptolemybarnes: Cucumber is about specifying what you want with examples. Using the string "name" will confused non-technical people, clients etc, as they'll be wondering in what situation you would want that functionality.

views/newgame.erb
```diff
@@ -0,0 +1,42 @@
+<head>
+<title>NEWGAME!</title>
+</head>
+<style>
```
ptolemybarnes: Nice job on styling but could you move this out to a separate .css file?

ptolemybarnes:
:checkered_flag:

PR #66, opened 2015-05-17
Sunday evening cutoff

dan-bolger:
After a solid working irb-ready ruby app, hit some bumps in the road with trying to get on to the web after earlier blockages setting up cucumber/sinatra pushed things back


features/play_game.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Playing the game
+  So that I can enjoy myself away from the daily grind
+  As a marketeer
+  I would like to be able to play rock/paper/scissors
+
+  Scenario: Playing a move
```
silvabox: This should be 2 separate scenarios as you are testing 2 separate steps.  If you find yourself writing 'when' statements after 'then' statements (`And I press "Rock"` is really `When I press "Rock"), then that's a clue.

features/register_name.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Registering your name
+  So that I can see my name in lights
+  As a marketeer
+  I would like to register my name before playing an online game
+
+Scenario: Putting the name in the form
```
silvabox: See previous comment on Thens and Whens.

features/step_definitions/step_definitions.rb
```diff
@@ -0,0 +1,27 @@
+Given(/^I am on the homepage$/) do
+  visit '/'
+end
+
+Then(/^I should see "([^"]*)"$/) do |arg1|
+  page.has_content?("What's your name?")
```
silvabox: The regex `^I should see "([^"]*)"$` will match any step starting `I should see"`followed by any characters other than `"` and ending with `"`.  The sequence of characters between `"` and `"` will be passed as the value if `arg1`.  So

`Then I should see "What's your name?"` will match this step, and `arg1` be `What's your name?`.  So you should not use the string literal in the step, but `arg1` instead; i.e.: `page.has_content? arg1`.

Note also that this is a 'Then' step, so it should be an expectation - i.e.  `expect(page).to have_content arg1`

features/step_definitions/step_definitions.rb
```diff
@@ -0,0 +1,27 @@
...
+end
+
+When(/^I fill in "([^"]*)" with "([^"]*)"$/) do |arg1, arg2|
+  fill_in('name', :with => 'Dan')
+end
+
+When(/^I press "([^"]*)"$/) do |play|
+  click_button('play')
```
silvabox: See note above about regex.  What is the value of the argument `play`?  What should you be passing to `click_button`?

lib/paper.rb
```diff
@@ -0,0 +1,11 @@
+class Paper
```
silvabox: Having separate classes for `Rock`, `Paper` and `Scissors` is a nice design decision.  Ordinarily, we would discourage testing for classes as it prevents duck typing.  Where you are using the class as a 'value', that's less of an argument.

However, it does make it difficult to unit test as you can't easily use doubles

lib/paper.rb
```diff
@@ -0,0 +1,11 @@
...
+  def versus move
+    if move.class == Paper
```
silvabox: What you could do is have a `rock?`, `paper?` and `scissors?` method in each of your classes, which simply return `true` and `false` accordingly.  Then this logic could be:
```
if move.paper?
  :draw
elsif move.scissors?
  :loss
else
  :win
end
```
This would enable you to test with doubles.  i.e.:
```
scissors = double :scissors, scissors?: true, paper?: false, rock?:false
```

spec/paper_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'paper'
+require 'stone'
+require 'scissors'
+
+describe Paper do
```
silvabox: Because you are using class identity to test who wins, it's very difficult to use doubles.  This means you have 'Chicago style' unit tests, which are identical to your feature tests, so one or other is redundant.

With the `paper?` etc. technique I mentioned above, you can unit test Paper independently of `Scissors` and `Rock`.  By making your code testable, you'd have achieved an even better design.

silvabox:
Hi Dan, thanks for submitting you RPS challenge.  I think you stumbled over some conceptual issues with Cucumber and probably felt there was something not quite right with all those duplicated tests.  Hopefully my comments will make sense.

However, separating the design into the three classes has considerable merit and in more complex systems, those decisions have great payoffs.

:construction:

PR #67, opened 2015-05-18
RPS Challenge

ashleigh090990:


lib/computer.rb
```diff
@@ -0,0 +1,12 @@
+class Computer
+  attr_reader :choice
+
+  def initialize
+  	@choice = 'Computer says no!'
+  end
+
+  def computer_choice
```
silvabox: Is this a method or an attribute?  As you have an `attr_reader :choice`, then I'm guessing this is a method.  Try to use verbs in your method names and avoid extraneous context (i.e. 'computer', when the class is already 'Computer'.  `choose` would be a much better name for this method.

lib/computer.rb
```diff
@@ -0,0 +1,12 @@
...
+  	@choice = %w(rock paper scissors).sample
```
silvabox: Good use of the `%w` shorthand, although you could define this array as a constant so that doesn't need to be created every time this method is called.

lib/game.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'player'
```
silvabox: You don't reference player or computer in this file - are these requires just for convenience?

lib/player.rb
```diff
@@ -0,0 +1,20 @@
+class Player
+  
+  attr_reader :choice
+
+  def initialize
+  	@choice = 'No choice has been made yet!'
+  end
+
+  def player_choice(choice)
+  	if choice == 'rock'
```
silvabox: This is a long and unnecessary `if` statement that can be handled in 2 lines:
```
fail "You can't use that in a rock, paper, scissors game!" unless %w(rock paper scissors).include? choice
@choice = choice
```
Although I would certainly refactor the `%w(rock paper scissors)` array into a constant somewhere.

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+require_relative 'computer'
+
+class Game
+  
+  
+  
+  def compare(player_choice, computer_choice)
```
silvabox: This method is problematic and ripe for a refactor.  The main issues are with all the repetitious logic and messages.

Ideally, you should try to keep program logic and user interface logic separate (separation of concerns) and give a method just one responsibility (single responsibility principle).  In this case, that would mean having one method deciding if it's a win, loss or draw, and another method translating that into a message.

Long decision trees like this are hard to read and you should look for other ways to represent them.  Could you use a hash for example?

Finally, since you are passing in `player_choice` and `computer_choice`, `Player` and `Computer` are largely redundant.  Instead, you could pass in `player` and `computer` parameters as the instances of `Player` and `Computer`, and test for `player.choice` and `computer.choice`.

Interestingly, you could call these parameters `player_1` and `player_2`, and pass in 1 `Player` + 1 `Computer`, or 2 `Player`s, or 2 `Computer`s - and support single and 2-player games with one method.  As long as both objects have a `choice` method (remember duck typing?)

silvabox:
Hi Ashleigh, thanks for submitting your RPS challenge!  I can't mark it as complete as it's not playable through the browser, but it's good to have a working irb version to start with.  Have a look through my comments.
:construction:

PR #69, opened 2015-05-18
RPS Challenge - Not yet finished

curlygirly:
Bug somewhere needs fixing and then some formatting to make it look good.


lib/game_engine.rb
```diff
@@ -0,0 +1,30 @@
+class GameEngine
+  attr_accessor :player_choice
```
silvabox: This `attr_accessor` is never used.  It is also not tested anywhere.  What is it for?

lib/game_engine.rb
```diff
@@ -0,0 +1,30 @@
...
+
+
+  def result(player_choice)
+
+    if
+      player_choice == :rock && cpu_choice == :scissors ||
```
silvabox: This logic is both unreadable and incorrect.  Firstly, you shouldn't mix `AND`s and `OR`s in this way because the operator precedence is not clear.  You should avoid having more than one logical assertion in one line of code where possible.  Readability is more important the brevity.
Secondly, `cpu_choice` is a method, so all 3 uses of it in this line of code will create a potentially different value.  So this logic will never work.

lib/game_engine.rb
```diff
@@ -0,0 +1,30 @@
...
+      puts :WIN
+    elsif
+      player_choice == cpu_choice
+      :TIE
+    else
+      :LOSE
+    end
+  end
+
+  def cpu_choice(rand_no=3)
```
silvabox: What's `rand_no` for?  It is never used.

lib/game_engine.rb
```diff
@@ -0,0 +1,30 @@
...
+    options = [:rock, :paper, :scissors]
```
silvabox: You should use a constant rather than creating a new array on every single method call.

lib/game_engine.rb
```diff
@@ -0,0 +1,30 @@
...
+    options[rand(0..2)]
+  end
+
+
+  def random
```
silvabox: This method is untested and unused.

lib/rps_web.rb
```diff
@@ -0,0 +1,44 @@
+require 'sinatra/base'
+require_relative 'game_engine'
+
+class RPSWeb < Sinatra::Base
+  set :public_dir, Proc.new{File.join( root, '..', "public") }
+  enable :sessions
+
+  @@game = GameEngine.new
```
silvabox: You are not persisting any state in GameEngine - you have used the session instead.  So having a class variable `@@game` here is unnecessary.

lib/rps_web.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  get '/' do
+
+    erb :index
+  end
+
+  post '/game1' do
```
silvabox: Why is this called `game1`.  If you think about REST, you are posting the name, so maybe `post '/name' would be more desriptive?

lib/rps_web.rb
```diff
@@ -0,0 +1,44 @@
...
+    session[:name] = params[:name]
+      @name = params[:name]
+    erb :game1
+  end
+
+
+  post '/game_choice' do
```
silvabox: `post '/game'` would be preferable I think.

lib/rps_web.rb
```diff
@@ -0,0 +1,44 @@
...
+    erb :game_choice
+  end
+
+
+  post '/results' do
+      puts 'is this thing really working?'
+      puts params
+      @name = session[:name]
+      @player_choice = session[:player_choice]
+      @mac_choice = :cpu
```
silvabox: Whenever I run it, I get `Mac chose cpu` as a response.  I want to know whether it picked rock, paper or scissors!

lib/views/results.erb
```diff
@@ -0,0 +1,25 @@
+<div>
+<form action="/" method = "get">
+ <div class="container">
+ <div class="jumbotron">
+      <h1><%= @name %>!</h1>
+        <h2>You chose <%=@player_choice%><h2><br/>
+        <h2>Mac chose <%=@mac_choice%><h2><br/>
+<br>
```
silvabox: don't use HTML for layout.  `<br>` tags are a red flag.  Use CSS instead.

lib/views/results.erb
```diff
@@ -0,0 +1,25 @@
...
+        <h1><center>YOU LOSE!!</h1>
+      <% elsif @outcome == :TIE %>
+        <h1><center>Hmmmmmm, it is a tie!</h1>
+      <% else %>
+        <h1><center><%= @name %> wins! You got lucky!</h1>
+      <% end %>
+<br>
+<br>
+ <p class="lead">Do you want to play again?
+        <p><button type="submit" class="btn btn-primary">For sure!</button></p>
```
silvabox: There are no feature tests for these buttons...

spec/game_engine_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'game_engine'
+
+describe GameEngine do
+  context 'Playing Rock Paper Scissors' do
+    it "TIE - Player and CPU both choose scissors." do
+      subject.cpu_choice
```
silvabox: You call `subject.cpu_choice` on every test - but it doesn't do anything.

spec/game_engine_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'game_engine'
+
+describe GameEngine do
+  context 'Playing Rock Paper Scissors' do
+    it "TIE - Player and CPU both choose scissors." do
```
silvabox: Think about how your tests read here.  RSpec uses `it` to encourage descriptions relating to the described thing:
"it TIE - Player and CPU both choose scissors" doesn't really make sense as a sentence.

There are different ways to think about the tests for `GameEngine`, but the way you've done it here, you are only testing 1 tie, 1 lose, and 1 win, all the other permutations are untested.

You might think about a test such as:

`it 'knows when there is a tie'`

then randomize, the player's choice, but mock the computer's to be the same and then test it returns `:TIE`

silvabox:
Hi Jennifer, thanks for submitting your RPS challenge.  I have mixed feelings about it.  On the one hand, I'm pleased you've had a go with Bootstrap.  On the other hand, your code is largely untested, the structure is confused and your game doesn't actually work.  A lo-fi playable game is better than a hi-fi unplayable one.

But don't worry - these are all good mistakes to be making - it helps me to see how you are thinking, and that's what is important about these challenges!
:construction:

PR #70, opened 2015-05-18
Im trying...

zanetton:


spec/feature/rps_feature_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'game'
+
+feature 'I can play against a computer'do
```
ptolemybarnes: You're testing that these classes respond to certain methods, but why not test that what those methods return is what you want?

views/index.erb
```diff
@@ -0,0 +1,4 @@
+<form action="/game" method="post">
+<input type = "text" name = "name">
```
ptolemybarnes: I
 N
   D
    E
   N
  T
 A
    T
   I
 O
N
!

lib/game.rb
```diff
@@ -0,0 +1,24 @@
+class Game
+
+  def computer_choice
+    computer_choice = %w(Rock Paper Scissors).sample
```
ptolemybarnes: You needn't have these variable assignments here as you're just returning a value from an array, Rock, Paper, Scissors.

server.rb
```diff
@@ -0,0 +1,25 @@
...
+  get '/' do
+    erb :index
+  end
+
+  post '/game' do
+    session[:name] = params[:name]
+    @p1_choice = params[:choices]
+    game = Game.new
+    @round = game.result(@p1_choice)
+    @final_result = @round[0]
```
ptolemybarnes: What're these magic numbers, 0 and 1? :confused:

lib/game.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def player_choice
+    player_choice = %w(Rock Paper Scissors)
+  end
+
+  def result(p1)
```
ptolemybarnes: Not bad but try to refactor some of the logic here out into separate methods that are descriptively named.

ptolemybarnes:
Nice try! I'm enjoying your progress. Keep up the good work (and don't go back to finance).

PR #72, opened 2015-05-19
Charlie's submission

CharlieKenny:


lib/computer.rb
```diff
@@ -0,0 +1,8 @@
+class Computer
+
+  attr_accessor :moves
+  def initialize
+    @moves = ["rock","paper","scissors"].sample
```
ptolemybarnes: This instance variable is misleadingly named. "@moves" suggests an array of moves, but what we're getting back is a single move. Also why have this in the initialize method? It means that the computer can only ever make a single move.

lib/game.rb
```diff
@@ -0,0 +1,30 @@
...
+require_relative 'computer'
+class Game
+
+  attr_accessor :player1, :player2
+  def initialize player1,player2
+    @player1 = player1
+    @player2 = player2
+  end
+
+  def winner?
```
ptolemybarnes: This is a good start but needs refactoring. Think about moving some of the logic out to separate, more descriptive, methods.

features/rps_cucumber_test.feature
```diff
@@ -0,0 +1,21 @@
...
+  Then I should see "What's your name?"
+
+  Scenario: Entering the game
+  Given I am on the register page
+  When I click "Submit"
+  Then I should see "Rock","Paper" and "Scissors"
+
+  Scenario: Playing the game
+  Given I am on the game page
+  When I select "Rock"
```
ptolemybarnes: This test will sometimes fail as you're assuming that the Computer is going to choose scissors, but it's random.

lib/rps.rb
```diff
@@ -0,0 +1,45 @@
...
+  post '/game/play' do
+    @name = session[:name]
+    player_choice = params[:weapon]
+    @outcome = play_game(player_choice)
+    erb :gameplay
+  end
+
+  helpers do
+
+    def play_game(player_choice)
```
ptolemybarnes: Nice helper method!

ptolemybarnes:
See if you can get those tests passing :construction:

PR #75, opened 2015-06-28
final version of rps challenge

MassimilianoMura:
basic version of the game, it's a single player game against a random computer


spec/spec_helper.rb
```diff
@@ -1,10 +1,103 @@
-require 'byebug'
+# Generated by rspec-sinatra. (2015-06-27 18:26:41 +0100)
```
MassimilianoMura: :+1:

spec/player_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'player'
+
+describe Player do
+
+  let(:player) { Player.new }
```
MassimilianoMura: Unnecessary because `subject`is already equal to `Player.new``

spec/player_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  it 'has a list of options to choose' do
+    expect(player.options).to eq(['Rock', 'Paper','Scissors','Lizard','Spock'])
```
MassimilianoMura: Prefer symbols to strings as they are immutable and they take up less memory.

spec/player_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+
+  describe 'play' do
```
MassimilianoMura: Describing a method, use `#` to do this.

spec/player_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    it 'can choose one option from the list' do
```
MassimilianoMura: Test descriptions: be definite: ìt 'chooses one option from the list' do`

spec/player_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      expect(player.play('Rock')).to eq 'Rock'
```
MassimilianoMura: This doesn't tell me very much. Why if I say to a real-life Player 'Do rock!' should I expect him to say 'Rock!' back to me_

spec/player_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    it 'reise an error if the choice is wrong' do
+      expect{ player.play('Air') }.to raise_error 'choice not available'
```
MassimilianoMura: :+1:

spec/features/index_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'spec_helper'
+
+feature 'Starting a new game' do
+
+  scenario "asking for the user's name" do
+  visit '/'
+  expect(page).to have_content "Ready to play? Digit yor name and press enter!"
```
MassimilianoMura: :+1:

spec/features/index_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+end
+
+  scenario "user can login with his name" do
+  visit '/'
+  fill_in 'name', with: 'Massi'
+  click_button ('Submit')
+  expect(page).to have_content "Welcome Massi, Sheldon has challenged you to a game"
```
MassimilianoMura: :+1:

lib/player.rb
```diff
@@ -0,0 +1,21 @@
+class Player
+
+
+  attr_reader :options
+
+
+  def initialize
+
+    @options = ['Rock', 'Paper','Scissors','Lizard','Spock']
```
MassimilianoMura: Better to use symbols

lib/player.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  end
+
+
+  def play choice
+    fail "choice not available" unless @options.include? choice
```
MassimilianoMura: Use `raise` instead of `fail`

lib/rps.rb
```diff
@@ -0,0 +1,40 @@
+require 'sinatra/base'
+require './lib/game'
+require './lib/player'
+
+
+class RPSchallenge < Sinatra::Base
+
+  set :views, proc { File.join(root, '..', 'views') }
```
MassimilianoMura: :+1:

lib/rps.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  get '/' do
+    if $player == ''
```
MassimilianoMura: Avoid using global variables as a go-to. If yoj wanted to store the user for a little while, use the `session`:

```ruby
enable :sessions

get '/' do
  if session[:player].empty?
    erb :challenge
  else
    erb :index
  end
end
```

lib/rps.rb
```diff
@@ -0,0 +1,40 @@
...
+      erb :challenge
+    else
+      erb :index
+    end
+  end
+
+  get '/challenge' do
+
+    $player1 = params[:name]
```
MassimilianoMura: Use session

lib/rps.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  end
+
+
+  post '/challenge' do
+
+    @choice = params[:choice].to_s
+    player = Player.new
+    game = Game.new player
+    @played = game.challenge(player.play(@choice))
```
MassimilianoMura: I think the naming of `player.play` is not very clear. It doesn't do anything more than check that the choice is valid: so it should be called `player.check_valid_move(@choice)`

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+  def initialize player
+
+    @player = player
+    @options = ['Rock', 'Paper','Scissors','Lizard','Spock']
+
+  end
+
+
+  def challenge player_choice # example player.play('Rock')
+
```
MassimilianoMura: Refactor this!

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+
+
+  def random_choice
+    rand = rand (5)
+    return 'Rock' if rand == 0
+    return 'Paper' if rand == 1
+    return 'Scissors' if rand == 2
+    return 'Lizard' if rand == 3
+    return 'Spock' if rand == 4
```
MassimilianoMura: `[:rock, :paper, :scissors, :lizard, :spock].sample`

spec/game_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+describe Game do
+
+
+  let(:player) { double :player }
+  let(:game) { Game.new player }
+
+
+  it 'respond to challenge' do
+    expect(game).to respond_to(:challenge)
+  end
```
MassimilianoMura: You can do this in one line

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    @player = player
+    @options = ['Rock', 'Paper','Scissors','Lizard','Spock']
+
+  end
+
+
+  def challenge player_choice # example player.play('Rock')
+
+
```
MassimilianoMura: To test this, you don't need to stub the `player_choice` method - you need to stub the `random_choice` method.

You also don't need to use `self` here

lib/game.rb
```diff
@@ -0,0 +1,96 @@
...
+    if player_choice == 'Rock' && self.random_choice == 'Rock'
+      'Tie'
+    elsif player_choice == 'Rock' && self.random_choice == 'Paper'
+      'You lost! Paper covers Rock'
```
MassimilianoMura: Return simple values from the game logic and make them more complicated in the view.

Return a symbol like `:tie` and in the view, do `The game was a <%= @result %>`

MassimilianoMura:
A lot of good code here: good understanding of Testing principles. Your code is not strictly TDD and this is something you need to fix.

Next time you try this, try doing a BDD approach. You write feature tests first, then afterwards write unit tests only when you need to. This way your application will be much simpler (you don't need to write the engine before you write the interface).

An example is here: https://github.com/sjmog/rps-refactor

MassimilianoMura:
:construction:

PR #77, opened 2015-06-28
submit rps challenge

katsuraku:
- Issues getting 2 player to work - 2 player game works in irb but not in webapp.
- Some tests failing which I can't fix.
- Tried to introduce alternative so game can be 1 player but couldn't figure out how to do it with my existing code-base


silvabox:
Hi Kirsten, thanks for submitting your RPS challenge.  As we discussed, you had some problems with your initial design which compounded as you got further in.  This is one of those things that happens sometimes!  It can be very demotivating when your code just seems to get worse the more you try to fix it, but it happens to all of us.

When you've had a clean break from it, take another look and decide whether to fix it, or start again.  If you get stuck a second time, stop and ask for help.

:construction:

PR #81, opened 2015-06-29
rps

gerard-morera:


features/rps_web_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'spec_helper'
+
+feature 'starting the game' do
+  scenario 'asking the name' do
+    visit '/'
+    fill_in "name", with: 'Gerard'
+    click_button "log in"
+    expect(page).to have_content 'Hello Gerard'
```
gerard-morera: Lovely test.

features/rps_web_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+    visit '/'
+    fill_in "name", with: 'Gerard'
+    click_button "log in"
+    expect(page).to have_content 'Hello Gerard' 
+  end
+
+  scenario 'game set up' do
+    visit '/name'
+    click_button 'one player game'
+    expect(page).to have_content 'choose'
```
gerard-morera: :+1:

spec/comparator_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "comparator"
+
+class ClassComparator
+  include Comparator
+end
```
gerard-morera: Lovely way to test a module
```ruby
ComparatorWrapper
```
is a good name too.

lib/rock.rb
```diff
@@ -0,0 +1,18 @@
+class Rock
```
gerard-morera: This could be the same class as the ´Scissor´class.

Something like:

´´´ruby
class Shape
  include Comparable
  def initialize(shape_name)
    @shape_name = shape_name
  end

  def to_sym
    @name.to_sym
  end

  def <=> other
    Comparator.compare(self, other)
  end
end
´´´

gerard-morera: In our `Comparator`class:

```ruby
class Comparator
  def self.compare(shape, other_shape)
    # Look up shape, see how it compares to other shape
  end

  private
  
  def rules
    { rock: { rock: 0, paper: 1, scissors: -1 },
      paper: { rock: 1, paper: 0, scissors: -1 },
      scissors: { rock: 1, paper: -1, scissors: 0 }
    }
  end
end
```

lib/game.rb
```diff
@@ -0,0 +1,45 @@
+require 'byebug'
```
gerard-morera: Odd to see `byebug`in a class file

lib/game.rb
```diff
@@ -0,0 +1,45 @@
...
+require_relative 'player_creator'
+require_relative 'move_holder'
+require_relative 'comparator'
+
+class Game
+
+  attr_reader :player_1, :player_2
+  
+  def initialize(play_mode, player_creator: PlayerCreator,
```
gerard-morera: This line means you can do `Game.new(play_mode)` somewhere - so why ever worry about all the rest?

You are providing a complex default hash - this is **fine**, but to keep line 10 - 13 shorter, maybe do

```ruby
def initialize(play_mode, opts=default_opts)
  # do stuff
end

private

default_opts
  { player_creator: PlayerCreator, ... }
end
```

gerard-morera: The downside of your approach is that to change any one item in the hash, you must rewrite the whole hash:

```ruby
Game.new(some_mode, player_creator: OtherPlayerCreatorClass) # WILL NOT WORK
```

lib/comparator.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module Comparator
+                
+  def self.compair comparable_player_1, comparable_player_2
+    operation comparable_player_1, comparable_player_2
+  end
+
+  def self.operation comparable_player_1, comparable_player_2
+
+    if (ShapeCreator.call(comparable_player_1.first) > comparable_player_2.first)
```
gerard-morera: If you will use `first` on an array like this, where I don't know what the `comparable_player`is (maybe rename to `comparable_player_array`), use a private method to be clearer:

```ruby
ShapeCreator.call(fetch_player_name(comparable_player_1) > fetch_player_name(comparable_player_2))
```

gerard-morera:
There is some properly genius-level code here: I'm super-impressed by the methods you've used to get around some problems.

I think, however, that most of the problems were of your own creation: things got really complicated really fast. You can do a lot of great, complicated stuff: but **only** do this once you have written a simple solution to the required feature. So refactor.

Feature tests will help a lot. They help keep you focussed on delivering a feature in the simplest way.

BDD Feature Test --> Unit Test --> Code PASS THEN Refactor.

PR #83, opened 2015-06-29
rps challenge 

minhajraz:
I did not write test and I did not do the user experience part of the challenge. I rewrote my code five times and I really got stuck on writing the tests for constants,doubles and stubs. I need to redo it and write all the test and do the sinatra part.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -110,3 +110,6 @@ DEPENDENCIES
   rubocop
   rubocop-rspec
   sinatra
+
+BUNDLED WITH
+   1.10.4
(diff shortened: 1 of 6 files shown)


silvabox:
Hi Mohamed, thanks for submitting your RPS challenge.  As we discussed, unfortunately you got stuck on the API layer with `puts` and `gets` - remember _Separation of Concerns_ and so you haven't done any of the web interface, which was the point of the challenge.  We're doing more web stuff this week, so make sure you understand what's going on and, if you get time, see if you can have another go at RPS.

:construction:

PR #86, opened 2015-06-29
Submission

arnottmj:


silvabox:
Thanks for completing your RPS challenge!  As we discussed, you need to concentrate on separation of concerns and single responsibility.  Although you've managed to create a 2-player game - which is quite an achievement! - there's quite a lot of cross over in your design and some missing tests...

:checkered_flag:

PR #91, opened 2015-06-29
Still in progress, game is not actually playable quite yet

lroliphant:


silvabox:
Hi Lucy, thanks for submitting your RPS challenge!  As discussed, you're so nearly there!  You have a nice UI and you've written some nice code - but be careful that you don't unwittingly write vacuous tests (e.g. testing stubs(!) and using `visit` in the middle of feature tests.)

I reckon another hour with a clear mind and you'll have it cracked.

:construction:

PR #92, opened 2015-06-29
Sorryyy will finish this tonight! (Monday)

natstar93:


ptolemybarnes:
CI is failing but only because it's trying to run Cucumber, which you're not using. You could edit the .travis file to tell it not to run Cucumber.
:checkered_flag:

PR #94, opened 2015-06-29
Still bugged

jackhalltipping:
Can't figure out the problem with undefined method join string - ruby works - web worked until actually posted the result


silvabox:
Hi Jack, thanks for submitting your RPS challenge.  So nearly there!  Just a few tweaks and some CSS and it'll be complete.

:construction:

PR #95, opened 2015-06-29
Best of three

TheoLeanse:


spec/features/playing_oneplayer_game_spec.rb
```diff
@@ -0,0 +1,26 @@
+# how to stub computer's randomness in these feature tests?
+
+feature 'One-player game' do
+  scenario 'Starting a game' do
+    visit '/'
+    expect(page).to have_content 'Welcome'
```
TheoLeanse: 1 expectation clause per scenario

spec/features/playing_oneplayer_game_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    when_I_am_registered
+    3.times { i_choose_rock }
+    expect(page).to have_content 'Winner is: '
+    click_on 'play again?'
+    expect(page).to have_content 'Choose your weapon:'
+  end
+end
+
+private
+def when_I_am_registered
```
TheoLeanse: :+1: on these private functions - just like reading English, I love it

spec/features/playing_oneplayer_game_spec.rb
```diff
@@ -0,0 +1,26 @@
+# how to stub computer's randomness in these feature tests?
+
+feature 'One-player game' do
+  scenario 'Starting a game' do
+    visit '/'
+    expect(page).to have_content 'Welcome'
+    when_I_am_registered
```
TheoLeanse: This is crying out for a `context` block

spec/paper_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    expect(subject.result_against_scissors).to eq "Scissors wins against paper"
+  end
+
+  it 'wins against rock' do
+    expect(subject.result_against_rock).to eq "Rock loses against paper"
+  end
+
+  it 'draws against paper' do
+    expect(subject.result_against_paper).to eq "Paper draws against paper"
+  end
```
TheoLeanse: 1. There's repetition in the logic used here. This indicates a refactor would be a good idea.
2. You're returning entire strings where a shorter result would do.

Refactor:

```ruby
def result_against(move)
  case move
  when :scissors
    :lose
  when :rock
    :win
  else
    :draw
  end
end
```

TheoLeanse: You could have a `Rules` class with a constant for the rules (or a method):

```ruby
class Rules
  def game_rules
    { paper: { rock: :win, scissors: :lose, paper: :draw },
      rock: { paper: :lose, scissors: :win, rock: :draw },
      scissors: { paper: :win, rock: :lose, scissors: :draw }
    }
  end

  def lookup_winner(human_move, computer_move)
    # some sort of lookup in the game_rules method above
  end
end
```

app/rps_helper.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'sign_game'
+require_relative 'rock'
+require_relative 'paper'
+require_relative 'scissors'
+
+module RPSSetterUpper
```
TheoLeanse: Helpers are primarily designed to belong in Views.

app/RockPaperScissorsWeb.rb
```diff
@@ -0,0 +1,37 @@
...
+  get '/' do
+    erb :index
+  end
+
+  get '/register' do
+    erb :register
+  end
+
+  post '/register' do
+    register
```
TheoLeanse: This is a good thing to put in a helper, as it's essentially a private method from this class that we've moved 'elsewhere'.

app/RockPaperScissorsWeb.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  get '/play' do
+    erb :play_page
+  end
+
+  post '/play' do
+    @result = new_game.choose params[:choice]
+    update_status
+    check_for_winner
```
TheoLeanse: `WinnerCalculator.check_for_winner(session)`

TheoLeanse:
This is really awesome: the code is DRY, the methodology is secure and the modularity is very impressive. I love a lot of this.

Make sure you're clear on the purpose of helpers, and have a go from a BDD angle -> this will help to slim down and DRY up your feature tests (which are pretty mad at the moment).

Also check out the [Four-Phase testing methodology](https://robots.thoughtbot.com/four-phase-test).

PR #98, opened 2015-06-29
RPS challenge submission

duskyshelf:


silvabox:
Hi David, thanks for submitting your RPS challenge.  You have a really simple solution that works.  There are some opportunities to refactor and perhaps improve the design with a few more classes - for example separating the presentation concerns from the business logic.

:checkered_flag:

PR #105, opened 2015-08-09
Completed rps-challenge

RichardCharman:


spec/features/playing_the_game.rb
```diff
@@ -0,0 +1,139 @@
...
+feature 'I can play against the computer' do
+  scenario 'I can enter my name' do
+    visit '/'
+    click_link 'Play Against Computer'
+    expect(page).to have_content "What's your name?"
+    our_name="Richard"
+    fill_in "name", with: our_name
+    click_button 'Submit'
+    expect(page).to have_content "Hello #{our_name}"
+  end
```
sjmog: Nice test but it could fail for weird reasons. Each test should ideally make one assertion http://betterspecs.org/#single

spec/features/playing_the_game.rb
```diff
@@ -0,0 +1,139 @@
...
+  scenario 'Gives default name if none submitted' do
+    visit '/'
+    click_link 'Play Against Computer'
+    click_button 'Submit'
+    expect(page).to have_content "Hello Mr Anonymous"
+  end
```
sjmog: :+1:

spec/features/playing_the_game.rb
```diff
@@ -0,0 +1,139 @@
...
+  scenario 'I can select a hand to play' do
+    visit '/'
+    click_link 'Play Against Computer'
+    click_button 'Submit'
+    select "paper", :from => "choice"
```
sjmog: :+1:

spec/features/playing_the_game.rb
```diff
@@ -0,0 +1,139 @@
...
+    expect(page).to have_content "Mr Anonymous presented scissors"
+  end
+  scenario 'I can quit to start page' do
+    visit '/'
+    click_link 'Play Against Computer'
+    click_button 'Submit'
+    click_button 'Quit'
+    expect(page).to have_content "Play Against Computer"
+  end
+end
```
sjmog: :+1:

spec/features/playing_the_game.rb
```diff
@@ -0,0 +1,139 @@
...
+  scenario 'We can both select a hand to play' do
+    visit '/'
+    click_link 'Play Against A Friend'
+    click_button 'Submit'
+    select "paper", :from => "choice1"
+    click_button 'Present Choice'
+    select "rock", :from => "choice2"
+    click_button 'Present Choice'
+    expect(page).to have_content "Paper wraps rock - Player 1 wins!"
+  end
```
sjmog: This is really cool - stretch yourself and see if you can implement this using the session so we could play online multiplayer. Capybara can do it, but it's not easy.

spec/features/playing_the_game.rb
```diff
@@ -0,0 +1,139 @@
...
+    expect(page).to have_content "Mr Anonymous presented Spock"
+  end
+  scenario 'I can quit to start page' do
+    visit '/'
+    click_link 'Play Rock Paper Scissors Lizard Spock'
+    click_button 'Submit'
+    click_button 'Quit'
+    expect(page).to have_content "Play Against Computer"
+  end
+end
```
sjmog: I love this - very thorough and complete. Make sure you're also allowing the computer/versus random player to play a genuinely random move.

You could stub it like this:

```ruby
allow_any_instance_of(RandomMover).to receive(:play_random_move).and_return(:rock)
```

lib/rps_web.rb
```diff
@@ -0,0 +1,71 @@
+require 'sinatra/base'
+require "sinatra/session"
+
+class RockPaperScissors < Sinatra::Base
+  # force port 3000 for Nitrous
```
sjmog: Excellent time to leave a comment b/c the code is a bit bewildering/surprising

lib/rps_web.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+  
+  get '/play_computer' do
+    @name = params[:name] 
+    erb :play_computer
+  end
+  
+  post '/play_computer' do
+    @choice = params[:choice]
+    @computer = rand(3)
```
sjmog: This belongs in the app logic, not in the controller. Generating random stuff is business logic, not server logic.

So this should actually be something like

```ruby
@computer = Computer.new
@computer_choice = @computer.make_choice
```

lib/rps_web.rb
```diff
@@ -0,0 +1,71 @@
...
+  
+  get '/play_spock' do
+    @name = params[:name] 
+    erb :play_spock
+  end
+  
+  post '/play_spock' do
+    @choice = params[:choice]
+    @computer = rand(5)
+    erb :spock_result
```
sjmog: When using a `post`, it's canonical to `redirect` at the end of the action instead of rendering something. So

```ruby
post '/play_spock' do
  @choice = params[:choice]
  @computer = Computer.new
  @computer_choice = @computer.make_choice
  redirect "/result" # this will issue a GET request to /result
end
```

views/computer_result.erb
```diff
@@ -0,0 +1,77 @@
+<div style='width: 400px; height: 600px; border: 3px solid #b10d19; text-align:center; margin:0 auto; '>
+  <img src='http://4.bp.blogspot.com/-lYz_ZrnPp4Q/VcWhhY_wIqI/AAAAAAAAQv4/Cq2pe41m-64/s1600/rps%2B1.jpg' width="400">
+  <% session[:choice] = @choice %>
+  <% session[:computer] = @computer %>
```
sjmog: Avoid any sort of `session` manipulation or referencing in the view. Do this in the controller.

Both `session` and `params` belong in the controller, not anywhere else.

If you need to pass data around between layers (model - view - controller) then do it via instance variables. (`@`)

views/computer_result.erb
```diff
@@ -0,0 +1,77 @@
...
+  <p><font size="5"><%=session[:player1]%> presented <%=session[:choice]%></font></p>
+  <div style='width: 400px' >
+    <div style="float: left; width: 200px; height: 115px">
+    <% if session[:choice] == "rock" %>
+    <p><img src='http://4.bp.blogspot.com/-QKw7nAW6ywk/VcUrWGBVjLI/AAAAAAAAQvA/OIdn2Xb0JpY/s320/rps%2Brock%2B2.jpg' width="190" align="left" ></p>
+    <% elsif session[:choice] == "paper" %>
```
sjmog: If you have complex conditionals (i.e. anything that uses `elsif` once or more) you should move them into your controller - or even better, if you can, into your model layer.

views/index.erb
```diff
@@ -0,0 +1,11 @@
+<div style='width: 400px; height: 500px; border: 3px solid #b10d19; text-align:center; margin:0 auto; '>
+  <img src='http://4.bp.blogspot.com/-lYz_ZrnPp4Q/VcWhhY_wIqI/AAAAAAAAQv4/Cq2pe41m-64/s1600/rps%2B1.jpg' width="400">
+  <div style="float: left; width: 200px;">
+  <p><font size="6"><a href="/computer">Play Against<br />Computer</a></font></p>
```
sjmog: Avoid `br` as it is a presentational tag. Instead wrap content in something with the CSS `display: block;` and it will break the line after it.

views/play_computer.erb
```diff
@@ -0,0 +1,24 @@
+<div style='width: 400px; border: 3px solid #b10d19; text-align:center; margin:0 auto; '>
+  <img src='http://4.bp.blogspot.com/-lYz_ZrnPp4Q/VcWhhY_wIqI/AAAAAAAAQv4/Cq2pe41m-64/s1600/rps%2B1.jpg' width="400">
+    <p><% if @name != "" %>
```
sjmog: `if !@name.empty?`

or `unless @name.empty?`

sjmog:
Despite the lack of randomness stubbing (which I would still like to see) and the unusual placement of logic (which should be further 'down the stack' - i.e. logic in the view should move to the controller or model), this is funny, well presented, clever, NOICE

:checkered_flag:

PR #108, opened 2015-08-09
My Rock/Paper/Scissors Challenge

Yorkshireman:


lib/game.rb
```diff
@@ -0,0 +1,30 @@
...
+    @name = name
+  end
+
+  def play weapon
+    game_outcome(weapon, random_weapon)
+  end
+
+  private
+
+  def game_outcome player_weapon, opponent_weapon
```
Yorkshireman: Not sure how to remedy this?!

lib/game.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'random_weapon_generator'
+
+class Game
+  attr_reader :name
+  attr_accessor :opponent_weapon
+
+  include RandomWeaponGenerator
```
silvabox: including a module is probably not the right approach here.  `RandomWeaponGenerator` is not really reusable behaviour.  It would be better to pass in an object that provides a random choice - like a computer player for example.

silvabox:
Hi Andy, thanks for submitting your RPS challenge.  Your code is simple and clean.  I think you would need to introduce more classes to attempt the 2-player game, but for now it's OK.

Please check the [common issues](https://github.com/makersacademy/rps-challenge/wiki/Common-issues) as some of these apply to your code.

:checkered_flag:

PR #109, opened 2015-08-09
initial pull request

emily-jane:


features/rock_paper_scissors.feature
```diff
@@ -0,0 +1,28 @@
...
+  	When I press "Submit"
+  	Then I should see "What's your name?"
+
+  Scenario: Filling in a name
+    Given I am on the name page
+    And I fill in "name" with "Fred"
+    When I press "Submit"
+    Then I should see "Hello Fred, welcome to Rock Paper Scissors Spock Lizard!"
+
+  Scenario: Playing the game
```
silvabox: Does this belong in the `initiating a game` feature?

lib/player.rb
```diff
@@ -0,0 +1,30 @@
+class Player
```
silvabox: You have major indentation issues in this file!  This should not still be happening after week 3...

lib/player.rb
```diff
@@ -0,0 +1,30 @@
...
+	end
+
+def pick object
+	fail "Must pick rock, paper, scissors, spock or lizard" unless @options.include?(object)
+	component = computer
+	person = (options.index(object)) + 1
+	who_has_won(component,person)
+end
+
+def computer
```
silvabox: `computer` does not really belong inside `Player`

lib/player.rb
```diff
@@ -0,0 +1,30 @@
...
+	rand(1..5)
+end
+
+def who_has_won (component, person)
+	if component == person
+		"It's a tie!"
+	elsif component % 2 == person % 2
```
silvabox: How do I know this algorithm works?  I don't have time to work out the maths and compare every permutation and there are no tests for me to run.  Think about readability.  The maths may indeed work, but I can't check it easily.  Also think about the sustainability of your code - it's not clear elsewhere that the order of the values in the `@options` array is critical.  This could very quickly break and there are no tests to mitigate it.

silvabox:
Hi Emily, thanks for submitting your RPS challenge.  Please read through my comments and study the [common issues](https://github.com/makersacademy/rps-challenge/wiki/Common-issues).

You have a nice simple app, but it feels to me like it was not test-driven, which is a concern.

:construction:

PR #112, opened 2015-08-09
mvp

Aftermaths:


lib/player.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def initialize
+    @played = []
+    @last_played = ''
+  end
+
+  def play(rps)
+    @played << rps
+    case rps
+      when :rock then @last_played = 1
```
silvabox: Why are you using magic numbers here?   Why store `1` when you could store `:rock`?

lib/player.rb
```diff
@@ -0,0 +1,25 @@
...
+      when :paper then @last_played = 2
+      when :scissor then @last_played = 3
+      else
+        'Please pick rock, paper or scissor'
+    end
+  end
+
+  def reset_last_play
+    @last_played = ''
```
silvabox: Avoid switching between types - `@last_played` is numeric elsewhere and now it is an empty string.

lib/rps_web.rb
```diff
@@ -0,0 +1,82 @@
...
+    session[:player] = params[:first_name]
+    @name = session[:player]
+    redirect '/' if @name == ''
+    p params
+    p session
+    erb :index
+  end
+
+  get '/botgame' do
+    $game ? $game : $game = Game.new(Player)
```
silvabox: us lazy initialization `$game ||= Game.new(Player)`

lib/rps_web.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  get '/two_player_results' do
+    @player_1_count = $game.player_1.played.count
+    @player_2_count = $game.player_2.played.count
+    @winner = $game.player_1_win?
+    @player = session[:player]
+    erb :two_player_results
+  end
+
+  def bot_play
```
silvabox: This is business logic and does not belong in the controller.

lib/rps_web.rb
```diff
@@ -0,0 +1,82 @@
...
+    random_play = [:rock, :paper, :scissor].sample
+    $game.player_2.play(random_play)
+  end
+
+  def session_id_even?
```
silvabox: it's good to have refactored this into another method - but your method name simply reveals its implementation.  What is this method for?  Name methods according to the context of their purpose - not their implementation

lib/rps_web.rb
```diff
@@ -0,0 +1,82 @@
...
+    $all_session.index(session[:session_id]).even?
```
silvabox: Using odd and even in this way is somewhat arbitrary and not reliable

spec/game_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'game'
+
+describe Game do
+  let(:player_1) { double :player_1, play: nil }
+  let(:player_2) { double :player_2, play: nil }
+
+  it 'know if player lose the round' do
+    subject = Game.new(Player)
```
silvabox: 1. You are injecting `Player` class here so these tests are not isolated.
1. You should use RSpec `subject` method rather than assigning a local variable

spec/game_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'game'
+
+describe Game do
+  let(:player_1) { double :player_1, play: nil }
```
silvabox: Your doubles are never used

spec/player_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'player'
+
+describe Player do
+  it 'allows players to play' do
+    subject = Player.new
```
silvabox: This is redundant - `subject` is already a new `Player` because of RSpec's implicit subject.

views/index.erb
```diff
@@ -0,0 +1,21 @@
+<h2>Welcome to Rock Paper Scissors Online</h2>
```
silvabox: This should be 2 different views with the conditional controlled elsewhere.  Only presentational logic in views.

views/lobby.erb
```diff
@@ -0,0 +1,9 @@
+<head>
+  <meta http-equiv="refresh" content="3" >
+</head>
+
+<% if $all_session.count.odd? %>
```
silvabox: Your view should have no knowledge of sessions

views/multiplayer.erb
```diff
@@ -0,0 +1,24 @@
+<% if session_id_even? %>
```
silvabox: Why is there an if statement here - both branches are identical?

views/two_player_results.erb
```diff
@@ -0,0 +1,31 @@
+<head>
+  <meta http-equiv="refresh" content="3" >
+</head>
+
+<% if @player_1_count == @player_2_count %>
```
silvabox: This is controller logic

silvabox:
Hi Ken, thanks for submitting your RPS challenge.  Please study my comments above and also the [common issues](https://github.com/makersacademy/rps-challenge/wiki/Common-issues) as some of these relate to your code.

Your approach is problematic and does not really utilize OO properly.  Think carefully about SRP and separation of concerns.

:construction:

PR #113, opened 2015-08-09
Version 1 complete

ojlamb:


lib/controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  enable :sessions
+
+  get '/' do
+    erb :register
+  end
+
+  post '/gamepage' do
+    params[:name] == "" ? @name = "Player 1" : @name = params[:name]
+    session[:name] = @name
+    $game ? $game : $game = initialize_game
```
danldb: `$game ||= initialize_game`

lib/controller.rb
```diff
@@ -0,0 +1,39 @@
...
+class RPS < Sinatra::Base
+  set :views, proc { File.join(root, '..', 'views') }
+  enable :sessions
+
+  get '/' do
+    erb :register
+  end
+
+  post '/gamepage' do
+    params[:name] == "" ? @name = "Player 1" : @name = params[:name]
```
danldb: Prefer if else to ternary - it is more verbose, easier to understand, and takes up many lines, thus reminding you to avoid using too many of them.

lib/controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    outcome = $game.evaluate_game
+    if outcome == true
+      @win = true
+    elsif outcome == false
+      @win = false
+    elsif outcome == :draw
+      @win = "Draw"
+    end
+    @name = session[:name]
+    erb :gamepage
```
danldb: ### Fat controllers

lib/controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def initialize_game
+    game = Game.new
+  end
```
danldb: this method seems unnecessary - also deeply coupled to Game class (no dependency inversion).

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,25 @@
+class Game
+  attr_reader :user, :computer, :winner, :winning_pairs
+  def initialize
+    @winning_pairs = [[:Paper, :Rock],[:Scissors, :Paper],[:Rock, :Scissors]]
+    @user
+    @computer
+    @winner
+    @draw
+  end
```
danldb: Do you need all these variables?

Also use a private attr_accessor to avoid seeing `@` throughout your code.

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def selection(choice)
+    @user = choice.to_sym
+  end
+
+  def computer_choice
+    @computer = [:Paper, :Scissors, :Rock].sample
+  end
```
danldb: ### Not encapsulating the 'computer' in a separate class

spec/Features/gamepage_specr.rb
```diff
@@ -0,0 +1,19 @@
+feature "Playing the game" do
```
danldb: typo in the file name - did you see these tests fail?

Also the directory name should be `features` not `Features`.

spec/rps_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    expect(subject.computer).to eq computer
+  end
+
+  it "evaluates the game and declares winner, or draw" do
+    allow(subject.computer).to receive :Paper
+    allow(subject.user).to receive :Rock
+    subject.evaluate_game
+    expect(subject.winner).to eq subject.computer
+  end
+end
```
danldb: ### Not testing all of the game logic in unit tests

spec/rps_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    expect(subject.user).to eq :paper
+  end
+
+  it "randomly assigns computer selection" do
+    computer = subject.computer_choice
+    expect(subject.computer).to eq computer
+  end
+
+  it "evaluates the game and declares winner, or draw" do
+    allow(subject.computer).to receive :Paper
```
danldb: Should be using doubles instead of subject.computer and subject.user

danldb:
:construction: 

No sticker this week Owen :(

You did a great job of understanding the problem and crafting a solution. However your basic Ruby skills let you down.

Revisit attr_accessors, what they mean and why we use them.

Test first - see the test fail, then make it pass.

Be aware of file names - a type in a file name can mean your tests don't work.

Revisit doubles and how to use them.

Revisit Dependency Inversion principle.

Good luck!

PR #119, opened 2015-08-09
submission

giusepped:


lib/game.rb
```diff
@@ -0,0 +1,62 @@
+require_relative './player'
+
+class Game
+  attr_reader :p_1, :p_2, :played_matches, :beats
+
+  def initialize(p_1, p_2)
```
giusepped: Use dependency injection so that players are already initialized when starting a new game

lib/game.rb
```diff
@@ -0,0 +1,62 @@
...
+    else
+      2
+    end
+  end
+
+  def match_winner
+    return "#{p_1.name.capitalize} wins the match" if match_result == 1
+    return "#{p_2.name.capitalize} wins the match" if match_result == 2
+    "This match is a draw"
+  end
```
giusepped: This should also say which user used which move

lib/rps_web.rb
```diff
@@ -0,0 +1,91 @@
...
+      $player_2 = Player.new(params[:name].capitalize)
+      session[:id_2] = session[:session_id]
+    else
+     $player_1 = Player.new(params[:name].capitalize)
+     session[:id_1] = session[:session_id]
+    end
+    redirect '/registration' if session[:name] == ''
+    erb :mode
+  end
+
```
giusepped: use 'post' when changing parameters

lib/game.rb
```diff
@@ -0,0 +1,62 @@
+require_relative './player'
+
+class Game
+  attr_reader :p_1, :p_2, :played_matches, :beats
```
silvabox: don't use cryptic attribute names like `p_1` - prefer `player1` or `player_1` instead

lib/game.rb
```diff
@@ -0,0 +1,62 @@
...
+      0
+    elsif beats[p_1.move].include?(p_2.move)
+      1
+    else
+      2
+    end
+  end
+
+  def match_winner
+    return "#{p_1.name.capitalize} wins the match" if match_result == 1
```
silvabox: you should return the winning player object here and allow the view to decide on the message

lib/player.rb
```diff
@@ -0,0 +1,28 @@
+class Player
+  attr_reader :name, :move, :name, :wins, :losses, :draws
```
silvabox: `:name` is duplicated

lib/game.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def match move_1, move_2
+    @played_matches += 1
+    select_moves move_1, move_2
+    wins_incrementer
+    match_result
+    match_winner
+  end
+
+  def select_moves move_1, move_2
```
silvabox: This is confused - the players should be responsible for their own selections.  If the game has 2 players, then it should just compare the players' moves

lib/game.rb
```diff
@@ -0,0 +1,62 @@
...
+    @p_1 = p_1
+    @p_2 = p_2
+    @played_matches = 0
+  end
+
+  def match move_1, move_2
+    @played_matches += 1
+    select_moves move_1, move_2
+    wins_incrementer
+    match_result
```
silvabox: this method call is redundant

lib/game.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def match_winner
+    return "#{p_1.name.capitalize} wins the match" if match_result == 1
+    return "#{p_2.name.capitalize} wins the match" if match_result == 2
+    "This match is a draw"
+  end
+
+  def wins_incrementer
+    p_1.win if match_result == 1
```
silvabox: each call to `match_result` runs the comparison again - you should only need to compare the players once

lib/game.rb
```diff
@@ -0,0 +1,62 @@
...
+    elsif beats[p_1.move].include?(p_2.move)
+      1
+    else
+      2
+    end
+  end
+
+  def match_winner
+    return "#{p_1.name.capitalize} wins the match" if match_result == 1
+    return "#{p_2.name.capitalize} wins the match" if match_result == 2
```
silvabox: Again, you are calling `match_result` multiple times unnecessarily

silvabox:
Hi Giuseppe, thanks for submitting your RPS challenge.  Well done on trying to get the 2 player game working - that's quite a challenge - especially testing it.

However, your approach is somewhat problematic as you are not really utilizing OO to its main advantage.  Having special routes for each player is also likely to cause difficulties.  HTTP should be stateless and you are relying very heavily on server-side state and controller logic.

But don't worry - the important thing is to have had a go.  The thing to do now is to think about whether you could take a different approach which frees the controller from any logic other than simply handling requests and sending responses.

Please also study the [common issues](https://github.com/makersacademy/rps-challenge/wiki/Common-issues) as some of these apply to your code.

:construction:

spec/game_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+  let(:p_1) { double :p_1, name: 'player 1' }
+  let(:p_2) { double :p_2, name: 'COMPUTER' }
+  subject { Game.new(p_1, p_2) }
+
+  before do
+    allow(p_1).to receive(:make_move)
+    allow(p_2).to receive(:make_move)
+  end
+
+  context '#select_moves' do
```
silvabox: use `describe`s when describing methods - not `context`s

spec/game_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      allow(subject).to receive(:random_move) { :rock }
+      subject.select_moves(:scissors, :paper)
+    end
+  end
+
+  context '#match_result' do
+    it 'should return 0 if it\'s a draw' do
+      allow(p_1).to receive(:move) { :rock }
+      allow(p_2).to receive(:move) { :rock }
+      expect(subject.match_result).to eq 0
```
silvabox: Use of magic numbers!!!  `0`, `1` and `2`

spec/player_spec.rb
```diff
@@ -0,0 +1,29 @@
+require_relative '../lib/player'
+
+describe Player do
+  subject = Player.new 'examplename'
```
silvabox: You mustn't do this!  `subject` is now a local variable within the block instead of an RSpec method.  This will lead to unexpected behaviour.  Use `subject { Player.new 'examplename' }`.

spec/player_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  it 'should be initialized with a name' do
+    expect(subject.name).to eq 'examplename'
+  end
+
+  it 'should be able to make a move' do
+    expect(subject.make_move(:rock)).to eq :rock
```
silvabox: You are only testing that `make_move` returns `:rock` which is irrelevant - you need to be testing that it actually changes the state of the object.

PR #120, opened 2015-08-09
Trying to make it work with installed gem

GabyML:


silvabox:
Hi Gaby, thanks for submitting your RPS challenge.  I realize you got stuck trying to get the gem to work, but that would have been a good point to abandon the gem and write the game logic yourself.

You should always try to deliver at least the minimum business benefit (in this case a lo-fi 1-player game) even if it's crude.

:construction:

PR #122, opened 2015-08-09
Incomplete

richgeog:
Unable to get results and player name to appear. 1 error in cucumber failing too, unable to fix at this time.


lib/game.rb
```diff
@@ -0,0 +1,58 @@
...
+  OPTIONS = ['rock', 'paper', 'scissors']
+
+  attr_reader :player_selection, :opponents_selection
+
+  def initialize player_selection
+    @player_selection = player_selection
+    @opponents_selection = ['rock', 'paper', 'scissors'].sample
+  end
+
+  def result
```
silvabox: `result` is not a verb, so it's not clear that is method calculates the results, rather than just returning an existing result

lib/game.rb
```diff
@@ -0,0 +1,58 @@
...
+    return 'Draw' if player_selection == opponents_selection
+    if player_selection == 'rock'
+      rock_results
```
silvabox: I like the fact that you've refactored the if statements for readability.  However, this approach is still difficult to extend.  To add Lizard and Spock, you would need to change this class.

lib/game.rb
```diff
@@ -0,0 +1,58 @@
...
+    elsif player_selection == 'paper'
+      paper_results
+    elsif player_selection == 'scissors'
+      scissors_results
+    end
+  end
+
+  def winner
```
silvabox: It's good to refactor the winning message into another method, but should these be private?  Also, presentation logic is not the responsibility of the game.

lib/rock_paper_scissors_web.rb
```diff
@@ -0,0 +1,43 @@
+require 'sinatra/base'
+require './lib/game'
+
+class Rps_challenge < Sinatra::Base
+  set :views, Proc.new { File.join(root, "..", "views") }
+
+  get '/' do
+    'Hello rps-challenge!'
```
silvabox: this line is redundant

lib/rock_paper_scissors_web.rb
```diff
@@ -0,0 +1,43 @@
...
+    erb :home
+  end
+
+  get '/SignIn' do
+    @name = params[:name]
```
silvabox: will `params[:name]` ever exist at this point?

spec/game_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'game'
+
+describe Game do
+
+  describe '#winning combination' do
```
silvabox: the hash prefix is a convention for describing methods - it's not appropriate here.

lib/rock_paper_scissors_web.rb
```diff
@@ -0,0 +1,43 @@
...
+
+class Rps_challenge < Sinatra::Base
+  set :views, Proc.new { File.join(root, "..", "views") }
+
+  get '/' do
+    'Hello rps-challenge!'
+    erb :home
+  end
+
+  get '/SignIn' do
```
silvabox: we tend to prefer snake_case for paths i.e. `sign_in`

views/play_game.erb
```diff
@@ -0,0 +1,16 @@
...
+  <form action="/PlayGame" method="post">
+    Please select your piece of equipment:
+      <select name="Selection">
+        <option value="Rock"> Rock </option>
+        <option value="Scissors"> Scissors </option>
+        <option value="Paper"> Paper </option>
+      </select>
+      </form>
+      <br>
+      <form action='/Result'>
```
silvabox: I think you've confused your routes - you have a form within a form which has uncertain results.

lib/rock_paper_scissors_web.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  post '/PlayGame' do
+    @player_selection = params[:selection]
+    new_game = Game.new @player_selection
+    @opponents_selection = new_game.opponents_selection
+    erb :play_game
+  end
+
+  get '/Result' do
+    @player_selection
```
silvabox: `@player_selection` and `@opponents_selection` lines are redundant - what were you intending here?  Both values are `nil` in any case.

silvabox:
Hi Richard, thanks for submitting your RPS challenge.  Please see my notes and also compare your code to the [common issues](https://github.com/makersacademy/rps-challenge/wiki/Common-issues) as a number of them apply.

There are good bits and bad bits - and some fundamental errors - but they're all fairly easy to fix.  Well done for having a go with Cucumber though.

:construction:

PR #126, opened 2015-08-10
RPS Challenge

haroonhassan:
Please see my Readme file. I more or less finished with the 9bfc93d commit. But then made changes major changes to try something out. Please look at the 9bfc93d commit, but would like to discuss the approach I took after 9bfc93d. All explained in my Readme. Thanks.


silvabox:
Hi Haroon, thanks for submitting your RPS challenge.  You've taken an interesting approach and while I can see what you were trying to do, there are some fundamental problems with your code.  It's too complicated to go into here, perhaps we can discuss 1-2-1?

Please also take some time to compare the [common issues](https://github.com/makersacademy/rps-challenge/wiki/Common-issues) as some of these are relevant.

:construction:

PR #129, opened 2015-08-10
Submitting rps-challenge

JoshuaTatterton:
Here's my rock paper scissors website.


spec/features_spec.rb
```diff
@@ -0,0 +1,70 @@
+require "spec_helper"
+
+feature "Can register name" do
+  scenario "takes input name and displays it on the next page" do
+    visit "/"
+    expect(page).to have_content "Please enter your name so you can start playing Rock, Paper, Scissors"
```
sjmog: These could be two different specs

sjmog: Take a look at isolating your tests. http://betterspecs.org/#single

spec/features_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    name = "Gon"
+    fill_in "name", with: name
+    click_button "continue"
+    expect(page).to have_content "Welcome #{name}"
+  end
+  scenario "registers default name if none given" do
+    visit "/"
+    click_button "continue"
+    expect(page).to have_content "Welcome Player"
+  end
```
sjmog: :+1:

spec/features_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+end
+feature "playing a game of RPS," do
+  scenario "can start a game" do
+    visit "/"
+    click_button "continue"
+    click_link "Start Game"
+    expect(page).to have_content "Hi Player"
+    expect(page).to have_content "Please choose Rock, Paper or Scissors"
+  end
```
sjmog: :+1:

spec/features_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  scenario "can choose rock" do
+    c = (double :cpu, choice: "Rock")
+    allow(CPU).to receive(:new) { c }
```
sjmog: `allow_any_instance_of(CPU).to receive(:choice).and_return "Rock"`

spec/features_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    visit "/"
+    click_button "continue"
+    click_link "Start Game"
+    click_link "Rock"
+    expect(page).to have_content "draw"
```
sjmog: :+1:

spec/features_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+  scenario "can choose scissors" do
+    c = (double :cpu, choice: "Rock")
+    allow(CPU).to receive(:new) { c }
+    visit "/"
+    click_button "continue"
+    click_link "Start Game"
+    click_link "Scissors"
+    expect(page).to have_content "lose"
```
sjmog: :+1:

spec/features_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    c = (double :cpu, choice: "Rock")
+    allow(CPU).to receive(:new) { c }
+    visit "/"
+    click_button "continue"
+    click_link "Start Game"
+    click_link "Rock"
+    expect(page).to have_content "draw"
+  end
+  scenario "can choose scissors" do
+    c = (double :cpu, choice: "Rock")
```
sjmog: Prefer to use symbols instead of strings

spec/features_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    expect(page).to have_content "lose"
+  end
+  scenario "can choose paper" do
+    c = (double :cpu, choice: "Rock")
+    allow(CPU).to receive(:new) { c }
+    visit "/"
+    click_button "continue"
+    click_link "Start Game"
+    click_link "Paper"
+    expect(page).to have_content "win"
```
sjmog: :+1:

spec/features_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    click_button "continue"
+    click_link "Start Game"
+    click_link "Paper"
+    expect(page).to have_content "Play Again"
+    visit "/"
+    click_button "continue"
+    click_link "Start Game"
+    click_link "Scissors"
+    expect(page).to have_content "Play Again"
+  end
```
sjmog: :+1:

spec/cpu_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "cpu"
+
+describe CPU do
+  it "can choose rock" do
+    all = []
+    20.times { all << subject.choice }
+    expect(all.include?("Rock")).to be true
```
sjmog: This is not a good way to stub randomness.

Better is to stub the random result of something lower-level. For example, say you're using the following code:

```ruby
def choice
  [:rock, :paper, :scissors].sample
end
```

Then you would stub out the result something like this:

```
allow_any_instance_of(Array).to receive(:sample).and_return(:rock)
```

spec/rps_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rps"
+
+describe "rps" do
```
sjmog: `GameDecider`?

sjmog: This could be in a module `GameDecider` and called like this:

```ruby
GameDecider.decide_game
```

views/home.erb
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <link rel="stylesheet" type="text/css" href="styles.css">
+</head>
+<body>
+  <div>
+    <center>
```
sjmog: Don't use presentational tags any more. They went out a while ago.

to center stuff horizontally, use a `div`, give it a class `center-div` and then do:

```css
.center-div {
  margin: 0 auto;
}
```

views/home.erb
```diff
@@ -0,0 +1,22 @@
...
+      <% if @name %>
+        <h1>Welcome <%= @name == "" ? session[:name] = "Player" : session[:name] = @name %></h1>
```
sjmog: Avoid accessing the session in the view layer.

Instead, pipe it into the view layer using an instance variable set in the controller:

```ruby
@name = session[:name]
```

views/rps.erb
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <link rel="stylesheet" type="text/css" href="styles.css">
```
sjmog: :+1:

views/rps.erb
```diff
@@ -0,0 +1,41 @@
...
+</head>
+<body>
+  <div>
+    <center>
+      <% if @choice %>
+        <% cpu = CPU.new.choice %>
+        <% outcome = rps @choice, cpu %>
```
sjmog: Definitely this should be in the controller. Does not belong in the view. Complex logic should be pushed as far down the stack as possible.

Ideally in the controller:

```ruby
@choice = params[:choice]
cpu = CPU.new
@outcome = GameDecider.decide_game(@choice, cpu.choice)
```

Then just render the instance variables.

views/rps.erb
```diff
@@ -0,0 +1,41 @@
...
+        <% if outcome == "win" %>
+          <h1>Congratulations <%=session[:name]%>!</h1>
```
sjmog: Pass as `@name`

lib/rps_web.rb
```diff
@@ -0,0 +1,42 @@
...
+  
+  enable :sessions
+
+  get '/' do
+    erb :home
+  end
+
+  post "/" do
+    @name = params[:name]
+    erb :home
```
sjmog: Avoid rendering a template directly as a result of `POST`. It's more canonical to use `redirect` to a `GET` version.

```ruby
post "/" do
session[:name] = params[:name]
redirect "/" # this will GET / and then in that method you can fetch @name from the session
end
```

lib/rps_web.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  get "/rpscpu" do
+    erb :rps
+  end
+
+  get "/rpscpu/rock" do
+    @choice = "Rock"
+    erb :rps
+  end
```
sjmog: You could use a wildcard parameter:

```ruby
get "/rpscpu/:choice" do
  @choice = params[:choice].to_sym
  erb :rps
end
```

sjmog:
There's a lot to like here: great TDD, nice use of Capybara, cool HTML etc. and some innovative solutions.

Make sure you're focussing on separation of concerns (`session` staying in the controller etc) and stubbing randomness properly. In case of that 0.0000000286224% chance everything goes wrong :+1: 

:checkered_flag:

PR #138, opened 2015-09-18
Tim Harrison submission

trwh:
RPS!


spec/hand_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  context 'when taking an RPS desgination,' do
+
+    it 'can take a shape and return it.' do
+      subject.shape!(:rock)
+      expect(subject.shape?).to eq :rock
+    end
+
+    # Don't do this - breaks duck typing.
+    # it 'rejects an incorrect RPS shape.' do
+    #   expect { subject.shape!('rock') }.to raise_error 'This is not a recognised RPS shape.'
```
stephenlloyd: This shape? method returns a symbol. If its a boolean method it has to return a false.

stephenlloyd:
:construction:

trwh:
@stephenlloyd, thanks for the one-on-one feedback, I've acted on it now so could you take a look and consider marking this as complete, please?

Thanks!

Tim

spec/game_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+describe Game do
+
+  let(:rock_hand) { double :hand, shape: :rock }
+  let(:paper_hand) { double :hand, shape: :paper }
+  let(:scissors_hand) { double :hand, shape: :scissors }
+
+  context 'when running a game,' do
+
+    it 'detects a tie.' do
+      tied_game = Game.new(rock_hand, rock_hand)
```
stephenlloyd: you shouldn't use game. use `subject.new
`

spec/game_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'rspec'
+require_relative '../lib/game.rb'
```
stephenlloyd: `require 'game'` should work

spec/hand_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rspec'
+require_relative '../lib/hand.rb'
+
+describe Hand do
+
+subject { Hand.new('Tim') }
```
stephenlloyd: `described_class.new('Tim')`

lib/hand.rb
```diff
@@ -0,0 +1,21 @@
+class Hand
+  ALLOWED_SHAPES = [:rock, :paper, :scissors]
+  attr_reader :name, :shape
+
+  def initialize(name)
```
stephenlloyd: I would do `def initialize(name:)` look up what that does :)

rps-web.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  get '/new_multiplayer_game' do
+    @user_name = session[:user_name]
+    erb :new_multiplayer_game
+  end
+
+  post '/multiplayer_result' do
+
+    if $opponent_hand != nil
```
stephenlloyd: I really hate this `$` why's it being used. I know why it was used in the beggining but I would be really worried if where you are now lets this slip through. Hiring partners are really scared about things like this.

stephenlloyd:
This was one of the best RPS challenges I've seen. Really good job. All of my points are minor apart for the `$` variable point. THats quite a serious issue. Otherwise this couldn't be better :) :checkered_flag:

PR #139, opened 2015-09-18
Jongmin's rps-challenge

jongmin141215:


lib/player.rb
```diff
@@ -0,0 +1,41 @@
+class Player
+
+  def initialize
+    @rps = [:rock, :scissors, :paper]
```
roidriscoll: I would suggest that these choices are integral to the game. They are essentially the building blocks laid out by the rules and should at least be in a frozen constant as they are unlikely to change: `OPTIONS = [:rock, :scissors, :paper].freeze`

However, I would go a step further and suggest that these would belong in some kind of eternal `Rules` module.

lib/player.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def rock
+    rps[0]
+  end
```
roidriscoll: These methods seem to be overcomplicating the code unneccesarily

lib/player.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def paper
+    rps[2]
+  end
+
+  def random_rps
+    rps.sample
+  end
+
+  def compare(a, b)
```
roidriscoll: This method is a code smell. It looks very much like a case statement, which is something we try to avoid at all costs. Not only this, but `compare` is violating SRP. How could you split it out into separate methods in order to reduce complexity?

server.rb
```diff
@@ -0,0 +1,85 @@
+require 'sinatra/base'
+require './lib/player'
+
+class RPS < Sinatra::Base
+  enable :sessions
```
roidriscoll: A small point, but remember to use line spacing to make our code more readable.
```
class RPS < Sinatra::Base

  enable :sessions

  get '/' do
    erb :index
  end
```

views/player1_result.erb
```diff
@@ -0,0 +1,11 @@
+<div style="text-align: center; margin-top: 100px">
+  <% if @result.nil? %>
+    <h2>Waiting for your friend...</h2><br>
+    <a href="/player1_result" style="text-decoration: none">
+      See the result</a>
+  <% else %>
+    <h2><%= @player.compare(session[:player1_rps].to_sym,
+      session[:player2_rps].to_sym) %></h2>
```
roidriscoll: I would suggest that input sanitisation logic would be better placed in the application logic.

roidriscoll:
Hi Jongmin, great work here. I was particularly happy to see that you had implemented a multi-browser two player session. I have left some comments if you get a chance to refactor :) :checkered_flag:

PR #140, opened 2015-09-18
bat020's RPS – version 2 (with scoring and lizards)

bat020:


roidriscoll:
Hi Bats, first of all, this is really impressive. You’re clearly comfortable around code and are showing some very advanced understanding at this stage. That said, this was a difficult challenge to mark because it was rather tricky to read. Stephen and I paired while marking this, and we both had to pause several times and piece together the flow. Code should scream not conceal!

It's a good rule of thumb to use things for their intended purpose - `extend` is to primarily used to allow you to bring in modules. In this codebase, your use was somewhat unrelated to that purpose therefore quite dangerous.

You’ve also gone down the route of avoiding state, but because you are keeping score you then unavoidably mix in a bit of state, which is a code smell.

Major points:

* Impressive use of advanced HTML syntax (required, autofocus)
* I notice you’ve used the `class << self` syntax, yet you are also declaring `self.method_name` elsewhere. It’s usually preferable to be consistent, whichever approach you go after.
* Computer and Player could be separate class, which would:
* Keep track of player score
* Make a choice of weapon (and randomise if nothing given)
* Store name of player
* Be able to reset the score (losing the need for the score array)
* One should avoid redefining class methods - now if you want to use `extend` you may run into problems
* Put this in your .rspec file: `--order rand` and run it several times. Why are your tests sometimes failing?
* Mocks are preferable to seeding rand. It communicates intent more clearly.

Again, you've put together a really impressive challenge response (with decent UX on the final product) and I enjoyed going through it - but it's a good idea to concentrate on making your code as readable as possible, aiming for elegant simplicity above all else. Great job!

:checkered_flag:

PR #143, opened 2015-09-19
rps challenge

Proskurina:


lib/game.rb
```diff
@@ -0,0 +1,35 @@
+class Game
+
+  attr_reader :options, :result, :player1, :player2, :score1, :score2
+
+  def initialize
+    @result = {rock: 'scissors', scissors: 'paper', paper: 'rock'}
```
pitchinvasion: extract this to a constant

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+    @player1 = nil
+    @player2 = nil
+  end
+
+  def add_player(player)
+    (@player1 ? @player2 = player : @player1 = player) unless two_players?
```
pitchinvasion: Nested ternaries like this are pretty hard to understand for someone else looking at the code. Although can you think of ways of simplifying this? If there was just an array of players, would that simplify this line here?

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def computer_choice
+    player1.make_choice(result.keys).to_s if player1.is_a? Computer
```
pitchinvasion: Where possible you want to avoid checking for types - the issue here is that I would argue that is the Player/Computer who's responsibility it is to make a choice, if you moved this logic into those classes you would avoid the type check and the repetition in your code

lib/player.rb
```diff
@@ -0,0 +1,16 @@
...
+  attr_accessor :choice
+
+  def initialize(name)
+    @name = name
+    @score = 0
+  end
+
+  def win
+    @score +=1
+    name
```
pitchinvasion: Why does this return the name?

spec/computer_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'computer'
+
+describe Computer do
+
+  it 'can make a choice' do
+    options = ['rock', 'paper', 'scissors']
+    expect(options).to include subject.make_choice(options)
```
pitchinvasion: Does this really test that they have made a choice? The expectations need to be the subject you're testing, in this case the computer class

views/result.erb
```diff
@@ -0,0 +1,15 @@
+Your choice is <%=@player_choice%> <br>
+
+Computer's choice is <%=@computer_choice%><br>
+
```
pitchinvasion: This could be better marked up, more semantically. `<br>` tags should generally be avoided and you should research how to mark up a form more accessibly so people with limited sight can use your form too with a screen reader

pitchinvasion:
Generally this is good, nice to see you making an attempt to get it working where you play with a computer, although I think that some of the two player logic could have been cleaned up a bit more and would be worth a refactor at some point. But good work overall! :checkered_flag:

PR #144, opened 2015-09-19
Harriet's RPS Challenge

harrietc52:


lib/comp_player.rb
```diff
@@ -0,0 +1,9 @@
+class Computer
+
+  attr_reader :shape
+
+  def shape
+    @shape = [:rock, :paper, :scissors, :lizard, :spock][rand(0..2)]
```
roidriscoll: It would be good to put `shape` into a constant as it's unlikely to change, then you could make a `Computer.choose_move` method which returns your random shape.

lib/game.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class Game
+
+RULES = {
+  rock:     {rock: :draw,   paper: :loser,  scissors: :winner, lizard: :winner, spock: :loser},
+  paper:    {rock: :winner, paper: :draw,   scissors: :loser,  lizard: :loser,  spock: :winner},
+  scissors: {rock: :loser,  paper: :winner, scissors: :draw,   lizard: :winner, spock: :loser},
+  lizard:   {rock: :loser,  paper: :winner, scissors: :loser,  lizard: :draw,   spock: :winner},
+  spock:    {rock: :winner, paper: :loser,  scissors: :winner, lizard: :loser,  spock: :draw}
+}
```
roidriscoll: This is purely a question of taste, but I would be tempted to extract the tie logic out into a separate function to reduce complexity of the RULES hash. That said, this way does keep it all under a well-named data structure so your design choice is very respectable.

I would also suggest that perhaps the `SHAPES` constant I suggested earlier might live in here.

lib/game.rb
```diff
@@ -0,0 +1,26 @@
...
+  def play(player, computer)
+    @winner = RULES[player.shape][computer.shape]
+
+    if @winner == :draw
+      "It is a tie!"
+    elsif @winner == :winner
```
roidriscoll: `elsif` is a code smell. How would you refactor to reduce the branches of this conditional statement?

lib/game.rb
```diff
@@ -0,0 +1,26 @@
...
+       "#{player.name}, you are the winner!"
+    else
+      "#{player.name}, you are the loser!"
```
roidriscoll: Why? As a player, I am expected to accept defeat without knowing WHAT beat me! It would be a nice UX touch to include the name of the item that the computer picked for some extra context :)

lib/player.rb
```diff
@@ -0,0 +1,15 @@
+class Player
+
+  attr_reader :name, :shape
+
+  def initialize(name)
+    @name = name
+    @shapes = ["rock", "paper", "scissors", "lizard", "spock"]
```
roidriscoll: We're duplicating the work of creating `shapes`, and introducing a risk of error - if at some point one of the `shapes` array changes and the two classes don't agree, we could be in trouble.

lib/player.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def choose(shape)
+    fail "Please choose between rock, paper, scissors, lizard or spock." unless @shapes.include?(shape)
```
roidriscoll: '@shapes` doesn't need the `@` because you have declared it as an `attr_reader`.

lib/rps-web.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  get '/name' do
+    @name = params[:name]
+    session[:name] = params[:name]
+    @game = Game.new
+    session[:game] = @game
+    @player = Player.new(@name)
+    session[:player] = @player
+    @comp = Computer.new
+    session[:computer] = @comp
```
roidriscoll: Do we need all these assignments to variables which are going straight into the session? It seems like an extra step. If we need to pass a value to the view then by all means use an instance variable, but if it's going into the session anyway, it's an extra step and not really needed.

lib/rps-web.rb
```diff
@@ -0,0 +1,49 @@
...
+    @name = session[:name]
+    erb :choose_shape
+  end
+
+  get '/result' do
+    @player_choice = params[:shape]
+    session[:shape] = params[:shape]
+    session[:player].choose(session[:shape])
+    session[:winner] = session[:game].play(session[:player], session[:computer])
+    # p session[:computer]
```
roidriscoll: Remember to delete comments. Also, now that you have `binding.pry` in your toolbox, I expect you will be debugging using that in future ;)

spec/comp_player_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'comp_player'
+
+describe Computer do
+
+  it "can select rock, paper or scissors" do
+    expect(subject.shape).to eq(:rock).or eq(:paper).or eq(:scissors).or eq(:lizard).or eq(:spock)
```
roidriscoll: Again, if we had a `SHAPES` constant, it would be easy to test these more thoroughly by using the `include?` method.

views/new_game.erb
```diff
@@ -0,0 +1,25 @@
+<% if @name==nil %>
+  <form action="/name">
+    <p style="font-size: 50px">
```
roidriscoll: Try and keep your styles in an external css sheet

roidriscoll:
Hi Harriet - great work! Left some notes for you, if you'd like to refactor at some point. :checkered_flag:

stephenlloyd:
:checkered_flag:

PR #147, opened 2015-09-19
Usman - RPS Challenge

UsmanJ:


lib/computer.rb
```diff
@@ -0,0 +1,9 @@
+class Computer
+
+  attr_reader :auto
+
+  def auto
```
danldb: ### Encapsulating the computer into a separate class, but not making it a duck type with Player
The game does not need to know if it's comparing two players or one player vs a computer or even two computers. Computer and Player should share the same interface, so that one can be substituted for another.

lib/computer.rb
```diff
@@ -0,0 +1,9 @@
+class Computer
+
+  attr_reader :auto
```
danldb: This attr_reader is being overwritten by the `#auto` class below it. You do not need the attr_reader here.

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+    "That's a draw!"
+  end
+
+  def winner1
+    "Player 1 wins!"
+  end
+
+  def looser1
+    "Player 2 wins!"
+  end
```
danldb: Including presentation strings in business logic layer

Your Game class (or similar) should not return presentation strings like "Congratulations - you won!". This is a presentation concern and should be handled in another layer of code (separation of concerns). Instead, return representative codes, such as :win and :draw from the Game class which can be translated by the presentation layer.

This approach makes it possible to change the presentation layer (e.g. to add support for a different language) without changing the lower-level code (open/closed principle).

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+    elsif @winner == :computer_wins
+      looser1
+    else
+      winner1
+    end
+  end
+
+
+  # private
+
```
danldb: The below methods are unnecessary.

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+      draw
+    elsif @winner == :computer_wins
+      looser
+    else
+      winner
+    end
+  end
+
+  def play2(player1, player2)
+    @winner = RULES[player1.object][player2.object]
```
danldb: this method should not be necessary

lib/game.rb
```diff
@@ -0,0 +1,65 @@
...
+   spock:    {rock: :player_wins, paper: :computer_wins, scissors: :player_wins,
+     spock: :draw, lizard: :computer_wins},
+   lizard:   {rock: :computer_wins, paper: :player_wins, scissors: :computer_wins,
+     spock: :player_wins, lizard: :draw}
+  }
+
+  def play(player, computer)
+    @winner = RULES[player.object][computer.auto]
+
+    if @winner == :draw
```
danldb: these conditionals are not necessary

lib/player.rb
```diff
@@ -0,0 +1,10 @@
+class Player
+
+  attr_reader :object, :name
+
+  def choose(object)
+    objects = Game::RULES.keys
```
danldb: This creates a dependency on the Game - maybe the responsibility for checking the validity of the object should be with the game class.

`object` is too vague - be more specific in your naming, as object means almost any element of a ruby program.

rpsweb.rb
```diff
@@ -0,0 +1,113 @@
...
+class Rps_web < Sinatra::Base
+
+  enable :sessions
+  set :views, proc { File.join(root, 'views') }
+
+  get '/' do
+    erb :index
+  end
+
+  post '/' do
```
danldb: `post '/'` does not give any semantic information.  What is the purpose / responsibility of this route? It should be possible to express that purpose in the path.

rpsweb.rb
```diff
@@ -0,0 +1,113 @@
...
+  end
+
+
+
+  get '/start_game_double2' do
+    session[:player1].choose(session[:object1])
+    session[:player2].choose(session[:object2])
+    session[:winner] = session[:game].play(session[:player], session[:computer])
+    erb :game
+  end
```
danldb: Lots of repetition - your routes should be much simpler. Reduce repetition by only not having start_game and start_game2 - instead, use sessions to establish what the user should see.

spec/computer_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  it 'shoud select an item from the array' do
+    allow(subject).to receive(:auto).and_return(:rock)
+    expect(subject.auto).to eq :rock
+  end
+
+  it 'should only select one out of rock, paper, scissors, spock or lizard' do
+    expect(subject.auto).to eq(:rock).or eq(
+      :paper).or eq(:scissors).or eq(:spock).or eq(:lizard)
+  end
```
danldb:
```ruby
weapons = [:paper,:scissors,:rock]
expect(subject.auto).to satisfy { |weapon| weapons.include?(weapon) }
```

spec/game_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    allow(player).to receive(:choose).with(:scissors).and_return :scissors
+    allow(player1).to receive(:choose).with(:scissors).and_return :scissors
+    allow(player2).to receive(:choose).with(:rock).and_return :rock
+    allow(player3).to receive(:choose).and_return :paper
+
+    allow(computer).to receive(:auto).and_return(:scissors)
+  end
+
+  it 'should declare if there is a draw' do
+    player.choose(:scissors)
```
danldb: This is not doing anything - you cannot change the state of a double

spec/player_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+describe Player do
+
+  describe '#choose' do
+    it 'returns the chosen object' do
+      expect(subject.choose(:rock)).to be :rock
+      expect(subject.choose(:paper)).to be :paper
+      expect(subject.choose(:scissors)).to be :scissors
+      expect(subject.choose(:spock)).to be :spock
+      expect(subject.choose(:lizard)).to be :lizard
+    end
```
danldb: Only 1 expect per unit test

spec/player_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    it '@object is equal to spock if it is selected' do
+      subject.choose :spock
+      expect(subject.object).to eq :spock
+    end
+
+    it '@object is equal to lizard if it is selected' do
+      subject.choose :lizard
+      expect(subject.object).to eq :lizard
+    end
+  end
```
danldb: You could write a loop to make all these tests in a few lines of code.

rpsweb.rb
```diff
@@ -0,0 +1,113 @@
...
+    redirect ('/game_double')
+  end
+
+  get '/game_double' do
+    session[:player1].choose(session[:object1])
+    session[:player2].choose(session[:object2])
+    session[:winner_double] = session[:game].play2(session[:player1], session[:player2])
+    erb :game_double
+  end
+
```
danldb: Your controllers are really over complicated - remember to refactor and simplify where ever possible.

spec/feature/starting_rps_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    visit ('/')
+    click_button '2 Player'
+    fill_in('name1', with: "Usman")
+    fill_in('name2', with: "Jamil")
+    click_button 'Start Game'
+    click_button 'rock'
+    click_button 'rock'
+    expect(page).to have_content "That's a draw!"
+  end
+end
```
danldb: ### Not testing all game outcomes in feature tests
Although you do not need to test all possible combinations, your feature tests should test every possible outcome - i.e.:

- a win
- a loss
- and a draw.

spec/game_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'game'
+
+describe Game do
+
+  let(:player) {double :player, object: :scissors }
+  let(:player1) {double :player, object: :scissors }
+  let(:player2) {double :player, object: :rock }
+  let(:player3) {double :player, object: :paper }
+  let(:computer) {double :computer}
```
danldb: many doubles - I don't think you need all these.. you could probably just use a single double and set its object in an `allow` at the start of the test.

spec/game_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    player3.choose(:paper)
+    expect(subject.play2(player, player3)).to eq("Player 1 wins!")
+  end
+
+  it 'should declare player 2 is the winner' do
+    player1.choose(:scissors)
+    player2.choose(:rock)
+    expect(subject.play2(player1, player2)).to eq("Player 2 wins!")
+  end
+
```
danldb: ### Not testing all of the game logic in unit tests

danldb:
:construction: 

The controllers are greatly over-complicated. Although the code provides a solution, it does not solve the problem in a reusable or maintainable manner. Have a check through this: https://github.com/makersacademy/rps-challenge/wiki/Common-issues/ as many of the issues apply to this code.

PR #148, opened 2015-09-19
Gavin's pull request - rps challenge

gavinwcheng:


lib/rps_challenge.rb
```diff
@@ -0,0 +1,82 @@
...
+class RpsChallenge < Sinatra::Base
+  set :views, proc { File.join(root, '..', 'views') }
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  get '/username' do
+    p session
```
spike01: Remember to take code that is only concerned with debugging out of your code once you're done with it - you can see on your CI build that the test output is polluted with the output from this line

lib/game.rb
```diff
@@ -0,0 +1,25 @@
+class Game
+  attr_reader :options, :player1, :player2
+  attr_accessor :choice1, :choice2
+
+  def initialize player1, player2
+    @options = { 'r'=>'s', 's'=>'p', 'p'=>'r' }
```
spike01: Where possible, try not to abbreviate - it should be clear from your code what the "real life" concept you're modelling is, in this case the game of rock, paper and scissors.

Also, can you think why it would be better to use symbols rather than strings here?

lib/game.rb
```diff
@@ -0,0 +1,25 @@
...
+    @player1, @player2 = player1, player2
+    @choice1, @choice2 = '', ''
```
spike01: Setting anything to an empty string is generally a code smell - instead, if something doesn't exist, it may better be represented by `nil`

lib/game.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def all_choices?
+    (@choice1 != '') && (@choice2 != '')
```
spike01: So here, `@choice1.nil?` etc.

lib/game.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def play
+    if @choice1 == @choice2
+      'Draw!'
+    elsif @options[@choice1] == @choice2
+      "#{@player1} wins"
+    else
+      "#{@player2} wins"
+    end
+  end
```
spike01: So, a few things here:

- Try not to return strings: this mixes the fact that this class contains business logic as well as what will be presented to the user
- Assuming I'm refactoring without taking the point above into account, you could clean this up in a few ways - 1) using guard clauses

```ruby
def play
  return 'Draw' if @choice1 == @choice2
  return "#{@player1} wins" if @options[@choice1] == @choice2
  "{@player2} wins"
end
```

2) more clearly named methods

```ruby
def play
  return draw if draw?
  ...
end

def draw?
  @choice1 == @choice2
end
```

etc. for the other conditionals. And of course you can combine both of these techniques

lib/rps_challenge.rb
```diff
@@ -0,0 +1,82 @@
...
+
+private
+
+  def setup_usernames
+    $usernames ||= []
+  end
+
+  def setup_invites
+    $invites ||= {}
+  end
```
spike01: Yeah, these methods are an unfortunate side effect of using global variables.

lib/rps_challenge.rb
```diff
@@ -0,0 +1,82 @@
...
+    $usernames << params[:username]
+    redirect '/opponent'
+  end
+
+  get '/opponent' do
+    @username = session[:username]
+    @usernames = $usernames
+    setup_invites
+    @opponent = $invites[@username]
+    $invites[@username] ? erb(:accept_invite) : erb(:opponent)
```
spike01: If you're going to use parentheses around the argument to `erb`, be consistent

spec/features/rps_challenge_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+feature 'choose opponent' do
+  before do
+    $usernames = []
+    $invites = {}
+    $game = nil
+  end
+
+  scenario 'asked to choose an opponent after entering username' do
+    visit 'username'
```
spike01: It's not so much an issue here, but usually you'd `visit /username` to make sure you're visiting the absolute path

spec/features/rps_challenge_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    fill_in 'username', with: 'username1'
+    click_button 'Enter'
+    expect(page).to have_content 'Hi username1,'
+  end
+
+  scenario 'asked to wait for another player when no one else is online' do
+    visit 'username'
+    fill_in 'username', with: 'username1'
```
spike01: If you see something repeated 3 times (the two lines above), it's probably a sign that it's time to refactor this out

spec/features/rps_challenge_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+    end
+    expect(page).to have_content 'Please choose one of the three options:'
+  end
+end
+
+feature 'play' do
+  before do
+    $usernames = []
+    $invites = {}
+    $game = Game.new :username1, :username2
```
spike01: In a feature test, it's usually a bit more reliable to set up the test state by using user actions (in this case, clicking and filling stuff with Capybara) instead of reaching into the internals - there's a risk of you setting up a state that may not actually be possible through the interface

spec/game_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    expect(subject).to respond_to :play
+  end
+
+  it 'compares choices from 2 players to determine winner' do
+    game = Game.new :player1, :player2
+    game.choice1 = :r
+    game.choice2 = :r
+    expect(game.play).to eq 'Draw!'
+  end
+end
```
spike01: You don't seem to have tested a win condition or a loss condition

spec/game_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+describe Game do
+  subject { Game.new :player1, :player2 }
+
+  it 'responds to all_choices' do
+    expect(subject).to respond_to :all_choices?
+  end
+
+  it 'returns true if choce1 and choice2 are empty' do
+    game = Game.new :player1, :player2
```
spike01: This is the same setup as your `subject` block - do one or the other, not both

views/opponent.erb
```diff
@@ -0,0 +1,32 @@
...
+  <% else %>
+  Please choose an opponent from those online as shown below: <br><br>
+    <!-- <ul>
+      <% @usernames.each do | username | %>
+        <% if username != @username %>
+          <a href='/invite?opponent=<%= username %>' method='post'> <%= username %> </a> <br>
+        <% end %>
+      <% end %>
+    </ul>
+ -->
```
spike01: Always get rid of commented code! It either needs to be there or not.

lib/rps_challenge.rb
```diff
@@ -0,0 +1,82 @@
...
+    setup_usernames
+    $usernames << params[:username]
+    redirect '/opponent'
+  end
+
+  get '/opponent' do
+    @username = session[:username]
+    @usernames = $usernames
+    setup_invites
+    @opponent = $invites[@username]
```
spike01: It also seems a bit strange that you need two global variables to hold state on your server

lib/rps_challenge.rb
```diff
@@ -0,0 +1,82 @@
...
+    else
+      $game.choice2 = params[:choice]
+    end
+    if $game.all_choices?
+      session[:result] = $game.play
+      redirect '/result'
+    else
+      erb :wait
+    end
+  end
```
spike01: Try not to stack multiple conditionals in a route

lib/rps_challenge.rb
```diff
@@ -0,0 +1,82 @@
...
+    $invites[@username] ? erb(:accept_invite) : erb(:opponent)
+  end
+
+  post '/opponent' do
+    session[:opponent] = params[:opponent]
+    $invites[session[:opponent]] = session[:username]
+    redirect '/invite'
+  end
+
+  get '/invite' do
```
spike01: The invite concept, while cleverly implemented, adds complexity into the system and isn't specified in the brief

spike01:
:checkered_flag: An interesting implementation, although I'm not totally sold on the invite system or the page refreshes. Generally think about not relying heavily on if/elsif/else statements for control flow - there is almost always a better way.

spec/game_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  it 'responds to play' do
+    expect(subject).to respond_to :play
+  end
+
+  it 'compares choices from 2 players to determine winner' do
+    game = Game.new :player1, :player2
+    game.choice1 = :r
+    game.choice2 = :r
```
spike01: Your game does all its other comparison with strings, but here you're passing in symbols

PR #151, opened 2015-09-19
D Bebb - RPS

davidbebb:


lib/rps_web.rb
```diff
@@ -0,0 +1,69 @@
...
+
+class Rps < Sinatra::Base
+  register Sinatra::Partial
+  set :views, proc { File.join(root, "../", 'views') }
+  set :partial_template_engine, :erb
+  enable :sessions
+
+
+  get '/' do
+    redirect '/name'
```
silvabox: you could check here to see if a name is already in the session and redirect to `new_game` in that case, rather than forcing a name to be entered again

lib/rps_web.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  post '/name' do
+    session[:name] = params[:name]
+    redirect '/name'
+  end
+
+  get '/new_game' do
+    @name = session[:name]
+    redirect '/name' if @name=="" || @name.nil?
+    session[:game] = Rockpaperscissors.new
```
silvabox: don't add class instances to session.  Sessions should only contain strings.  The only thing that needs to persist between calls is the player's name

lib/rps_web.rb
```diff
@@ -0,0 +1,69 @@
...
+    @player2 = session[:player2]
+    erb :new_game
+  end
+
+  get '/play' do
+    @name = session[:name]
+    output = ""
+    output << partial(:name)
+    output << partial(:game)
+    output
```
silvabox: This is an unusual approach and is best avoided.  The controller's responsibility is to select the view, but not to compose it.  This should be in a `play.erb`

lib/rps_web.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  post '/play' do
+    move = params[:move]
+    session[:player1].choose(move)
+    session[:player2].choose
+    redirect '/winner'
+  end
+
+  get '/winner' do
+    @p1name = session[:player1].name
```
silvabox: You don't need all of these variables.  just have `@player1` and call `@player1.name` and `@player1.move` etc.

lib/rockpaperscissors.rb
```diff
@@ -0,0 +1,18 @@
+class Rockpaperscissors
```
silvabox: Careful of your naming.  This should be `RockPaperScissors` and `rock_paper_scissors.rb`

lib/rps_web.rb
```diff
@@ -0,0 +1,69 @@
...
+  post '/name' do
+    session[:name] = params[:name]
+    redirect '/name'
+  end
+
+  get '/new_game' do
+    @name = session[:name]
+    redirect '/name' if @name=="" || @name.nil?
+    session[:game] = Rockpaperscissors.new
+    @list_of_moves = session[:game].beaten_by.keys
```
silvabox: Your sort of defeating the object of OO here.  See my comments at the end

silvabox:
Hi David, thanks for submitting your RPS challenge.  It's a really strong submission - lots of tests and generally clean code.

Think about your `RockPaperScissors` class.  Is this really a class at the moment - or just a convenient place to store a hash and a method?  How could you refactor this to be more OO?  The point of OO is to encapsulate properties and their associated behaviour in object instances rather than just modules of code.  By simply passing native types (i.e. hashes and strings) between objects, your code is more procedural.

For example:  you have the equivalent of `game.beaten_by.keys`.  While this is slightly cryptic to read, it also forces me to know that `beaten_by` returns a hash, whose keys are the available choices for the game.  Instead, that knowledge should be hidden inside `RockPaperScissors` - maybe as a class method - `RockPaperScissors.moves`

Now consider why `RockPaperScissors` is even a class currently - what state or behaviour does it model?  To address that, why not instantiate it with a move, then you could do the following:
```
choice = RockPaperScissors.new(:rock)
choice.beats? :paper
```
Currently, it has a `winner` method that takes four arguments.  The class should have no concept of 'names' - that is part of the running context of the game and is handled by the controller - and methods should never have four arguments - that's usually a sign that something is wrong with the design.

Hope that's helpful - happy to talk to you more about it.

:checkered_flag:

lib/rockpaperscissors.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  attr_reader :beaten_by
+
+  def initialize
+    @beaten_by = {rock: [:paper, :spock],
```
silvabox: encapsulate the rules in a constant to avoid duplicate hashes with every instance

lib/rockpaperscissors.rb
```diff
@@ -0,0 +1,18 @@
...
+                  scissors: [:rock, :spock],
+                  paper: [:scissors, :lizard],
+                  lizard: [:scissors, :rock],
+                  spock: [:lizard, :paper]}
+  end
+
+  def winner(name1, choice1, name2, choice2)
+    return :draw if choice1==choice2
+    @beaten_by.values_at(choice1)[0].include?(choice2) ? name2 : name1
```
silvabox: This is very cryptic.  One of the goals of Ruby is maximum expression and readability.  This line requires far to much cognitive effort to parse.

lib/player.rb
```diff
@@ -0,0 +1,15 @@
...
+  attr_accessor :name
+  attr_reader :choice
+
+  def initialize(name, list_of_moves)
+    @list = list_of_moves
+    @name = name
+  end
+
+  def choose(weapon)
+    @list.include?(weapon.to_sym) ? @choice = weapon.to_sym : @choice = 'No known weapon'
```
silvabox: refactor out each responsibility.  This line of code has 2 responsibilities - the first can be extracted to another method for example:
```
@choice = valid_weapon?(weapon) ? weapon.to_sym : 'No known weapon
```
notice the slight change in the use of the ternary too.  Don't use ternaries to branch logic - only to switch between two values.

spec/feature/game_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'spec_helper'
+
+feature '/play asks you to select a weapon' do
```
silvabox: Think about your feature and scenario naming.  The feature might be 'Choosing a weapon' and the scenario 'I can choose rock'.  You shouldn't include the path as this is an implementation detail at this level.

PR #155, opened 2015-09-20
Dan St Paul RPSLS Challenge

danielstpaul:


lib/game.rb
```diff
@@ -0,0 +1,35 @@
+require_relative 'rpsls_rules'
+
+class Game
+
+  attr_reader :score, :shapes
+
+  def initialize(rules=RPSLS.new)
```
roidriscoll: Nice to see dependency injection.

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+    @rules = rules.outcomes
+    @shapes = rules.shapes
+    @score = [0, 0]
```
roidriscoll: Although this works well enough, I would prefer to see the separation of scores into players, as it would arguably be more clear than just an array where the scores are attributed in a sequence which is potentially easy to misconfigure.

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def result(throw1, throw2)
+    outcome = @rules[throw1.downcase.to_sym][throw2.downcase.to_sym]
```
roidriscoll: Your use of `.downcase.to_sym` twice suggests that you could extract this function out into some sort of data sanitisation method.

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+    if outcome == :win
+      win
+    elsif outcome == :lose
```
roidriscoll: `elsif` is a code smell. How would you refactor to reduce the branches of this conditional statement?

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def win
+    @score[0] += 1
+    'WINNER!'
+  end
+
+  def lose
+    @score[1] += 1
+    'LOSER!'
+  end
```
roidriscoll: Here you have two score totals, both of which are attributed to the player, and none to the computer. You've done a nice job of separating the rules out, so that the game can be extended easily to include different rule sets etc - but that great work is undone here by a restrictive scoring solution. You could perhaps implement a `Player` class which  keeps track of its own score, then you can open up the game to have as many players as you like without having to revisit your scoring system.

lib/rpsls_rules.rb
```diff
@@ -0,0 +1,15 @@
+class RPSLS
+  def outcomes
```
roidriscoll: One might argue that `outcomes` and `shapes` shouldn't change so therefore should be constants with the `freeze` method applied to stop the data mutating. Great to see you have extracted this out from everything else - I would suggest that since you shouldn't need to instantiate this, a module might work better.

This last one is purely a question of taste, but I would be tempted to extract the `tie` logic out into a separate function to reduce complexity of the outcomes hash. That said, this way does keep it all under a well-named data structure so your design choice is very respectable.

rpsls-web.rb
```diff
@@ -0,0 +1,51 @@
+require 'sinatra/base'
```
roidriscoll: Quiet hound! Sit!

rpsls-web.rb
```diff
@@ -0,0 +1,51 @@
...
+  set :views, proc { File.join(root, 'views') }
+
+#Home Page
+  get '/' do
+    erb :home_page
+  end
+#register player, initialize game & computer
+  get '/name' do
+    session[:game] = Game.new
+    session[:computer] = Computer.new
```
roidriscoll: In this situation, you can get away with it, but having the whole game stored in sessions is risky. Also, given the simplicity of the `Computer` class, it might make more sense as a module if there's no `Player` class.

rpsls-web.rb
```diff
@@ -0,0 +1,51 @@
...
+    erb :name
+  end
+
+  post '/name' do
+    session[:name] = params[:name]
+    redirect('/name') if params[:name].empty?
```
roidriscoll: Maybe this should go above `session[:name] = params[:name]` so you don't have an unnecessary assignment if `params[:name]` is empty?

roidriscoll:
Hi Dan - great work! Left a few pointers in case you want to refactor :)

:checkered_flag:

PR #156, opened 2015-09-20
Zhivko Siderov: RPS

zsid:


lib/options.rb
```diff
@@ -0,0 +1,19 @@
+class Options
+
+  attr_reader :choices
+
+  def initialize
+    @choices = []
+  end
+
+  def add_choice(*new_entries)
+    new_entries.each { |new_entry| choices << new_entry unless choice_added?(new_entry) }
```
spike01: This could probably be split up over several lines with `do...end`

rps_web.rb
```diff
@@ -0,0 +1,137 @@
...
+
+  get '/player2' do
+    $player2 = Player.new(session[:name2])
+    redirect '/multiplayer-type'
+  end
+
+  get '/multiplayer-type' do
+    $game_type = GameType.new
+    $game_type.add_scenario :rock, :scissors
+    $game_type.add_scenario :scissors, :paper
```
spike01: Is it really the controller's job to assemble the scenarios like this?

rps_web.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  get '/paper' do
+    session[:player_choice] = "paper"
+    redirect '/check'
+  end
+
+  get '/scissors' do
+    session[:player_choice] = "scissors"
+    redirect '/check'
+  end
```
spike01: You shouldn't need these intermediate routes just to set the session!

rps_web.rb
```diff
@@ -0,0 +1,136 @@
...
+
+  get '/check' do
+    if $player2.nil?
+      redirect '/result'
+    else
+      redirect '/multi-result'
+    end
+  end
```
spike01: And again, intermediate routes just to do conditional logic...

spec/features/start_game_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    click_button "Submit"
+    click_link("paper")
+    click_link("scissors")
+    expect(page).to have_content "Ben wins!"
+  end
+
+  scenario 'multiplayer game - player1 wins' do
+    click_link("Multiplayer")
+    fill_in("name", with: "Ben")
+    click_button "Submit"
```
spike01: Repeated 3 times = time to refactor

spike01:
:checkered_flag: This is a generally good submission, but has a few small problems:

- Usage of intermediate routes and sessions to handle multiplayer
- HTML (how many `<br>` tags in a row?) and inline CSS styling (which is a shame, because the Ruby syntax and structuring is otherwise very strong)
- The general user flow of using sessions for multiplayer: does this work properly for people using 2 different browsers?
- Making the system extremely flexible (good idea) to the point of the game of rock/paper/scissors or even RPSLS not easily visible at the logic level of the game (not a good idea)

PR #157, opened 2015-09-20
Harry Venables RPS challenge

hvenables:


lib/computer.rb
```diff
@@ -0,0 +1,9 @@
+class Computer
+
+  attr_reader :choice
+
+  def selection
+    @choice = [:rock, :paper, :scissors, :lizard, :spock].sample
```
roidriscoll: Prime candidate for a constant, that...

lib/game.rb
```diff
@@ -0,0 +1,33 @@
...
+  def result
+    return "Draw" if draw?
+    win? ? "Player 1 Win's" : "Player 1 Loses"
+  end
+
+  private
+
+  def draw?
+    @player_2_selection = player2.selection
+    player1.selection == @player_2_selection
```
roidriscoll: Is it necessary to perform the variable assignment in this method? As a pattern I am used to seeing this kind of thing:

lib/game.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def win?
+    WINNER[player1.selection].include?(@player_2_selection)
```
roidriscoll: I would expect to see `@player_2_selection` as an `attr_`

https://github.com/makersacademy/course/blob/master/pills/wrapping_instance_variables_in_accessor_methods.md

lib/player.rb
```diff
@@ -0,0 +1,9 @@
+class Player
+
+  attr_accessor :selection
+
+  def choice(selection)
```
roidriscoll: Or `attr_reader` even...

lib/rps-web.rb
```diff
@@ -0,0 +1,89 @@
...
+    erb :account
+  end
+
+  get '/rules' do
+    erb :rules
+  end
+
+  post '/account' do
+    if params[:game] == "one-player"
+      session[:game] = Game.new(Player, Computer)
```
roidriscoll: Be careful with putting a whole game in the session. Yes we can get away with it here, but it's dangerous, as you will have seen in Battleships Web ;)

lib/rps-web.rb
```diff
@@ -0,0 +1,89 @@
...
+      redirect '/one-player-game'
+    else
+      session[:game] = Game.new(Player, Player)
```
roidriscoll: Unless you mean to pass the same argument twice, it's pretty vital for code readability to differentiate. Don't make the dev after you look at external files any more than necessary! `Game.new(Player1, Player2)` would suffice.

lib/rps-web.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  get '/one-player-game' do
+    @user = session[:name]
+    @game = session[:game]
+    erb :one_player_game
+  end
+
+  post '/one-player-game' do
+    session[:selection] = params[:selection].to_sym
```
roidriscoll: Is this enough to sanitise the input effectively? What if I entered "roCk"?

lib/rps-web.rb
```diff
@@ -0,0 +1,89 @@
...
+    session[:game].player1.choice(session[:selection])
+    @result = session[:game].result
+    @computer = session[:game].player2.choice
+    erb :one_player_result
+  end
+
+  get '/two-player-game' do
+    @user = session[:name]
+    @game = session[:game]
+    erb :two_player_game_player1
```
roidriscoll: I see that you are doing a turn-by-turn game in the same browser window. Obviously this is problematic in terms of flow... How would you do it so that a second player could connect to your session using a different browser?

roidriscoll:
Hi Harry! Great work, really enjoyed playing through the game and you are demonstrating encouraging levels of understanding. I have left some notes for you, should you decide to refactor when you get a chance. :checkered_flag:

PR #158, opened 2015-09-20
Lucy's RPS Challenge

lucetzer:


lib/game.rb
```diff
@@ -0,0 +1,27 @@
+require_relative 'player'
+
+class Game
+
+  attr_reader :player, :computer
+
+  def initialize(player = Player.new, computer = Player.new)
```
pitchinvasion: nice dependency injection

lib/game.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def draw?
+    player.choice == computer.choice
+  end
+
+  def winner?
+    player.choice == :rock && computer.choice == :scissors ||
+    player.choice == :paper && computer.choice == :rock ||
+    player.choice == :scissors && computer.choice == :paper
```
pitchinvasion: Could you clean this up so Hound stops complaining? Hashes might be your friend here....

lib/game.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def play
+    return "It's a tie!" if draw?
```
pitchinvasion: You should avoid strings in your libraries, these should go in your views, instead you want your library to maybe just return a symbol indicating the result

lib/player.rb
```diff
@@ -0,0 +1,25 @@
+class Player
+
+  attr_reader :choice, :options
+
+  def initialize
+    @options = [:rock, :paper, :scissors]
```
pitchinvasion: This could be extracted to a constant as it never changes

lib/player.rb
```diff
@@ -0,0 +1,25 @@
...
+    @choice = ""
```
pitchinvasion: Again this shouldn't be a string, also why does this need to be an instance variable? Is the choice stored in the object? I would say no

lib/rps_web.rb
```diff
@@ -0,0 +1,47 @@
...
+    session[:player] = Player.new
+    session[:comp_player] = Player.new
+    session[:choice] = params[:choice]
+    redirect '/result'
+  end
+
+  get '/result' do
+    @player = session[:name]
+    session[:player].choose(session[:choice])
+    session[:comp_player].random_choice
```
pitchinvasion: There's a lot of logic here, could you extact some of these out into your libraries?

spec/game_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'game.rb'
+
+describe Game do
+
+  it "game ties if both players have the same hand as rock" do
```
pitchinvasion: Make sure your rspec reads properly, "it game ties" doesn't make sense, I would just say it "ties if...."

spec/game_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  it "player 1 loses if it has a weaker hand with paper" do
+    player1 = double("player", choice: :paper)
+    player2 = double("player", choice: :scissors)
+    subject = Game.new(player1, player2)
+    expect(subject.play).to eq "You lost!"
+  end
+
+  it "player 1 loses if it has a weaker hand with rock" do
+    player1 = double("player", choice: :rock)
+    player2 = double("player", choice: :paper)
```
pitchinvasion: DRY? You can extract to methods in tests too

pitchinvasion:
This is really great, a few minor tweaks to improve (and particularly focus on minimal logic in your controllers) but this is in general really clean and well thought out. Nicely tested too! :checkered_flag:

PR #159, opened 2015-09-20
RPS result

sarahkristinepedersen:
Capybara tests left out - have tried to write a few, but apparently I don't get the syntax yet. This is definitely something I'm going to read up on!


lib/computer_player.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def name
+    "Computer"
+  end
+
+  def return_throw
+    [:rock, :paper, :scissors].sample
+  end
+
+  def win!
```
danldb: no `!` is necessary here.

lib/computer_player.rb
```diff
@@ -0,0 +1,21 @@
+class ComputerPlayer
+
```
danldb: Excellent work isolating the computer player into its own class

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+    if BEATS[player_1_throw] == player_2_throw
+      player_1.win!
+      return "Congratulations, #{player_1.name} won this game! #{player_1.name} played #{player_1_throw}, #{player_2.name} played #{player_2_throw}."
+    elsif player_1_throw == player_2_throw
+      return "Oh, this is a draw! #{player_1.name} played #{player_1_throw}, #{player_2.name} played #{player_2_throw}."
+    else
+      player_2.win!
+      return "Sorry, #{player_1.name} lost this game! #{player_1.name} played #{player_1_throw}, #{player_2.name} played #{player_2_throw}."
+    end
+  end
```
danldb: ### Including presentation strings in business logic layer

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+
+
+  # def player_1
+  #   @player_1
+  # end
+
+  # def player_2
+  #   @player_2
+  # end
```
danldb: ### Not removing comments before committing
Old code should be deleted before you commit - it is distracting and makes your code hard to read. There is no reason to keep commented-out code - if you are commiting regularly, all your code will be in git so you can easily look back at how it looked before you made changes.

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+    player_2_throw = player_2.return_throw
+    if BEATS[player_1_throw] == player_2_throw
+      player_1.win!
+      return "Congratulations, #{player_1.name} won this game! #{player_1.name} played #{player_1_throw}, #{player_2.name} played #{player_2_throw}."
+    elsif player_1_throw == player_2_throw
+      return "Oh, this is a draw! #{player_1.name} played #{player_1_throw}, #{player_2.name} played #{player_2_throw}."
+    else
+      player_2.win!
+      return "Sorry, #{player_1.name} lost this game! #{player_1.name} played #{player_1_throw}, #{player_2.name} played #{player_2_throw}."
+    end
```
danldb: Great job extracting the rules into a constant. You could have further reduced the use of if/else statements - for instance, you could have made individual classes for each weapon type.

lib/human_player.rb
```diff
@@ -0,0 +1,23 @@
+class HumanPlayer
+
+  attr_reader :name, :wins
+
+  def initialize(name)
+    @wins = 0
+    @name = name
+    fail "You did not enter a name. Please enter your name." unless @name != ""
```
danldb: Not the best place for a fail - if the name has not been entered, you should catch that before you create a new player.

lib/human_player.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def set_throw(select_throw)
+    @select_throw = select_throw
+  end
```
danldb: I would name the variable `@throw` - it is more conventional to use a noun rather than an action.

lib/human_player.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def return_throw
+    @select_throw
+  end
+
+  def win!
+    @wins +=1
+  end
+
+end
```
danldb: Excellent use of duck typing - your player has the same external interface as your computer.

lib/rps_app.rb
```diff
@@ -0,0 +1,63 @@
...
+    if params["rock.x"]
+      session[:human].set_throw(:rock)
+    elsif params["paper.x"]
+      session[:human].set_throw(:paper)
+    elsif params["scissors.x"]
+      session[:human].set_throw(:scissors)
+    else
+      fail "Illegal throw"
+    end
+    redirect "/game"
```
danldb: ### Fat controllers

In this case your params look a bit strange. I your case I would expect to be able to write the following method:
```ruby
post "/submit_throw" do
  session[:human].set_throw(params[:throw])
  redirect "/game"
end
```

lib/rps_app.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  get "/game" do
+    @play = session[:game].play
+
+    @human_result = session[:human].wins
+    @computer_result = session[:computer].wins
+    @name = session["name"]
+    erb :game
+  end
```
danldb: Setting ruby objects as sessions is generally a bad idea as ruby objects can get very big.

spec/computer_player_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect(subject.return_throw).
+      to satisfy {|v| valid_throws.include? v}
+  end
+
+  it "counts the number of wins" do
+    subject.win!
+    subject.win!
+    expect(subject.wins).to eq(2)
+  end
+
```
danldb: good

spec/feature/play_game_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "spec_helper"
+
+feature "Starting a new game" do
+
+  scenario "asks the user to enter a name" do
+    visit "/enter_name"
+    fill_in("name", with: "Sarah")
+    click_button("Let's play!")
+    expect(page).to have_content "Welcome to Rock, Paper, Scissors! Begin the game by entering your name."
```
danldb: When you click 'let's play', you submit a post request to '/submit_name', which then redirects to '/select_throw' - so you should no longer see "begin the game by entering your name".

spec/feature/play_game_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  scenario "after user has entered a valid name" do
+    visit "/select_throw"
+    expect(page).to have_content "Hi Sarah! What is your throw?"
```
danldb: Remeber that every test exists in isolation - so in this scenario, you have not given your name yet. instead of `"Hi Sarah!"` you will get `"Hi #{nil}!"`

spec/game_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+describe Game do
+
+  subject do
+    Game.new(player_1, player_2)
+  end
+
+  it "decides if player 1 wins" do
+    player_1 = HumanPlayer.new("Sarah")
+    player_2 = HumanPlayer.new("Martin")
```
danldb: use doubles instead of real players

spec/game_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  it "counts the number of wins" do
+    player_1 = HumanPlayer.new("Sarah")
+    player_2 = HumanPlayer.new("Martin")
+    player_1.win!
+    player_1.win!
+    expect(player_1.wins).to eq(2)
+  end
+
+end
+
```
danldb: ### Not testing all of the game logic in unit tests

spec/feature/play_game_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  scenario "if no name is entered, user should be asked again to enter a name" do
+    visit "/enter_name"
+    fill_in("name", with: "")
+    click_button("Let's play!")
+    expect(page).to have_content "Please enter a name to begin the game."
+  end
+
+end
```
danldb: ### Not testing all game outcomes in feature tests
Although you do not need to test all possible combinations, your feature tests should test every possible outcome - i.e.:

- a win
- a loss
- and a draw.

spec/spec_helper.rb
```diff
@@ -11,4 +11,109 @@
 # run `open coverage/index.html` from the command line to view details
```
danldb: You managed to keep the original coveralls set up and also run rspec-sinatra - well done!

danldb:
:construction: 

Really good effort - you totally nailed the use of duck types for your computer opponent, and you made great use of a constant for your rules. Some of your tests were not passing, if you can fix those you can get your sticker. 

Well done!

PR #160, opened 2015-09-20
RPS challenge

emilyworrall:


silvabox:
Hi Emily, thanks for submitting your RPS challenge.  You have a playable game, which is good, but your approach is kind of lacking OO.  Your `game` class is just a convenient place to stick a hash of rules, which you then just reach into from everywhere else.  Ideally, the hash should be private to the game and the game should provide and interface over that (like you have done with `moves`).  Perhaps, instead of being `Game`, this class could be `Weapon`, which is instantiated with a move.  Then you could have something like:
```ruby
Weapon.new(player_throw).beats?(computer_throw)
```

:checkered_flag:

views/index.erb
```diff
@@ -0,0 +1,16 @@
+<div style="font-family:'Courier New'">
+  <style> h1 {text-align: center;}
```
silvabox: Do not include styles in you html files.  Styles should be extracted into CSS

views/result.erb
```diff
@@ -0,0 +1,25 @@
...
+          form {text-align: center;}
+          body {background-image: url("hands.png");
+          background-repeat: no-repeat;}
+  </style>
+    <body bgcolor="#66FFCC">
+
+      <h1>You chose <%= @player_throw %>!</h1>
+      <p>The result is...</p>
+
+        <% if @player_throw == @computer_throw %>
```
silvabox: Game logic should never be in the view.  It's acceptable to have presentation logic (i.e. choosing a message depending on the outcome), but the mechanics of determining the outcome (i.e. the algorithm) belong in the model.  Here, your view is _deciding_ the winner, rather than just _responding_ to it.  It's a subtle but important difference

PR #162, opened 2015-09-20
Kyle's RPS

KyleMacPherson:


KyleMacPherson:
Brandon recommended bootstrap and advised to be wary of multiple elusive statements

capybara-201509191709327671017857.html
```diff
@@ -0,0 +1,11 @@
+<div><center><font size=10>
+  Enter your move! Rock, paper or scissors?!</font><center></div>
+
+  <div>
```
pitchinvasion: Delete these!

lib/computer.rb
```diff
@@ -0,0 +1,13 @@
+class Computer
+
+  attr_reader :moves_array
+
+  def initialize
+    @moves_array = [:rock, :paper, :scissors]
```
pitchinvasion: If this never changes, you could extract this into a constant, rather than using a variable

lib/game.rb
```diff
@@ -0,0 +1,28 @@
+class Game
+
+  attr_reader :rock_hash, :paper_hash, :scissors_hash
+
+def initialize
+  @rock_hash = {:paper => "PAPER WINS", :scissors => "ROCK WINS", :rock => "DRAW", :lizard => "ROCK WINS", :spock => "SPOCK WINS"}
```
pitchinvasion: Same as above, extract to constant

lib/game.rb
```diff
@@ -0,0 +1,28 @@
...
+  @paper_hash = {:paper => "DRAW", :scissors => "SCISSORS WIN", :rock => "PAPER WINS", :lizard => "LIZARD WINS", :spock => "PAPER WINS"}
+  @scissors_hash = {:paper => "SCISSORS WIN", :scissors => "DRAW", :rock => "ROCK WINS", :lizard => "SCISSORS WIN", :spock => "SPOCK WINS" }
+  @spock_hash = {:paper => "PAPER WINS", :scissors => "SPOCK WINS", :rock => "SPOCK WINS", :lizard => "LIZARD WINS", :spock => "DRAW"}
+  @lizard_hash = {:paper => "LIZARD WINS", :scissors => "SCISSORS WIN", :rock => "ROCK WINS", :lizard => "DRAW", :spock => "LIZARD WINS"}
+end
+
+def result player1_move, player2_move
+  if player1_move == 'rock'
+    result = @rock_hash[player2_move.to_sym]
```
pitchinvasion: Could you refactor this so it didn't use lots of if/else? This sort of thing would be frowned upon in a tech test (as are case statements), as there is almost always a simpler solution - and there's definitely ways of doing this which would avoid this and mean you don't have all those different instance variables

lib/player.rb
```diff
@@ -0,0 +1,9 @@
+# class Player
+#
+#   attr_reader :possible_moves, :move, :name
+#
```
pitchinvasion: Delete this

lib/rps_web.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  get '/result' do
+    names = %w(rock paper scissors lizard spock)
+    unless names.include? params[:move]
+      session[:error] = "#{params[:move]} is not a valid move try again"
+      redirect '/form'
+    end
+    @name = params[:name]
+    @move = params[:move].downcase
+    game = Game.new
```
pitchinvasion: There's too much logic here in your controller, is there a way you can extract this out into your libraries? The controller's responsibility should be to do as little as possible

lib/rps_web.rb
```diff
@@ -0,0 +1,81 @@
...
+    get '/2pform' do
+      @error = session[:error]
+      erb :twoplayerform
+    end
+
+    post '/submit_name' do #this is telling it what to do when it receives a post request to /submit_name
+      names = %w(rock paper scissors lizard spock)
+      unless names.include? params[:move]
+        session[:error] = "#{params[:move]} is not a valid move try again"
+        redirect '/2pform'
```
pitchinvasion: Again this is is very messy, if you were coming to this code that someone else had written, would you consider this clear? How would you make it clearer?

spec/computer_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'computer.rb'
+
+describe Computer do
+
+  #it 'chooses rock, paper or scissors' do
+    #expect(subject.play).to
```
pitchinvasion: No tests? You need to do TDD! It's basically the main thing we're looking for

spec/game_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+describe Game do
+
+ it 'returns the winning move based on both players moves' do
+   game = Game.new
+   expect(game.result('rock', 'scissors')).to eq('ROCK WINS')
+ end
+
+ it 'returns the winning move based on both players moves including lizard/spock' do
+   game = Game.new
```
pitchinvasion: how is this different from the first test? are you checking both players here?

views/index.erb
```diff
@@ -0,0 +1,14 @@
+<head>
+<link rel="stylesheet" type="text/css" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css">
+</head>
+<body>
+<div>
+
+  <font size = 10> <center> Hello! Welcome to Rock Paper Scissors Lizard Spock.
```
pitchinvasion: These tags are old and deprecated, you should use semantic HTML and use CSS for styling

pitchinvasion:
It's good to see you've got this working and got all your feature tests written and passing, but this needs some work in terms of the game logic - by putting more logic out of your controllers and into your other classes you'll see the code clean up, this will also allow you to write more unit tests to test all your logic and will mean you're really taking advantage of OOP and things like single responsibility principle which is crucial for you to understand on your way to becoming an awesome junior developer. :construction:

PR #163, opened 2015-09-20
rps-challenge submission

Harryandrew:


lib/rps_challenge.rb
```diff
@@ -0,0 +1,39 @@
+require 'sinatra/base'
+require_relative 'rps_helper'
+
+class Rps_Challenge < Sinatra::Base
+  set :views, proc {File.join(root,'..','/views')}
+  set :public, proc {File.join(root,'..','/public')}
```
roidriscoll: Warning when running `rackup`: `:public is no longer used to avoid overloading Module#public, use :public_folder or :public_dir instead`

lib/rps_challenge.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  get '/new_game' do
+    erb :new_game
+  end
+
+  get '/new_session' do
+    redirect '/new_game' if params[:name] == ""
+    session[:name] = params[:name]
+    $name = session[:name]
```
roidriscoll: We shouldn't need a global for `name` here as it is in your session...

lib/rps_game/cpu_player.rb
```diff
@@ -0,0 +1,10 @@
+class Cpu
+  attr_reader :last_move
+  
+	def cpu_move			
+    choice = rand(0..2)
```
roidriscoll: Indentation!

lib/rps_game/cpu_player.rb
```diff
@@ -0,0 +1,10 @@
...
+    rps = [:rock,:paper,:scissors]
+    rps[choice]
+    @last_move = rps[choice]
+	end
```
roidriscoll: This method is a bit overcomplicated and is screaming out for a refactor. I think if you were to extract `rps` into a constant that would take a load off, whilst also being sound practice - but can you make this method less complex?

lib/rps_game/game.rb
```diff
@@ -0,0 +1,47 @@
...
+
+# P1/P2_move required???
+  attr_reader :rules, :p1_move, :p2_move, :winning_move, :p1_score, :p2_score
+
+  def initialize
+    @rules = {
+    rock: {rock: :draw, paper: :paper, scissors: :rock},
+    paper: {rock: :paper, paper: :draw, scissors: :scissors},
+    scissors: {rock: :rock, paper: :scissors, scissors: :draw}
+  }
```
roidriscoll: Is this likely to change? If not, perfect candidate for a constant. Remember to `.freeze`!

lib/rps_game/game.rb
```diff
@@ -0,0 +1,47 @@
...
+    @p1_score = 0
+    @p2_score = 0
+  end
+
+  def p1_move(move)
+    @p1_move = move.to_sym
```
roidriscoll: Is this enough to sanitise the input? What if I entered `"rOck"`?

lib/rps_game/game.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def p2_move(move)
+    @p2_move = move.to_sym
+  end
+
+  def play
+    @winning_move = @rules[@p1_move][@p2_move]
+    update_score
+    return :p1_wins! if @p1_move == @winning_move
+    return :p2_wins! if @p2_move == @winning_move
```
roidriscoll: this is a bit of a violation of SRP. While `play` is certainly wide open in terms of scope, our method here is responsible for calculating the winning move, updating the score AND informing the player of a loss or win. Phew.

roidriscoll:
Hi Harry, good work - the game was fun to play. From a UX perspective it would be good to see my name used, otherwise as a user I am confused as to why it was required. Also, I like the idea that you restrict the game to 5 turns - perhaps it would be good to have an overall winner/loser message to mark the end of the game?

Also there are a disturbing amount of global variables. I appreciate that Battleships Web introduces these as a viable solution but at the same time we have them going on everywhere in your code. I am marking the challenge as complete because the game is very playable, but I would caution that a refactor is needed, and I have left some notes to assist. :checkered_flag:

Harryandrew:
Thanks Roi, well noted, Sanjay touched on this also, will go through it
over the weekend.

Harry Dix
*harryandrew.dix@gmail.com <harryandrew.dix@gmail.com>*
(+44)*7966 545 802*

PR #167, opened 2015-09-21
first release

TJQKAs:


spec/spec_helper.rb
```diff
@@ -1,10 +1,103 @@
...
-require 'coveralls'
-require 'simplecov'
-
-SimpleCov.formatters = [
-  SimpleCov::Formatter::HTMLFormatter,
-  Coveralls::SimpleCov::Formatter
-]
-Coveralls.wear!
\ No newline at end of file
+require 'rspec'
```
danldb: ### Overwriting spec_helper
When using generators such as `rspec-sinatra` beware that spec_helper.rb will be rewritten. Make sure you make a copy of all the pre-written CI code, otherwise you will break your coveralls CI, causing silent failure of your pull request.

lib/comp.rb
```diff
@@ -0,0 +1,9 @@
+class Comp
+
+  attr_reader :choice
+
+  def choice
+    @choice = [:rock, :paper, :scissors].sample
+  end
+
+end
```
danldb: Great that you have made a separate computer player. BE careful of your naming - I wasn't immediately sure what this class was doing. If you are building a computer player class, maybe call it a computer_player or similar.

Great that you have created a duck type by using the method name `choice` in both your computer and your player. Beware that you have used an `attr_reader :choice` and a choice method - one is overwriting the other.

lib/player.rb
```diff
@@ -0,0 +1,15 @@
+class Player
+
+  attr_reader :name, :choice, :choices
+
+  def initialize(name)
+    @name = name
+    @choices = ["rock", "paper", "scissors"]
+  end
+
+  def select1(choice)
```
danldb: again - be careful of your method naming. Why is this `select1`?

lib/server.rb
```diff
@@ -0,0 +1,55 @@
...
+   post '/request' do
+     session[:name] = params[:name]
+    #  @game = Game.new
+     session[:game] = Game.new
+    #  @player = Player.new(session[:name])
+     session[:player] = Player.new(session[:name])
+    #  @comp = Comp.new
+     session[:comp] = Comp.new
+     redirect '/game'
+   end
```
danldb: Don't leave commented code in your commit.

lib/server.rb
```diff
@@ -0,0 +1,55 @@
...
+     erb :index
+  end
+
+  get '/request' do
+     erb :new_game
+   end
+
+   post '/request' do
+     session[:name] = params[:name]
+    #  @game = Game.new
```
danldb: setting objects in your sessions is bad practice - sessions are very small and can easily overflow. Better to use a single global variable to hold the state of your game.

spec/comp_spec.rb
```diff
@@ -0,0 +1,18 @@
+# require 'cpu'
+# describe Cpu do
+#   let(:cpu) { described_class.new }
+#
+#   it 'can make a random selection' do
+#     expect(cpu.cpu_choice).to be==(double :Scissors)
+#   end
+# end
```
danldb: Try not to commit comments - especially old code, which should never ever appear in a git commit.

spec/features/starting_a_game_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    expect(page).to have_content "What's your name?"
+  end
+
+
+  # scenario 'After player wrote his name, he can see his name and prompt' do
+  #    visit '/request'
+  #    fill_in('name', with: 'John')
+  #    click_button('submit')
+  #    expect(page).to have_content "John! press button to start the game"
+  #  end
```
danldb: old code, shouldn't be here

spec/features/starting_a_game_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+scenario 'After player made choice he is able to see the results' do
+   visit '/game'
+   click_button('scissors')
+   visit '/result'
+   expect(page).to have_content "scissors"
+end
```
danldb: This test is currenlty failing

views/new_game.erb
```diff
@@ -0,0 +1,16 @@
+<% if @name==nil %>
+  <form action="/request" method="post">
+    <p style="font-size: 25px; font-family: Times New Roman">
+      What's your name?
+    </p>
+    <input type="text" name="name" required>
+    <input type="submit" value="submit">
+  </form>
+<!-- <% else %>
```
danldb: you need to close the `if` with and `<% end %>`

danldb:
:construction: 

Good effort - this is almost a pass. I have highlighted a few improvements for you, and would also direct you to this https://github.com/makersacademy/rps-challenge/wiki/Common-issues as a lot of these issues apply to you.

PR #168, opened 2015-09-21
Chris's challenge

chweeks:


RpsWeb.rb
```diff
@@ -0,0 +1,53 @@
+require_relative 'lib/player'
```
chweeks: ?

RpsWeb.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  get '/result' do
+    choice = session[:choice]
+    game = Game.new
+    player1 = Human.new
+    player2 = Computer.new
+    player1.chooses(choice.to_sym)
+    player2.rand_choice
+    @choice = player1.hand
```
silvabox: rather than `@choice` and `@comp_choice` you should have `@player1` and `@player2` and call `@player1.hand` from the view.

lib/game.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'human'
+require_relative 'computer'
+
+class Game
+
+  def winner?(human, computer)
+
+    winner = { rock: [:scissors, :spock],
```
silvabox: every time `winner?` or `loser?` is called, another hash object is created.  These rules only need to be defined once - maybe in a constant.  It is unnecessary to define both winner and loser hashes as they are just the inverse of one another.

lib/human.rb
```diff
@@ -0,0 +1,9 @@
+require_relative 'player'
+
+class Human < Player
+
+  def chooses(choice)
+    @hand = choice
+    fail 'Invalid Choice' if options.include?(choice) == false
```
silvabox: prefer `unless` to test negative conditions i.e.:
```ruby
fail 'Invalid Choice' unless options.include?(choice)
```

lib/computer.rb
```diff
@@ -0,0 +1,8 @@
+require_relative 'player'
+
+class Computer < Player
+
+  def rand_choice
```
silvabox: is this method necessary? just initialize the hand when the instance is created.

lib/player.rb
```diff
@@ -0,0 +1,9 @@
+class Player
+
+  attr_reader :hand, :options
+
+  def initialize
+    @hand = nil
+    @options = [:rock, :paper, :scissors, :lizard, :spock]
```
silvabox: This is another duplication of the list of moves.

lib/human.rb
```diff
@@ -0,0 +1,9 @@
+require_relative 'player'
+
+class Human < Player
+
+  def chooses(choice)
```
silvabox: prefer `choose` - it's more intention revealing

spec/game_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'game'
+
+describe Game do
+
+  let(:goku) {double :goku, hand: :rock}
+  let(:krillin) {double :krillin, hand: :spock}
+
+  it 'should return true if human is winner' do
```
silvabox: what should return true?  Your test wording is a bit confusing - it suggests the game should return true for all outcomes, which doesn't make sense.  Also, avoid 'should' - you can have `it `returns ...` instead.  Perhaps part of the problem is that `game.winner?` implies that the game itself is the winner - or that the game _has_ a winner, rather than that the first argument passed beats the second argument passed.  This is a common design problem.

silvabox:
Hi Chris, thanks for submitting your RPS challenge.  Well done - you have a fully playable game.  Please check my comments above.  Also, be careful with your UI - it was very difficult to read on my screen!

:checkered_flag:

PR #170, opened 2015-09-21
Nat's RPS Pull Request v1.0

nathanielgreen:


spec/spec_helper.rb
```diff
@@ -7,8 +7,10 @@
...
+Coveralls.wear!
 # run `open coverage/index.html` from the command line to view details
-
 require 'byebug'
-require 'capybara/rspec'
\ No newline at end of file
+require 'capybara/rspec'
+require 'capybara'
+
+Capybara.app = RPSWeb
```
danldb: You forgot `require './lib/RPS_web'` so your tests were not running

RPS_web.rb
```diff
@@ -0,0 +1,32 @@
...
+    $game = Game.new
+    erb :index
+  end
+
+  get '/newgame' do
+    erb :newgame
+  end
+
+  get '/gamepage' do
+    @commentary = ["AND IT'S A CLEAN LEFT HOOK, HIS OPPONENT'S KNOCKED STRAIGHT OUT OF THE RING, THERE'S NO COMING BACK FROM THIS ONE",
```
danldb: You should not be setting messages for the end user here. The responsibility for the view is the view template. What happends if the designer wants to change the message? In this case they would have to go into the back end code, which may not be a domain they are comfortable with.

RPS_web.rb
```diff
@@ -0,0 +1,32 @@
...
+                   "REFEREE'S TRYING TO BREAK THIS ONE UP, THAT'S A MEAN UPPERCUT! HE CAN'T SEE STRAIGHT! IT'S ALL OVER!",
+                   "WATCH OUT, WATCH OUT, WATCH OU- RKO! RKO FROM OUTTA NOWHERE! RANDY ORTON!",
+                   "AND THERE IT IS! DELIVERED! ASK YOU SHALL RECEIVE! WE SAW THIS ALL ON MONDAY NIGHT!",
+                   "And there they ar- WAIT A MINUTE WAIT A MINUTE RANDY ORTON WITH THE RKO!",
+                   "OFF THE TOP ROPE, IT'S OFF THE TOP ROPE, NO NO, OH MY LOR-",
+                   "HALFWAY ACROSS THE RING! THERE'S ABSOLUTLEY NO WAY! HE'S GOING FOR AIRBORNE LADIES AND GENTLEMEN! YOU GOTTA BE KIDDING ME! OH HOW DID HE DO IT! OH MY GOD! HE GOT HIM IN AN RKO IN MIDAIR!"]
```
danldb: ### Including presentation strings in business logic layer

lib/game.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def play(player_choice)
+    @comp = self.computer_play
+      if player_choice == @comp
+        :draw
+      elsif player_win?(player_choice)
+        :win
+      elsif player_lose?(player_choice)
+        :lose
+      end
```
danldb: ### Use of `if/elsif` conditionals for business logic
Long `if` and `elsif` trees are very difficult to read and nested `if` statements require too much working memory for a reader to quickly scan.

There are a number of approaches to the game logic of Rock Paper Scissors,  e.g.:

- Use a hash to map the rules:
```
rules = { rock: scissors,
          paper: :rock,
          scissors: :paper }
```
or for RPSLS:
```
rules = { rock: [scissors, lizard],
          paper: [:rock, :spock],
          scissors: [:paper, :lizard],
          lizard: [:paper, :spock],
          spock: [:rock, :scissors] }
```
- Use individual classes for each weapon (i.e. `Rock`, `Paper` etc.) with a `beats?` or similar method that takes another weapon as a parameter.

lib/game.rb
```diff
@@ -0,0 +1,52 @@
+class Game
+
+
+
+  def computer_play
+    [:rock,:paper,:scissors].sample
+  end
```
danldb: ### Not encapsulating the 'computer' in a separate class

lib/game.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def play(player_choice)
+    @comp = self.computer_play
```
danldb: you don't need this. instead of calling `@comp` every time, why not just use `computer_play`?

lib/game.rb
```diff
@@ -0,0 +1,56 @@
...
+        "You and your opponent both chose #{player_choice}!"
+      elsif player_win?(player_choice)
+        "You Win!"
+      elsif player_lose?(player_choice)
+        "You Lose!"
+      end
+  end
+
+  def player_win?(choice)
+    true if choice == :rock && @comp == :scissors || choice == :paper && @comp == :rock || choice == :scissors && @comp == :paper
```
danldb: This logic is very hard to understand as it is so heavily nested. Look at ways to simplify this code.

spec/game_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  describe '#computer_player' do
+    it 'computer player can return a random choice' do
+      allow(subject).to receive(:computer_play).and_return(:rock)
+      expect(subject.computer_play).to eq(:rock)
+    end
+  end
+
+  describe '#player' do
+    it 'player can win' do
+      subject.play(:choice)
```
danldb: `:choice` is just a simple. Did you mean to use the double `choice`?

views/gamepage.erb
```diff
@@ -0,0 +1,66 @@
+<body background="http://i.imgur.com/mtW4Th6.jpg" height="1000px" width="1600px">
+<div style="text-align: center; color:white; font-family:Arial">
+
+  <% if @visitor=="" %>
```
danldb: could use `if @visitor.empty?` as it is slightly more semantic.

views/gamepage.erb
```diff
@@ -0,0 +1,66 @@
...
+    Submit your name before you enter the ring!
+    <form action="/" style="padding: 20px">
+      <input type="submit" value="Go Back">
+    </form>
+  <% elsif @visitor %>
```
danldb: If there are only 2 possiblities, use `if` `else` instead of `if` `elsif`.

views/gamepage.erb
```diff
@@ -0,0 +1,66 @@
...
+    IT'S A DRAW! You and your opponent both chose <%= @choice_1 %>!
+      <form style="padding: 20px">
+        <input type="button" value="Replay?" onClick="history.go(-1);return true;">
+      </form>
+
+  <% end %>
+
+
+</div>
+</body>
```
danldb: You went beyond the brief by adding music. Having said that it looked fun, so will not mark this against you - but be aware that if this was a real tech test or client brief, going beyond the brief is very bad news and could lose you a job. Keep it simple.

danldb:
:construction: 

Good attempt, but no sticker this week. Your CI is failing as you did not correctly set up your test environment, therefore your code is currently untested. Have a look through https://github.com/makersacademy/rps-challenge/wiki/Common-issues as many of these apply to you.

PR #172, opened 2015-09-21
Working with roi

timhyson:


roidriscoll:
Hi Tim, as per our discussion earlier, do make sure you are happy with everything we went through and if you can refactor the game class that would be ace. :construction:

PR #173, opened 2015-09-21
Alex Person's rps-challenge

AlexPerson:


spec/features/starting_a_game_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'spec_helper'
```
danldb: Not initialising capybara correctly

Don't forget to add `Capybara.app = MyRackApp` or similar. You can use generators such as `rspec-sinatra init myApp lib/myapp.rb` but beware that the spec_helper will be overwritten, so you may want to save all the CI first.

spec/features/starting_a_game_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+		fill_in('player_name', :with => 'Alex')
+		click_button 'New Game'
+		expect(page).to have_content "Alex's game:"
+	end
+
+	scenario 'after a player inputs their name, a new game is created and a player is asked for their play' do
+		visit '/'
+		fill_in('player_name', :with => 'Alex')
+		click_button 'New Game'
+		page.check 'Rock'
```
danldb: Unable to find checkbox 'Rock'

spec/rps_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+	end
+
+	it 'only allows plays of Rock, Paper, or Scissors' do
+		expect{subject.play(:banana)}.to raise_error 'Please play Rock, Paper, or Scissors.'
+	end
+
+	it 'can tie' do
+		subject.play(:rock)
+		# subject.computer_choice = :rock
+		expect(subject.result).to eq "It's a tie! Try again!"
```
danldb: this test fails

spec/rps_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+	end
+
+	it 'can lose' do
+		subject.play(:scissors)
+		expect(subject.result).to eq "Bummer! You lose! Try Again!"
+	end
```
danldb: this test also fails

lib/rpsclass.rb
```diff
@@ -0,0 +1,32 @@
+class Rps
```
danldb: Inconsistent file naming

Ruby class files should be named with the snake_case version of the class name. Class names should be PascalCase. Hence:

Good

class Rps -> rps.rb
class RpsWeb -> rps_web.rb
Bad

class RPS_web -> rps_web.rb
class RpsWeb -> rps.rb
Note, naming conventions tend to prefer acronyms to be 'wordified' i.e. RPS becomes Rps or rps as appropriate.

lib/rpsclass.rb
```diff
@@ -0,0 +1,32 @@
...
+		@defeat = {scissors: :rock, rock: :paper, paper: :scissors}
+		@player_choice = ""
+	end
+
+	def play shape
+		@player_choice = shape.downcase.to_sym
+		if !@valid_plays.include?(player_choice)
+			raise 'Please play Rock, Paper, or Scissors.'
+		end
+	end
```
danldb: the play method is not really doing anything other than raising an error if the choice is invalid. Did you mean to set the result? Your tests are failing due to this.

lib/rpsclass.rb
```diff
@@ -0,0 +1,32 @@
+class Rps
+	
+	attr_reader :defeat, :player_choice, :result, :computer_choice
+	# attr_accessor :player_choice
+
+	def initialize
+		@valid_plays = [:rock, :paper, :scissors]
+		@defeat = {scissors: :rock, rock: :paper, paper: :scissors}
```
danldb: ### Not storing the weapons in a constant
If you have something like this:
```
def weapons
  ['Rock', 'Paper', 'Scissors']
end
```
Then 4 new objects will be created every time you call `weapons`.  Use a constant with symbols instead:
```
```

lib/rpsclass.rb
```diff
@@ -0,0 +1,32 @@
...
+	end
+
+	def game_eval computer_choice, player_choice
+		if player_choice == computer_choice
+			@result = "It's a tie! Try again!"
+		elsif computer_choice == @defeat[player_choice] 
+			@result = "Bummer! You lose! Try Again!"
+		else
+			@result = "Congrats! You're a winner!"
+		end
```
danldb: Where does game_eval get called? Your tests are checking for the result, but I can't see where game_eval is actually being executed.

lib/rps_web.rb
```diff
@@ -0,0 +1,39 @@
...
+  	@player_name = session[:player_name]
+  	$rps = Rps.new
+  	erb :game_setup
+  end
+
+  post '/game-result' do
+  	@player_shape = params[:player_shape]
+    player_shape = $rps.play(@player_shape)
+    computer_shape = $rps.computer_choice
+    $rps.game_eval(computer_shape, player_shape)
```
danldb: Just seen that you are calling game_eval here instead of in the logic - that is why your unit tests are failing because they never hit this layer. 

### Fat controllers

lib/rps_web.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  get '/game-setup' do
+  	@player_name = session[:player_name]
+  	$rps = Rps.new
+  	erb :game_setup
+  end
+
+  post '/game-result' do
+  	@player_shape = params[:player_shape]
```
danldb: ### POSTing to `/result`
Sending a `POST` to `/result` implies that you are _setting_ the result rather than submitting a go.  `/play` would be better.

danldb:
:construction: 

Fix tests, then have a look at https://github.com/makersacademy/rps-challenge/wiki/Common-issues as most of these apply to you.

PR #174, opened 2015-09-21
Game always ends in draw - need to refactor and implement capybara

MoeSadoon:


tansaku:
Hey @MoeSadoon some good work here - all test green and 100% coverage - excellent - some tips to help you improve

* Please do update your README following the [contribution notes](https://github.com/makersacademy/rps-challenge/blob/master/CONTRIBUTING.md), i.e.

  * Make sure you have written your own README that briefly explains your approach to solving the challenge.
  * If your code isn't finished it's not ideal but acceptable as long as you explain in your README where you got to and how you would plan to finish the challenge.

* The above is a relatively straightforward thing to do that doesn't involve much programming - I'll often get it done while thinking about other problems in the back of my mind :-)

* Also it's a great idea to show the full story of how your app is used (from a user's perspective) in the README, e.g. an irb transcript


* Furthermore when you have a web interface (and your one works nicely) you can show it off in the README:

interface screenshot


* note that tests like this:

```ruby
  it "is expected to pick a random shape" do
    expect(subject).to respond_to :pick_shape
  end
```
can be collapsed like this

```ruby
    it { is_expected.to respond_to(: pick_shape) }
```

This is particularly important in your code above, since you have a mismatch between what the test says it is doing, and what it is actually doing.  Using the one liner removes that issue.  If you really want to test whether ComputerOpponent can pick a random behaviour you need to stub out the randomness like so:

```ruby
    it 'is expected to pick a random shape' do
      allow_any_instance_of(Array).to receive(:sample).and_return(:scissors)
      expect(subject.computer_choice).to eq('scissors')
    end
```
Note that the respond_to tests are an initial step you go through using the tests to drive the creation of the objects public interface, and can safely be deleted once you have more sophisticated tests that check both the interface methods and their responses (and/or associated changes in state).  Note that the above also shows how you can effectively stub out the randomness to more closely cover your Game class in tests.  Using `allow_any_instance_of(Array)` is a bit of a hack - probably better to set your array of choices as a constant and then control what it receives and returns, but the above at least shows the principle

* You've called your main RPS class `ComputerOpponent`, which makes sense given if you consider the web app is the game, and you are just having some code in lib that is being the alternative to the human opponent, however you have the majority of the rules for the RPS game in the `ComputerOpponent` class.  This all works, but we are breaking the Single Responsibility Principle (SRP).  Currently `ComputerOpponent` has the responsibility of generating a random selection for the computer choice, AND knowing how to compare that with another strategy (rock, paper, scissors) to see if that is the winner.  I would be tempted to split these two pieces of functionality into two classes/modules, i.e. one for the computer opponent, and the other for the rule set.  The hope is that such an arrangement will be more maintainable in future.  That you would be able to add many different kinds of opponents (that used different approaches) and that the game logic for RPS could be made independent of types of opponent or types of interface

* You don't seem to be checking for any of the draw conditions, e.g. paper vs paper

* You might want to move your server file into the lib directory, and remove the blank lines which make things look disorganised

* Look carefully at what [Reek](https://github.com/troessner/reek) (`gem install reek ; reek -U`) and [Rubocop](https://github.com/bbatsov/rubocop) (what Hound uses, `gem install rubocop; rubocop -DS`) have to say about your code:

>   [18, 34]:ComputerOpponent#outcome calls @shape == :paper 2 times (DuplicateMethodCall) [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
  [27, 32]:ComputerOpponent#outcome calls @shape == :rock 2 times (DuplicateMethodCall) [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
  [20, 25]:ComputerOpponent#outcome calls @shape == :scissors 2 times (DuplicateMethodCall) [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
  [16]:ComputerOpponent#outcome has approx 9 statements (TooManyStatements) [https://github.com/troessner/reek/blob/master/docs/Too-Many-Statements.md]
  [31, 24, 17]:ComputerOpponent#outcome is controlled by argument player_choice (ControlParameter) [https://github.com/troessner/reek/blob/master/docs/Control-Parameter.md]
  [6]:ComputerOpponent#pick_shape doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]

> game_server.rb:5:7: C: Style/ClassAndModuleCamelCase: Use CamelCase for classes and modules. (https://github.com/bbatsov/ruby-style-guide#camelcase-classes)
class Rps_challenge < Sinatra::Base
      ^^^^^^^^^^^^^
game_server.rb:41:23: C: Style/SpecialGlobalVars: Prefer $PROGRAM_NAME over $0. (https://github.com/bbatsov/ruby-style-guide#no-cryptic-perlisms)
  run! if app_file == $0
                      ^^
lib/computer_opponent.rb:11:3: C: Style/TrivialAccessors: Use attr_writer to define trivial writer methods. (https://github.com/bbatsov/ruby-style-guide#attr_family)
  def shape(computer_shape=pick_shape)
  ^^^
lib/computer_opponent.rb:16:3: C: Metrics/CyclomaticComplexity: Cyclomatic complexity for outcome is too high. [7/6]
  def outcome(player_choice)
  ^^^
lib/computer_opponent.rb:16:3: C: Metrics/MethodLength: Method has too many lines. [19/5] (https://github.com/bbatsov/ruby-style-guide#short-methods)
  def outcome(player_choice)
  ^^^
lib/computer_opponent.rb:16:3: C: Metrics/PerceivedComplexity: Perceived complexity for outcome is too high. [11/7]
  def outcome(player_choice)
  ^^^
lib/computer_opponent.rb:20:12: W: Lint/ElseLayout: Odd else layout detected. Did you mean to use elsif?
      else @shape == :scissors
           ^^^^^^^^^^^^^^^^^^^
lib/computer_opponent.rb:27:12: W: Lint/ElseLayout: Odd else layout detected. Did you mean to use elsif?
      else @shape == :rock
           ^^^^^^^^^^^^^^^
lib/computer_opponent.rb:34:12: W: Lint/ElseLayout: Odd else layout detected. Did you mean to use elsif?
      else @shape == :paper
           ^^^^^^^^^^^^^^^^

* The above static code analyzers should not be treated as gospel, but they give good hints (most of the time) about places that your code can be improved.  What's particularly noticeable here is that many of the comments are pointing to the overly long `ComputerOpponent#outcome` method, which is overly long and needs it's boolean logic cleaned up ...

* Your erb files look fine, but might be good to look at adding a layout.erb for some of the stuff you have in index.erb

Hope that's not too intimidating a list of suggestions :-) You've got a great base here - make some of the above changes and we can see about getting you a sticker.

:construction:

PR #179, opened 2015-10-29
Challenge review rubric

### LINKED ISSUE #178: Create challenge review rubric for RPS

silvabox:
closes #178

### DIFF
--- a/docs/review.md
+++ b/docs/review.md
@@ -86,13 +86,9 @@ Ruby class files should be named with the snake_case version of the class name. 
 - `class RPS_web` -> `rps_web.rb`
 - `class RPSWeb` -> `rps.rb`
 
-## Not initializing capybara correctly
+## Not initializing capybara/ spec_helper correctly
 
-In `spec/spec_helper.rb`, don't forget to add `Capybara.app = MyRackApp` or similar. You can use generators such as `rspec-sinatra init myApp lib/myapp.rb` but beware that the spec_helper will be overwritten, so you may want to save all the CI first.
-
-## Overwriting spec_helper
-
-When using generators such as `rspec-sinatra` beware that spec_helper.rb will be rewritten. Make sure you make a copy of all the pre-written CI code, otherwise you will break your coveralls CI, causing silent failure of your pull request.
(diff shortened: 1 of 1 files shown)


docs/review.md
```diff
@@ -74,11 +117,11 @@ Although you do not need to test all possible combinations, your feature tests s
 
 to ensure the user interface logic is correct.
 
-### Feature tests can have more than one expect statement per it block (i)
+### Feature tests can have more than one expect statement per it block
 
 * feature tests can have more than one expect statement per it block
```
spike01: They can, but they probably still shouldn't.

docs/review.md
```diff
@@ -95,100 +138,345 @@ feature 'Playing the game' do
...
+## Defining weapons in more than one place
 
 Don't Repeat Yourself (DRY)!  The list of available weapons should be defined in only one place.  It can be passed around or referenced or injected, but not duplicated!
 
-### POSTing to `/result`
+Let's DRY the code from the encapsulation example above:
 
-Sending a `POST` to `/result` implies that you are _setting_ the result rather than submitting a go.  `/play` would be better.
+```ruby
+WEAPONS = [:rock, :paper, :scissors]
```
spike01: Where is this constant meant to live?

docs/review.md
tansaku: I think they should because the point of the feature test is that it checks multiple outcomes. Some discussion of this here:

https://github.com/andreareginato/betterspecs/issues/5

I like this comment from @myronmarston:

* In isolated unit specs, you want each example to specify one (and only one) behavior. Multiple expectations in the same example are a signal that you may be specifying multiple behaviors.
* In tests that are not isolated (e.g. ones that integrate with a DB, an external webservice, or end-to-end-tests), you take a massive performance hit to do the same setup over and over again, just to set a different expectation in each test. In these sorts of slower tests, I think it's fine to specify more than one isolated behavior.

from http://betterspecs.org/

docs/review.md
tansaku: well we've kind of spammed the global name space for simplicity here - bad us ... I guess we should prefer:

```ruby
class Game
end
```

PR #191, opened 2015-11-01
working initial model

rochester234:


tansaku:
You had your rps_challenge reviewed by **David Bebb**.
### The good points are:
* The README has instructions
* Win, lose and draw outcomes are all tested at the feature level
* No global variables used
* Files and classes are named consistently
* Capybara is initialized correctly
* There is no commented code
* The controllers are thin with minimal business logic

* Not testing all of the game logic in unit tests
* Testing too much game logic in feature tests
* Including presentation strings in business logic layer
* Not storing the weapons in a constant
* Defining weapons in more than one place
* Avoid calling business logic from the view

### Additional comments:
Need to separate model from view.

Code is easy to read, although the erb is very overweight (see above)


see https://github.com/makersacademy/rps_challenge/blob/master/docs/review.md for more details

rochester234:
app.rb line 16
player class & model class - method choose

app.rb
```diff
@@ -0,0 +1,27 @@
...
+    redirect '/play'
+    end
+
+    get '/play' do
+      @player = Player.new(session[:player_name])
+      @computer = Computer.new
+      @player_selection = @player.choose(params[:selection])
+      @computer_selection = @computer.choose
+      @game = Game.new
+      @result = @game.battle(@player.weapon, @computer.weapon)
```
sjmog: You're right - this is a lot to be doing in the controller. As an additional bonus horror, you're passing six separate instance variables to the view (Our Lord Sandi says we must aim to pass only one).

I think we can eliminate a few of these straight away. For one, we don't access `@computer` directly in the view, so we can just make line 20 `@computer_selection = Computer.new.choose`. Same for `@game` - we can just use `@result = Game.new.battle(@Player.new.weapon, Computer.new.weapon)`.

Even better, though, is if we can wrap all this in either a 'God class' or a 'Service Object'. An example 'God class`:

```ruby
# in app.rb
get '/play' do
  @game = Game.new(session[:player_name], params[:selection], Computer, Player)
end
```
```ruby
# in game.rb
class Game
  def initialize(name, selection, computer_class, player_class)
    @player = player_class.new(name)
   # etc
  end
end
```

Of course, the real issue here is that I _think_ you've build the model before building the controller. As a result, you've got a bunch of unnecessary code: for example, you make the player `#choose` a weapon before you then access it again through `#weapon`. This makes sense in Ruby, because we might want to defer using the weapon we chose. In the controller, we want our choice and selection to essentially be synonymous. So we don't really need to summon up a player at all:

```ruby
# in app.rb
get '/play' do
  @game = Game.new(session[:player_name])
  @game.result = Attack.run(@game)
  erb :play
end
```

Then, following BDD, I'd design a model that would _allow me to do the thing I just wrote above_.

PR #249, opened 2016-02-05
RPS Challenge

vickymg:


silvabox:
Don't use @ variables in your tests:
```
    before do
      @new_game = Game.new(player)
    end
```
You already have `subject(:game) { described_class.new(player) }`

silvabox:
Don't try to persist complex objects (i.e. an instance of `Game`) in `session`.

PR #263, opened 2016-02-06
weekend challenge #3

eilw:


silvabox:
Avoid multiple instance variables in your controller routes:
```
  get '/outcome' do
    @game = session[:game]
    @player1_name = @game.player1.name
    @player2_name = @game.player2.name
    @player1_move = @game.player1.move
    @player2_move = @game.player2.move
    erb @game.outcome
  end
```
As a general rule, 2 at the most.  In this instance, you could just access `@game` from the view; or if you're worried about exposing too much to the view, just have `@player1` and `@player2`:
  get '/outcome' do
    game = session[:game]
    @player1 = game.player1
    @player2 = game.player2
    erb game.outcome
  end
```

PR #269, opened 2016-02-06
rps-challenge

valogopedina:
setting up and 1 user story implemented


README.md
```diff
@@ -1,94 +1,50 @@
...
-  Coveralls::SimpleCov::Formatter
-]
-Coveralls.wear! 
-```
 
-You can see your [test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) when you submit a pull request, and you can also get a summary locally by running:
 
-```
-$ coveralls report
-```
```
drjparry: You could add what programs you used to build it, Sinatra etc, perhaps your learning objectives and approach to solving the challenge.

app.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  post '/name' do
+    session[:player_name] = params[:player_name]
+    redirect '/game'
+  end
+
+  get '/game' do
+    unless params[:choice].nil?
+      computer = Weapons.new(['rock', 'paper', 'scissors'].sample)
```
drjparry: the logic here can be moved to he game file and just return the result.

app.rb
```diff
@@ -0,0 +1,31 @@
...
+    erb :index
+  end
+
+  post '/name' do
+    session[:player_name] = params[:player_name]
+    redirect '/game'
+  end
+
+  get '/game' do
+    unless params[:choice].nil?
```
drjparry: I don't think this guard clause is necessary as it will never be true because people have to click a choice to get here.

mechacoach:
You had your rps-challenge reviewed by **David**

### The good points are:

* The README has instructions
* Every possible outcome of the game logic is tested at the unit level
* The feature tests test a reasonable subset of game logic
* Win, lose and draw outcomes are all tested at the feature level
* No global variables used
* weapons are sensibly stored in a constant
* The code defines weapons in a single place
* files and classes are named consistently
* There is no commented code
* The code minimizes the amout of business logic in the view

* Including presentation strings in business logic layer
* Not initializing capybara/ spec_helper correctly
* Fat Controllers

### Additional comments:
All over pretty good but need to extract computer class. 

see https://github.com/makersacademy/rps-challenge/blob/master/docs/review.md for more details

PR #274, opened 2016-02-07
Lou's week 3 challenge

innlouvate:


silvabox:
Think about encapsulation here:
```
  def initialize(session)
    @player_1 = session["player_1"]
    @player_2 = session["player_2"]
    @no_players = session["no_players"]
    @turn = player_1
  end
```
Should the game have any notion of what a 'session' is - or does this belong solely in the controller?

PR #287, opened 2016-02-08
Initial Pull Request.

Jeremy-Barrass:
I appreciate I left this one a bit late, but hopefully this will be satisfactory.


mechacoach:
You had your rps-challenge reviewed by **Emma Beynon**

### The good points are:

* The README has instructions
* Presentation strings not in business layer
* No global variables used
* files and classes are named consistently
* There is no commented code
* The code minimizes the amout of business logic in the view
* The controllers are thin with minimal business logic

* Not testing all of the game logic in unit tests
* Testing too much game logic in feature tests
* Not testing all game outcomes in feature tests
* Not storing the weapons in a constant
* Defining weapons in more than one place
* Not initializing capybara/ spec_helper correctly

### Additional comments:
- Build out feature tests to check a few different scenarios produce the correct result
- Build out unit tests to test all outcomes for all combinations
- Change if/else statements to a hash of symbols to produce the correct outcome
- Extend to 2nd player

see https://github.com/makersacademy/rps-challenge/blob/master/docs/review.md for more details

silvabox:
Think about the naming of your `Winner` class - is the winner a _type_ within the domain, or is it a role or context of another type.  It's good to think about encapsulating the game logic in this way, but the naming is misleading and leads to code like `winner.player_wins?` which is semantically off.  Also, `player_wins?` is a boolean, so should have only true or false values.  You currently return `true`, `false` and `"A tie"`, which is mixing types and therefore likely to cause confusion.

PR #295, opened 2016-03-18
Misa: RPS challenge

MisaOgura:


silvabox:
Hi Misa, thanks for submitting your RPS Challenge.  Congratulations on managing to complete the extensions and get a multi-browser version deployed - that’s a great achievement!

But you now need to focus on your code quality.  You’ve found a way to solve the problem; but is it the best way?  Has it resulted in great code?  How would you feel about throwing it away and starting again with a new design?  (I’m not suggesting you do!  But it’s an interesting question to ask yourself).  If you can solve the problem _and_ have beautiful code, then you've cracked it!

PR #297, opened 2016-03-18
Initial commit and README

NickMountjoy:


silvabox:
Hi Nick, thanks for submitting your RPS challenge.  The structure of your routes looks fine, but the separate business logic implementation is missing; which is maybe what tripped you up.  I recommend coming back to it at the end of this week to re-evaluate.

PR #299, opened 2016-03-19
First Commit

peteburch223:


silvabox:
Hi Pete, thanks for submitting your RPS Challenge.  I can see from your code that you have a background in programming, but maybe not OOD?  Your testing is very thorough, which is good, but your solution is over-complicated and lacks encapsulation (the two things are closely related).  Give me a shout if you want to discuss it.

PR #301, opened 2016-03-19
RLN RPS challenge

rhiannonruth:


silvabox:
Hi Rhiannon, thanks for submitting your RPS challenge.  Well done - it's one of the best I've seen - you should be very pleased with it.

PR #302, opened 2016-03-19
RPS Challenge

MuriloDalRi:


silvabox:
Hi Murilo, thanks for submitting your RPS challenge.  I'm not sure you really need the `current_player` implementation in `Player`.  It's also unnecessary to have `@p1_weapon = @p1.weapon` when you can always just refer to `@p1`.  Avoid names like `p1`: `player1` is much more descriptive!

You may want to look at the indentation in your `.erb` files and Demeter violations such as `Game.current_game.p1.name`

PR #303, opened 2016-03-19
First Commit

SimonGlancy:


silvabox:
Hi Simon, thanks for submitting your RPS challenge.  Your code is a little difficult to read and is not ‘idiomatic Ruby’, for example:
```
@rounds[-1].winner != "Draw" ? @score[@rounds[-1].winner.to_sym] += 1 : nil
```
and:
```
@score[@player1.name.to_sym] > BEST_OF/2 || @score[@player2.name.to_sym] > BEST_OF/2 ? true : false
```
are slightly cryptic.  Good Ruby should read almost like English and the intention of the code should be obvious without any cognitive load: always code for readability.
Remember also that `@outcome[0] == true` and `RULES[weapon1] == weapon2 ? true : false` are tautological: everything in Ruby is either ‘truthy' or ‘falsy’ and so `@outcome[0]` and `RULES[weapon1] == weapon2` are equivalent to the above, respectively.  Finally, predicates (methods ending in a `?`) are intended only to return ’truthy' or ‘falsy’ values: your `outcome_win?` method returns a name.

silvabox:
Sorry that's a bit critical, I meant it to be useful though :)

PR #308, opened 2016-03-19
RPS

hanfak:


silvabox:
Hi Han, thanks for submitting your RPS challenge.  Your solution is perhaps a little over-simple.  You only have one model class - `Computer` and I would generally hope to see roughly three along the lines of `Player`, `Computer` and `Game`.  Although that might be considered overkill for such a simple implementation, it’s partly an opportunity to practise your OO skills too.

PR #317, opened 2016-03-20
RPS Challenge Pull Request

eripheebs:


silvabox:
Hi Erika, thanks for submitting your RPS challenge.  You’ve done a reasonable job and added some style, which is great.  There are some issues with your code though.  Your multiplayer game doesn’t display the results correctly, so there is a logic error there somewhere.  Please also take a look at the following:
- The convention of prefixing describes with a `#` is when you are specifically describing an instance method.  e.g.  `describe ‘#attack’` - don’t use it elsewhere and avoid prefixing your specs with extraneous symbols and `should`s.
- Only suffix methods with a `?` if they are predicates (i.e. return `true` or `false`).  `last_move?` returns a string, so should be `last_move`.
- In `@name.dup.capitalize` you don’t need the `dup` as `capitalize` returns a new string anyway.  Arguably, `@last_move.dup` is overkill too although, technically, a string can be mutated, so it’s defensible.
- Your controller has too much logic, particularly in `’/attack’`.  It’s very difficult to read and all of that logic should be pushed down into the `Game`.
- Avoid logic in ternary branches.  For example:
```ruby
@game.single_player? ? (@player_2 = "Your opponent") : (@player_2 = @game.player_2.name)
```
Reads better as:
```ruby
@player_2 = @game.single_player? ? "Your opponent" : @game.player_2.name
```
Although it’s still logic that belongs in game.
- Refactor your code for readability:
```ruby
(player_2 == nil) ? (player_2 = nil) : (player_2 = player_klass.new player_2)
```
is quite tricky to interpret, whereas:
```ruby
player_2 = player_2_name ? player_klass.new(player_2_name) : nil
```
Finally, you have a lot of duplication here:
```ruby
  def player_wins?
    ((@player.last_move? == "Rock") && (@attack == "Scissors")) || ((@player.last_move? == "Paper") && (@attack == "Rock")) || ((@player.last_move? == "Scissors") && (@attack == "Paper"))
  end

  def player_loses?
    ((@player.last_move? == "Scissors") && (@attack == "Rock")) || ((@player.last_move? == "Rock") && (@attack == "Paper")) || ((@player.last_move? == "Paper") && (@attack == "Scissors"))
  end

  def player_draws?
    ((@player.last_move? == "Scissors") && (@attack == "Scissors")) || ((@player.last_move? == "Rock") && (@attack == "Rock")) || ((@player.last_move? == "Paper") && (@attack == "Paper"))
  end
```
Which could be easily addressed like so:
```ruby
  def player_wins?
    ((@player.last_move? == "Rock") && (@attack == "Scissors")) || ((@player.last_move? == "Paper") && (@attack == "Rock")) || ((@player.last_move? == "Scissors") && (@attack == "Paper"))
  end

  def player_loses?
    !(player_draws? || player_wins?)
  end

  def player_draws?
    @player.last_move? == @attack
  end
```

PR #320, opened 2016-03-20
Sunday night solution

chris-parker:


silvabox:
Hi Chris, thanks for submitting your RPS challenge.  Defining all of your methods as class methods (instead of instance methods) on `Game` is a bit of a code smell as it prevents you ever having more than one game.  Really, you want to be defining these methods on an instance of `Game`.  Also, think about SRP - should the player be responsible for defining the opponent's choice?

PR #324, opened 2016-03-21
finished work

Managram:


silvabox:
Hi, thanks for submitting your RPS challenge.  I can see you have put all of the business logic in `result_page.erb`, which violates separation of concerns (think MVC).  If this were a tech test, that would be a very black flag.  Also, think about `Randomiser` as class name.  It's not very domain-oriented.  It should be called just `Computer`; the fact that it selects randomly is internal implementaion.

PR #325, opened 2016-03-21
game - removed computer class

sachinkaria:


silvabox:
Hi Sachin, thanks for submitting your RPS challenge.  I’m concerned that you haven’t finished it and your solution uses global variables.  Did you make it through all of the materials last week?

sachinkaria:
hi ben,

yeah i finished this challenge and removed all the global variables.
My last commit & push was the 'removed global variables' one, is that the last one that came through?
I finished it on monday morning...

sachinkaria:
or even after that was updated readme

PR #331, opened 2016-04-29
add general structure

elibar-uk:


silvabox:
Your basic structure is OK but you haven't quite got a fully working game.  It's not far off though - there are just a few simple fixes required.  Firstly, look at `Player`.  You have `attr_reader :move` but then you define another `move` method in the class which will override the attribute.  You need one method to read the value and one to set the value.  Also note that you are not calling `select_winner` anywhere.

elibar-uk:
Have just pushed the working version

PR #332, opened 2016-04-29
rps-challenge

missamynicholson:


danldb:
### Use of if/elsif conditionals for business logic

Long if and elsif trees are very difficult to read and nested if statements require too much working memory for a reader to quickly scan.

PR #334, opened 2016-04-30
HWW RPS challenge

harrywynnwill:


silvabox:
While your game is reasonably well tested, you need to focus more on naming and encapsulation.  In a small program like RPS, the impact of naming and encapsulation issues are not significant, but as programs get larger this impact increases exponentially as it become impossible to keep the whole program in your head at once.  For example, to use `Game`, I need to know that I call `choice` (which is a noun) with a parameter to set the player's choice of weapon, and that this in turn sets a property called `player_choice` to the choice, and a property called `winner` to a string message that announces the winner or declares a tie.

Note also that you only need one singleton object (i.e. `Game`) to keep the reference to the game between requests; you don't also need to do this with `Player` since player is referenced by the game anyway.

PR #336, opened 2016-04-30
RPS Challenge

festinalent3:


danldb:

### Not encapsulating the 'computer' in a separate class

PR #337, opened 2016-04-30
Initial commit

sitypop:


silvabox:
Overall, a nice clean solution and I like that `Player` and `Computer` have the same interface.  You need to remove all style elements from your `erb` files though as it's almost impossible to play with a poor internet connection (the browser executes the HTML in the order it reads it, so starts trying to download the background image immediately, which blocks the rest of the page from rendering.)

PR #339, opened 2016-04-30
First user story and first push

nmrichards:


lib/game.rb
```diff
@@ -0,0 +1,30 @@
+require_relative 'player'
+class Game
+  attr_reader :player1, :player2, :result
+  def initialize(player1, player2 = Player.new("Computer"))
+    @player1 = player1
+    @player2 = player2
+    @combos = {"Scissors" => "Paper", "Paper" => "Rock", "Rock" => "Scissors"}
```
danldb: use symbols instead of strings

danldb:

### Use of `if/elsif` conditionals for business logic
Long `if` and `elsif` trees are very difficult to read and nested `if` statements require too much working memory for a reader to quickly scan. You made good use of a Hash to map the relationship between the weapons, but then still used a long conditional to evaluate that map.

Consider using individual objects for each weapon (i.e. `Rock`, `Paper` etc.) with a `beats?` or similar method that takes another weapon as a parameter.

PR #342, opened 2016-04-30
RPSLS challenge

lexiht:


danldb:

PR #344, opened 2016-05-01
Weekend challenge submission

bpstein:


lib/game.rb
```diff
@@ -0,0 +1,75 @@
...
+    elsif player_1.choice == 'Paper' && computer_choice == 'Scissors'
+      "Sorry, you lose :("
+    elsif player_1.choice == 'Rock' && computer_choice == 'Paper'
+      "Sorry, you lose :("
+    elsif player_1.choice == 'Scissors' && computer_choice == 'Paper'
+      "Sorry, you lose :("
+    elsif player_1.choice == computer_choice
+      "It's a tie!"
+    end
+  end
```
danldb: ### Use of `if/elsif` conditionals for business logic
Long `if` and `elsif` trees are very difficult to read and nested `if` statements require too much working memory for a reader to quickly scan.

There are a number of approaches to the game logic of Rock Paper Scissors,  e.g.:

- Use a hash to map the rules:
```
rules = { rock: :scissors,
          paper: :rock,
          scissors: :paper }
```
or for RPSLS:
```
rules = { rock: [scissors, lizard],
          paper: [:rock, :spock],
          scissors: [:paper, :lizard],
          lizard: [:paper, :spock],
          spock: [:rock, :scissors] }
```
- Use individual classes for each weapon (i.e. `Rock`, `Paper` etc.) with a `beats?` or similar method that takes another weapon as a parameter.

danldb:
:eyes:

PR #346, opened 2016-05-01
initial commit

chriselevate:


danldb:
Good use of constants to avoid magic numbers in your player class.

PR #347, opened 2016-05-01
RPS Week3 Challenge

tigretoncio:


lib/choices.rb
```diff
@@ -0,0 +1,33 @@
...
+    CHOICES[choice.to_sym]
+  end
+
+  def self.verb(winner,loser)
+    VERBS[(winner+loser).to_sym]
+  end
+
+
+
+
```
danldb: I'm not sure what advantage the class methods are bringing here. This class seems to only exist to hold the mapping of the different choices - a clue that maybe you haven't quite nailed down the job of the class - what is its purpose?

lib/game.rb
```diff
@@ -0,0 +1,58 @@
+class Game
+
+  attr_reader :player_1, :player_2, :player_mode, :game_mode
```
danldb: naming - player_mode and game_mode are not very expressive. What are these variables?

lib/help_view.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def self.player_mode(players)
+
+    players == "one" ? ONE_PLAYER : TWO_PLAYERS
+  end
+
+  def self.game_mode(mode)
+    GAME_MODE[mode.to_sym]
+  end
+end
```
danldb: Again, what does this actually do? What is it's purpose? Is it necessary?

danldb:
:eyes:

concentrate on the responsibilities of your classes - did you draw a domain diagram? 

some of your classes have no clear responsibility

think about your method naming - are you being specific enough; if the next coder is a psychopath, and he has your address, you want to be sure he will understand what your variables are.

lib/game.rb
tigretoncio: player mode: 1 or 2 players
game mode: "classic RPS" or Spock and Lizard.

What do you suggest as good variable names?

lib/help_view.rb
tigretoncio: it helps with my view in the following situations:
- When player is only one, adds computer to player 2 and makes it read only.  
- game_mode makes to display or not the spock an lizard in the relevant view

I would say that it is totally necessary.

I used them to avoid to make an if to the views, and duplicate view, using the "tell don't ask" approach.

What would be your suggestion for this?

lib/choices.rb
tigretoncio: I am never sure what is the relevant way of cutting responsibilities when defining classes, it seems I always take it the wrong way, either too little or too much.  I considered addding the choice to Player, but it looked like a big class.  Choice class, as the name indicates and as the readme explains, deals with what I call choices and you call weapons. it converts the choice to numeric to calculate the algorithm that determines result, and choses the relevant verb to indicate in the result, (ex. the Rock crushes the scissors).
So what is the relevant best practices, add all this functionality into Player or Game?

PR #348, opened 2016-05-01
Barebone first draft

letianw91:
User stories implemented.
Nothing else has been done.


lib/rulesplus.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'rules'
+
+class RulesPlus < Rules
```
danldb: The behaviour of the class is not changing, it still compares one object to another object. RulePlus doesn't have a reason to exist at all, and certainly not through inheritance.

danldb:
### Defining weapons in more than one place
Don't Repeat Yourself (DRY)!  The list of available weapons should be defined in only one place.  It can be passed around or referenced or injected, but not duplicated!

PR #352, opened 2016-05-01
RPS 

MaxProvin:


lib/game.rb
```diff
@@ -0,0 +1,69 @@
...
+class Game
+
+  R = "ROCK"
+  P = "PAPER"
+  S = "SCISSORS"
+  SP = "SPOCK"
+  L = "LIZARD"
+  WIN = [{p1:R, p2:S},{p1:S, p2:P},{p1:P, p2:R},{p1:L, p2:SP},{p1:SP, p2:S}]
+  WIN2 =[{ p1:R,p2: L},{p1:S, p2:L},{p1:P, p2:SP},{p1:L, p2:P},{p1:SP, p2:R}]
+  BEST_OF = 5.0
```
danldb: why abbreviations? your code is less communicative this way than if you just used the long form.

danldb:
You have gone beyond the brief by adding a scoring system, and it has made your code rather complicated - responsibilities are not clear. I would advise you to write the single responsibility of your class at the top in a comment - if you need to use an 'and' or 'or' then you probably have too much going on.

PR #353, opened 2016-05-01
Initial commit

HannaEb:


silvabox:
Overall, this is a reasonable solution, but there are some improvements you can make to your encapsulation, particularly in `Game`.  Having `@player_move` and `@computer_move` instance variables is a sign of a design problem; these should probably be `@player.move` and `@computer.move` or similar.  This would also then remove the need for `instance_variable_set` in your tests (because you could stub the computer double).  You don't need the singleton pattern on `Computer`, since you are storing an instance of this in `Game` anyway.

PR #355, opened 2016-05-01
Ruta's RPS challenge

Melodija:


silvabox:
It's a slightly unusual approach to have `rock`, `paper` and `scissors` methods in your `Player` class, rather than using a parameterized method, e.g. `player.choose(:rock)`.  While it could be argued that your approach is better encapsulated, you subsequently query a string property to determine the player's selection.  To really benefit from encapsulation, you should hide this implementation detail behind additional `rock?`, `paper?` and `scissors?` predicates.

Your `Selected` class has too many responsibilities and should not be tracking scores.  An instance variable called `@player_score` is a strong clue that this should be `@player.score`.

PR #357, opened 2016-05-01
Rock, Paper, Scissors

pixelandpage:


silvabox:
Your solution is very thoroughly tested, which is excellent.  Well done for implementing the more complicated Lizard Spock expansion and a scoring system.

However, you need to think about encapsulation.  The fact that there are seven `attr_reader`s in your `Game` and an `attr_accessor` in your `Player` is a sure sign that something is wrong.  Another clue is `@player_move` in `Game`.  Anything prefixed `@player_` should probably be in player.  i.e. `@player.move`.  Can you make `Player` responsible for increasing and resetting their score and get rid of the writable attribute?

Just one other thing to look out for:  set a fallback background colour for your UI in case the image doesn't load.  Playing on the train this morning, I couldn't read any of the text because it was the same colour as the background!  😉

PR #358, opened 2016-05-01
Pull request rps-challenge

Wil0:
App does not show result


silvabox:
Overall, your structure is good, but there is a fundamental logic problem in your game which means it can't work.  Can you see what it is?

PR #359, opened 2016-05-02
Functional gameplay

macebake:


silvabox:
You've done well to achieve a fully playable game.  Your next job is to think about how you can improve your code.  Here are some suggestions:

Your `Game` class is confused about its relationship with `Player` and `Computer`.  You can pass these in when the game is created, which is good.  You then set their choices from `app.rb`:
```ruby
post '/choice' do
  session[:player_choice] = @game.player.choose(params[:choice])
  session[:computer_choice] = @game.computer.choose_at_random
  @game.add_selection(@computer_choice, @player_choice)
  redirect '/result'
end
```
Note how the call to `@game.add_selection` is redundant here as both `@computer_choice` and `@player_choice` are `nil`.  Then in `/result` you call `@game.add_selection` again:
```ruby
@computer_choice = session[:computer_choice]
@player_choice = session[:player_choice]
@game.add_selection(@computer_choice, @player_choice)
session[:winner] = @game.winner
erb(:result)
```
This is unnecessary since the selections are already stored inside the `player` and `computer` classes inside your game.  You have no need to add the winner to the session either.

This means you don't need `@choices` in `Game` as it's just duplicating information.  You can calculate the winner on the fly (and also get rid of `@winner` and `@winner_value`):
```ruby
def winner_calc
  array = [player, computer].map { |val| MOVES.find_index(val) }
  (array.first - array.last) % 9
end

def winner
  winner_value = winner_calc
  if winner_value == 0
    "tie"
  elsif winner_value.even?
    winner = computer
  else
    winner = player
  end
end
```

Finally don't use `<br>` tags for layout and avoid CSS in your HTML!

PR #360, opened 2016-05-02
User stories satisfied - tests passing

RichardWatkins1:


danldb:
:eyes:

Your Game class is doing too much, as is shown by the number of attributes required in the initial hash. Should game have a name attribute?

PR #362, opened 2016-05-02
submitting files rps

hnamitha1:


lib/computer.rb
```diff
@@ -0,0 +1,11 @@
...
+	def choice
+		play
+	end
+
+  private
+	
+	def play
+		["rock","paper","scissors"].sample
+	end
+end
```
danldb: Please fix your indentation @hnamitha1 let me know if you are unsure how to do this - it could make the difference between getting an interview or not getting an interview.

danldb:
:eyes:

lib/computer.rb
hnamitha1: @danldb if I open in sublime it looks cleaner, in github it looks like this, Please suggest me how I can fix this, I am using tabspace 2 in sublime.

PR #483, opened 2016-10-14
First commmit

alexanderwjrussell:


app.rb
```diff
@@ -0,0 +1,55 @@
...
+  before { @game = Game.instance }
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/names' do
+    player_1 = params[:player_1_name]
+    player_2 = params[:player_2_name]
+    Game.create(Player.new(player_1),Player.new(player_2 = :The_Computer))
```
silvabox: This line is too cryptic - what is the intention of `Player.new(player_2 = :The_Computer)`?

lib/game.rb
```diff
@@ -0,0 +1,101 @@
+require_relative "player"
+
+class Game
+
+  attr_reader :played_games, :player_1_score, :player_2_score
```
silvabox: everything in this file prefixed `player_n_` should be in the `Player` class

spec/features/rock_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'spec_helper'
+
+feature "Paper" do
```
silvabox: You don't really need three separate feature tests here; they are identical other than a parameter so you could combine them:
```ruby
  %w[ROCK PAPER SCISSORS].each do |weapon|
    scenario "Expect the player to be able to select #{weapon}" do
      sign_in_and_play
      click_button(weapon)
      expect(page).to have_content("Games played: 1")
  end
end
```

app.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  get '/play' do
+    erb(:play)
+  end
+
+  get '/game' do
+    erb(:game)
+  end
+
+  get '/rock' do
```
silvabox: This should be `post` as your intention is to alter state.  You'd be better off with `post '/play'` passing the weapon as a form parameter, that way you would also only need one form in the HTML.

PR #485, opened 2016-10-14
Glenn B's submission for Rock, Paper, Scissors - challenge

glynester:


app.rb
```diff
@@ -0,0 +1,47 @@
...
+require './lib/game'
+#require './lib/player_attack'
+
+class RPS < Sinatra::Base
+
+  get '/' do
+    erb(:index)
+  end
+  
+  post '/registered' do
```
silvabox: shouldn't this be `post '/register'`?

app.rb
```diff
@@ -0,0 +1,47 @@
...
+    @rps_game = Game.game_instance   
+  end
+  
+  get '/play' do
+    erb(:play)
+  end
+  
+  post '/result' do
+    p params                                    # Remove this later!!!!!!!!!!!!!!!!!
+    @rps_game.selection(params['rps_choice'])
```
silvabox: too much logic in the controller

lib/game.rb
```diff
@@ -0,0 +1,78 @@
+require_relative 'player'
+
+
+class Game
+
+attr_reader :player_1, :player_2, :current_turn, :player_1_choice, :player_2_choice, :auto_choice, :winner
+
+	def initialize(player1, player2)
```
silvabox: what's happened to your indentation?  You must fix this.

lib/game.rb
```diff
@@ -0,0 +1,78 @@
...
+	
+	def self.game_instance
+	  @rps_game
+	end
+	
+	def switch_turns       # Called in the controller. Ben, do I call this from in my class or in the controller?	
+		@current_turn = opponent_of(current_turn)	
+	end
+	
+	def selection(choice)
```
silvabox: `selection` is a noun - so this should not change the state of the object.  This method should be called `select`, which is a verb.  You do not need `record_player_choices`.  Any method or variable prefixed `player_n_` is in the wrong place and should be in the `Player` class.  Then all this code would reduce to `@current_turn.choose(choice)` (or, preferably `@current_player.choose(choice)`.

PR #488, opened 2016-10-15
Submission for RPS from Frances Maxwell

francesmx:


app.rb
```diff
@@ -0,0 +1,27 @@
...
+    erb(:index)
+  end
+
+  post '/play' do
+    player = Player.new(params[:name])
+    @game = Game.create(player)
+    erb(:play)
+  end
+
+  post '/result' do
```
silvabox: Are you sending the result to the server?  The route should represent the thing you are posting to, e.g. `post '/choice'`.  I'd be inclined to call _this_ route `play` and rename the other one `start` or `register` or similar.

lib/player.rb
```diff
@@ -0,0 +1,13 @@
+class Player
+
+  attr_reader :name, :choice
+
+  def initialize(name = "Mystery Person")
+    @name = name
+  end
+
+  def set_choice(choice)
```
silvabox: avoid `set_` type methods.  You could call this `choose`.

lib/computer.rb
```diff
@@ -0,0 +1,14 @@
+class Computer
+
+  attr_reader :name, :choice
+
+  def initialize(name = "The Computer")
+    @name = name
+    @choice = random_choice
+  end
+
+  def random_choice
```
silvabox: this should be private

PR #489, opened 2016-10-15
rps CI

crsanghani:


app.rb
```diff
@@ -0,0 +1,35 @@
+require 'sinatra/base'
+require 'haml'
+require 'rspec'
```
silvabox: you shouldn't ever be requiring `rspec` in your app.

lib/player.rb
```diff
@@ -0,0 +1,8 @@
+class Player
```
silvabox: where is this class used?

lib/turn.rb
```diff
@@ -0,0 +1,36 @@
...
+    paper:    {rock: :win, paper: :draw, scissors: :lose, lizard: :lose, spock: :win},
+    scissors: {rock: :lose, paper: :win, scissors: :draw, lizard: :win, spock: :lose},
+    lizard:   {rock: :lose, paper: :win, scissors: :lose, lizard: :draw, spock: :win},
+    spock:    {rock: :win, paper: :lose, scissors: :win, lizard: :lose, spock: :draw}
+  }
+  
+  attr_reader :player_name, :player_choice, :opponent_choice
+  
+  def initialize(options)
+    @player_name = options["player_name"]
```
silvabox: You have a player class, which should be responsible for its name and choice.

lib/turn.rb
```diff
@@ -0,0 +1,36 @@
...
+    rock:     {rock: :draw, paper: :lose, scissors: :win, lizard: :win, spock: :lose},
+    paper:    {rock: :win, paper: :draw, scissors: :lose, lizard: :lose, spock: :win},
+    scissors: {rock: :lose, paper: :win, scissors: :draw, lizard: :win, spock: :lose},
+    lizard:   {rock: :lose, paper: :win, scissors: :lose, lizard: :draw, spock: :win},
+    spock:    {rock: :win, paper: :lose, scissors: :win, lizard: :lose, spock: :draw}
+  }
+  
+  attr_reader :player_name, :player_choice, :opponent_choice
+  
+  def initialize(options)
```
silvabox: passing `options` here is too woolly.  Firstly, it forces the caller to know arbitrary key names.  Secondly, you have player and opponent objects which should encapsulate names and choices.

PR #492, opened 2016-10-15
Do you want to play a game?

darrenf54321:


app.rb
```diff
@@ -0,0 +1,39 @@
...
+      redirect '/play'
+    end
+
+    get '/play' do
+      erb :play
+    end
+
+    post '/weapon_choice' do
+      @game.player_1.weapon_choice(params[:weapon_choice])
+      @game.player_2.weapon_choice(Weapon::WEAPON.sample)
```
silvabox: This shouldn't be the controller's responsibility.  If player_2 was a different object (e.g. a Computer instead of a Player), it could handle its own random selection.

lib/game.rb
```diff
@@ -0,0 +1,30 @@
...
+    @game = Game.new(player_1, player_2)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def result
+    return "draw" if player_1.weapon == player_2.weapon
+      if player_1.weapon == "Rock" && player_2.weapon == "Scissors"
```
silvabox: avoid long `if` statements if you can; there's usually a better way.  Since you've defined a `Weapon` class, this would be the perfect place to move this logic to.

lib/player.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def initialize(name)
+    @name = name
+  end
+
+  def weapon_choice
+    @weapon = Weapon.new
+  end
+
+  def weapon_choice(weapon)
```
silvabox: you can't overload methods in Ruby, so this method just overwrites the one above - you can't have both!

lib/weapon.rb
```diff
@@ -0,0 +1,5 @@
+class Weapon
```
silvabox: This class has no behaviour.

PR #493, opened 2016-10-15
RPS game time

crispinandrews:


app.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  get '/play' do
+    erb(:play)
+  end
+
+#too much logic for controller??
+post '/choice' do
+  @game.player_1.player_choice(params[:choice])
+  @game.player_2.player_choice(@game.auto_turn)
```
silvabox: Yes, too much logic in the controller.  If player_2 was a different object (e.g. a Computer instead of a Player) it could handle its own random selection

lib/game.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def auto_turn
+    ['Rock', 'Paper', 'Scissors'].sample
+  end
+
+#For sure find a better way of doing this!
+  def result
+    if player_1.choice == 'Rock' && player_2.choice == 'Scissors'
+      return "#{player_1.name} wins!"
```
silvabox: avoid returning presentation strings in the model.  Return something machine readable then convert this in the view layer.

lib/player.rb
```diff
@@ -0,0 +1,15 @@
...
+
+class Player
+
+  attr_reader :name, :choice
+
+  def initialize(name)
+    @name = name
+  end
+
+  def player_choice(choice)
```
silvabox: This method name is confusing, it means you have `player.choice` and `player.player_choice`.  If it's a command, it should be verbular, i.e.  `choose`.

PR #494, opened 2016-10-15
RPS Weekend Challenge

andygjenkins:


app.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  post '/name' do
+    player = Player.new(params[:player_name])
+    @game = Game.new_game(player)
+    redirect '/play'
+  end
+
+  post '/move' do
+    @game.player.move(params[:move])
+    @game.computer_move
```
silvabox: although it's only one line of code, I would argue this logic does not belong in the controller.   The game should be responsible for knowing if and when a computer move is required.

PR #497, opened 2016-10-15
A place to practice you RPS skills

amaalali:
Web app kickback and play some RPS and blow off some steam or hone your awesome RPS skills for competition.

Ready? Rock .. Paper .. Scissors .. GO!


lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def player_1_weapon
+    @player_1.weapon
+  end
+
+  def result(p2)
+    if player_1_weapon == p2.weapon
+      'Draw'
+    elsif player_1_weapon == 'Rock' && p2.weapon == 'Scissors'
+      "Rock beats Scissors. <b>#{player_1_name} wins!</b>"
```
silvabox: Don't return presentation strings from your models.  Return an object or code and translate this in the view layer.

lib/player.rb
```diff
@@ -0,0 +1,21 @@
...
+  def self.instance
+    @player
+  end
+
+  def initialize(player_name)
+    @name = player_name
+    @weapon = nil
+  end
+
+  def weapon_choice(weapon)
```
silvabox: This is a command not a query, so it should be verbular - i.e. `choose`.

app.rb
```diff
@@ -0,0 +1,47 @@
...
+    @player_1 = Player.create(params[:player_1_name])
+    @game = Game.create_single(@player_1)
+    redirect '/main-game'
+  end
+
+  get '/main-game' do
+    erb(:main_game)
+  end
+
+  post '/referee' do
```
silvabox: name your routes more in line with the domain and the messages you are sending.  'Referee' does not exist in the domain language and the message you are sending is the user's selection.  I would be inclined to define this as `post '/play'`

PR #501, opened 2016-10-16
Malinna's Marketeer Stress Relief Programme

MalinnaLeach:


app.rb
```diff
@@ -0,0 +1,37 @@
+require 'sinatra/base'
+require_relative './lib/game'
+
+class RPS < Sinatra::Base
+
+  get '/' do
+    erb :register
+  end
+
+  post '/names' do
```
maryrosecook: Maybe change this URL to reflect the fact that only one player name is set?

app.rb
```diff
@@ -0,0 +1,37 @@
...
+  before do
+    @game = Game.instance
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post '/choice' do
+    @game.player1.set_choice(params[:choice].to_sym)
```
maryrosecook: This feels a bit reachy-inny.  Have a read about the law of Demeter.  Could this be delegated to game?

app.rb
```diff
@@ -0,0 +1,37 @@
...
+    @game.computer.random_choice
+    @game.set_choices
```
maryrosecook: It feels like one method could be called that tells the game to calculate the result, and that method could hide calling all these methods internally.

app.rb
```diff
@@ -0,0 +1,37 @@
...
+    @game.decide_winner
+    redirect to '/result'
+  end
+
+  get '/result' do
+    erb @game.result
```
maryrosecook: Love this!

lib/computer.rb
```diff
@@ -0,0 +1,13 @@
+class Computer
+
+  attr_reader :choice
+
+  def initialize
+    @choice = nil
```
maryrosecook: Could the choice be decided on initialisation, to avoid the possibility of a Computer instance being in an invalid state?

lib/game.rb
```diff
@@ -0,0 +1,49 @@
...
+  CHOICES = [:rock, :paper, :scissors]
+
+  attr_reader :player1, :computer, :result, :p_choice, :c_choice
+
+  def initialize
+    @player1 = nil
+    @computer = nil
+    @result = nil
+    @rules = {rock: :scissors, paper: :rock, scissors: :paper}
+    @p_choice = nil
```
maryrosecook: Could `p_choice`, `c_choice` and `result` all be methods so that they are computed on the fly? This would cut down on the number of instance variables, which would make the class easier to understand.

PR #502, opened 2016-10-16
RPS Challenge Pull Request

B-raw:


app.rb
```diff
@@ -0,0 +1,40 @@
...
+  before do
+    @game = Game.instance
+  end
+
+  get '/play' do
+    erb(:play)
+  end
+
+  post '/result' do
+    @game = Game.instance
```
silvabox: isn't this already handled by the `before` block?

lib/computer.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  attr_reader :name, :weapon
+
+  def initialize
+    @choices = [:rock, :paper, :scissors]
+    @name = "Computer"
+    @weapon = set_weapon
+  end
+
+  def set_weapon
```
silvabox: this should be private

lib/game.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def game_draw
+    @player_1.weapon == @player_2.weapon
+  end
+
+  def player_1_beats_player_2?
+    winning_moves[@player_1.weapon] == @player_2.weapon
+  end
+
+  def winning_moves
```
silvabox: This should be a constant.  Every call to `winning_moves` creates a new hash, which is unnecessary.

lib/game.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def self.instance
+    @game
+  end
+
+  def decide_winner
+    if game_draw
+      @winner = nil
+    elsif player_1_beats_player_2?
+      player_1_wins
```
silvabox: this is unnecessarily circuitous - just assign `@winner = @player_1` here.

lib/game.rb
```diff
@@ -0,0 +1,56 @@
+require_relative 'computer'
+
+class Game
+
+  attr_reader :player_1, :player_2, :player_2_choice, :winner
```
silvabox: what is `player_2_choice` for?

lib/player.rb
```diff
@@ -0,0 +1,17 @@
+class Player
+
+  attr_reader :name, :weapon
+
+  def initialize(name)
+    @name = name
+  end
+
+  def set_weapon=(weapon)
```
silvabox: you are confusing setters and writers here.  In Ruby, we generally avoid methods like `set_weapon` in favour of methods called `weapon=`; here you've combined the two!

lib/player.rb
```diff
@@ -0,0 +1,17 @@
...
+    @weapon = weapon
+  end
+
+  def computer?
```
silvabox: Player should have no notion of what a computer is.

lib/computer.rb
```diff
@@ -0,0 +1,15 @@
+class Computer
+
+  attr_reader :name, :weapon
+
+  def initialize
+    @choices = [:rock, :paper, :scissors]
```
silvabox: this should be a constant to avoid recreating the array for every new instance.

PR #505, opened 2016-10-16
ROCK, PAPER, SCISSORS.... GO!

fmlharrison:


app.rb
```diff
@@ -0,0 +1,29 @@
...
+require_relative './lib/computer'
+
+class RockGame < Sinatra::Base
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    Game.start_game(Player.new(params[:player_name]), Computer.new)
```
maryrosecook: Love this! Great that you inject the player and computer objects, and that you wait to instantiate the game until they are ready so you reduce the likelihood of the game having incomplete data.

app.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  get '/result' do
+    @player_choice = Game.current_game.player.weapon_choice(params[:player_choice])
+    @computer_choice = Game.current_game.computer.weapon_choice
+    erb Game.current_game.result(@player_choice, @computer_choice)
```
maryrosecook: This feels like it's breaking encapsulation a bit.  The controller understands how to calculate the result (by passing in the player and computer choices). Can the game do this for itself?

lib/computer.rb
```diff
@@ -0,0 +1,10 @@
+class Computer
+
+  WEAPONS = [:rock, :paper, :scissors, :lizard, :spock]
+
+  attr_reader :weapon
+
+  def weapon_choice
+    @weapon = WEAPONS.sample
```
maryrosecook: Can this be calculated on initialisation, so the Computer is less likely to be in a state where it doesn't have all the data it needs?

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+
+  RULES = {
+    rock:     {rock: :draw, paper: :lose, scissors: :win, lizard: :win, spock: :lose},
+    paper:    {rock: :win, paper: :draw, scissors: :lose, lizard: :lose, spock: :win},
+    scissors: {rock: :lose, paper: :win, scissors: :draw, lizard: :lose, spock: :lose},
+    lizard:   {rock: :lose, paper: :win, scissors: :lose, lizard: :draw, spock: :win},
+    spock:    {rock: :win, paper: :lose, scissors: :win, lizard: :lose, spock: :draw}
+  }
```
maryrosecook: Love this data driving of the logic!

lib/player.rb
```diff
@@ -0,0 +1,13 @@
...
+  attr_reader :name
+  attr_accessor :weapon
+
+  def initialize(name)
+    @name = name
+    @weapon = nil
+  end
+
+  def weapon_choice(choice)
+    @weapon = choice.downcase.to_sym
```
maryrosecook: This is a subtle point, but it feels like the controller's responsibility to massage the choice into a symbol, not the model's.

PR #508, opened 2016-10-16
Rock, paper, scissors - RG

rgollancz:


app.rb
```diff
@@ -0,0 +1,41 @@
...
+    @game.check_result
+    redirect to('/result')
+  end
+
+  get '/result' do
+    @game = Game.instance
+    erb :result
+  end
+
+  post '/reset' do
```
maryrosecook: All these POSTs are fine.  In each one, you're setting data on the server.

views/result.erb
```diff
@@ -0,0 +1,18 @@
+<h1><%= @game.player_1_choice.capitalize %> Vs. <%= @game.choice.capitalize %></h1>
```
maryrosecook: Your instinct about the view having too much logic is part right.  It's not about the logic in this case (though logic in views is not good), but about knowledge and coherence.  Because the view code says `@game.player_1_choice`, it has knowledge of game that it doesn't need.  The view's job is to display things. The controller's job is communicating between the model and view.  So it's the controller's job to get the data out of the model and send it to the view. So, the controller could have `@player_1_choice = @game.player_1_choice` and the view could then just have `<%= @player_1_choice %>`.

views/result.erb
```diff
@@ -0,0 +1,18 @@
...
+<br>
+<% if @game.winner != '' %>
+The winner is...<%= @game.winner %>!
+<%end%>
+<% if @game.winner == '' %>
```
maryrosecook: This feels a bit close to logic.  Could game have a method called `tie?`

app.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  get '/play' do
+    @game = Game.instance
+    erb :play
+  end
+
+  post '/calculate' do
+    @game = Game.instance
+    @game.set_player_1_choice(params[:choice])
```
maryrosecook: Could the result automatically be calculated as soon as set_player_1_choice is called? This would better encapsulate the logic of game.

PR #510, opened 2016-10-16
rps challenge

ALucking1:


lib/game.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def self.instance
+    @game
+  end
+
+  def initialize(player_1)
+    @player_1 = player_1
+    @choices = [:rock, :paper, :scissors]
+    @option = nil
```
maryrosecook: Can you wait to initialize the game until you have a player choice and computer choice available? This would prevent a situation where you call `winning?` before the choices are available, which would produce an erroneous result.

lib/game.rb
```diff
@@ -0,0 +1,54 @@
...
+    #@result = result
+  end
+
+  def player_1_name
+    @player_1.name
```
maryrosecook: I like this delegation!

lib/game.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def player_1_choice(option)
+    raise "Oops, that doesn't exist! Please choose rock, paper or scissors." unless option == "rock" || option == "paper" || option == "scissors"
+    @option = option.to_sym
+  end
+
+  def computer_choice
+    @choices.sample
```
maryrosecook: Could the responsibility for understanding how to pick a computer move be delegated to a new class?

views/go.erb
```diff
@@ -0,0 +1,3 @@
+<h1>Player 1's <%=@game.option%> vs. Computer's <%=@game.computer_choice%></h1>
```
maryrosecook: Could option and computer choice be passed as variables to the view, rather than the view reaching into `@game` to grab them.  This reaching in means the view has more knowledge of the game object than it needs to, which means if you game object changes, the view is more likely to have to change as well.

maryrosecook:
Nice! Way to go with continuing to refactor!

PR #523, opened 2016-10-16
Weekend Challenge Attempt

LawrenceHunt:


app.rb
```diff
@@ -0,0 +1,54 @@
...
+    trump = Trump.new
+    @game = Game.create(player, trump)
+    redirect '/choose'
+  end
+
+  get '/choose' do
+    erb :choose
+  end
+
+  post '/outcome' do
```
silvabox: are you sending the server the outcome?  This should be `post '/play'` or something similar.

app.rb
```diff
@@ -0,0 +1,54 @@
...
+    @game.player.choose_weapon('rock') if params[:rock]
```
silvabox: this is clumsy.  Why not send `'rock'` as the value of a parameter (eg `params[:choice]`, then just assign the value directly?

lib/game.rb
```diff
@@ -0,0 +1,128 @@
+require_relative 'player'
+require_relative 'trump'
+
+class Game
+
+attr_reader :list_of_wins, :current_winner, :overall_winner, :player_score, :trump_score, :games_played, :player, :trump
+
+  def self.create(player, trump)
+    @game = Game.new(player, trump)
```
silvabox: the game should not need to know who is playing - just have two objects that behave as players.

lib/game.rb
```diff
@@ -0,0 +1,128 @@
...
+
+  def self.instance
+    @game
+  end
+
+  def initialize(player, trump)
+    @player = player
+    @trump = trump
+    @list_of_wins = []
+    @player_score = 0
```
silvabox: the player should know its own score.  Any variables or methods prefixed `player_` are clues that the logic is in the wrong place.

lib/game.rb
```diff
@@ -0,0 +1,128 @@
...
+  def trump_wins?
+    @current_winner == 'Trump'
+  end
+
+  def draw?
+    @current_winner == :draw
+  end
+
+
+  def calculate_overall_winner_private
```
silvabox: don't do this.

lib/game.rb
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  def update_list_of_wins
+    update_list_of_wins_private
+  end
+
+private
+
+  def player_wins?
+    @current_winner == "#{@player.name}"
```
silvabox: Why assign `@current_winner` as a string only to have to convert it back again.  Assign `@current_winner` the actual object, then this becomes `@current_winner == @player`

PR #524, opened 2016-10-16
Add complete web app RPS

ktsein:
need to learn how to stub and mock random features in Capybara feature text


spec/features/computer_play_spec.rb
```diff
@@ -0,0 +1,10 @@
+# feature 'Computery Play' do
```
maryrosecook: To mock the random computer moves, try using srand to set a seed for this test.

lib/computer.rb
```diff
@@ -0,0 +1,22 @@
+class Computer
+
+attr_reader :computer_item
+
+def self.create
+  @computer = Computer.new
```
maryrosecook: Try not to use this pattern pattern everywhere (using a class method to create an instance of the class and then store it on the class as a class instance variable).  Having all your classes be singletons reduces the object-orientedness of your code.  Perhaps remove the code in player and computer that does this, and instead pass instance of player and computer into the game class? (If you're not sure what I mean, come and have a quick chat!)

lib/computer.rb
```diff
@@ -0,0 +1,22 @@
...
+end
+
+def self.instance
+  @computer
+end
+
+def initialize
+  @computer_item
```
maryrosecook: No need to initialize instance variables.

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+  def self.instance
+    @game
+  end
+
+  def initialize
+    @result
+  end
+
+  def who_win(player, computer)
+    if player == computer
```
maryrosecook: Could you rewrite this to not use conditionals and instead use some data that defines which moves beat which moves. eg `{ rock: { beats: :scissors }, paper: { beats: :rock } etc.......... }`

PR #525, opened 2016-10-16
initial pull request

jinimcoroneo:


app.rb
```diff
@@ -0,0 +1,39 @@
...
+  get '/' do
+    erb(:index)
+  end
+
+  post '/name' do
+    session[:name] = params[:name]
+    redirect '/choose'
+  end
+
+  get '/choose' do
```
maryrosecook: Could this URL be more descriptive? eg so it indicates that a move is being chosen?

app.rb
```diff
@@ -0,0 +1,39 @@
...
+    @player = session[:name]
+    erb(:choose)
+  end
+
+  post '/choose' do
+    session[:choice] = params[:choice]
```
maryrosecook: It's cool that the code uses the session to store data.  However, it kind of loses some of the power of OO programming by doing this.  Variables from different objects are stored in the session, so we miss out on some of the the benefits of encapsulation.

lib/opponent.rb
```diff
@@ -0,0 +1,6 @@
+class Opponent
+
+  def choice
+    [:ROCK, :PAPER, :SCISSORS].sample
```
maryrosecook: Nice.

views/choose.erb
```diff
@@ -0,0 +1,78 @@
...
+   <tr>
+     <form action = "/choose" method = "post">
+      <td align = "center">
+  	     <button type = "submit" name = "choice" value = "ROCK">ROCK</button>
+  	  </td>
+      <td align = "center">
+  	     <button type = "submit" name = "choice" value = "PAPER">PAPER</button>
+  	  </td>
+      <td align = "center">
+  	     <button type = "submit" name = "choice" value = "SCISSORS">SCISSORS</button>
```
maryrosecook: Great - good abstraction.

views/play.erb
```diff
@@ -0,0 +1,151 @@
+<table align = "center">
+<tr>
+  <td></td>
+  <td align = "center" id = "outcome">
+    <% if @choice == "ROCK" %>
```
maryrosecook: This abstraction could be improved.  First, it would be an improvement to move this conditional code to the model.  The view's responsibility is to show things.  The model's responsibility is to calculate and store things.  Second, is there a way to calculate the result without having to have a special case if for every situation.  Tom's code is a good example of this: he used data to drive the logic.

PR #527, opened 2016-10-16
RPS weekend challenge

Swinston88:


lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+  attr_reader :player_1, :player_2
+
+  def initialize(player_1)
+    @player_1 = player_1
+    @player_2 = Player.new(generate_name)
+    @selections = []
+  end
+
+  def player_select(selection)
+    @selections << selection
```
silvabox: This is arbitrary.  Surely `player_select` is a clue that the player should be responsible for their selection?

You are forcing the user of your game class to know they must call `player_selection` before `add_computer_selection` (otherwise the array will be in the wrong order).  This completely breaks encapsulation.

lib/game.rb
```diff
@@ -0,0 +1,64 @@
...
+    @selections << computer_select
+  end
+
+  def choices
+    @selections.dup
+  end
+
+  def result
+    if win?
+      p "#{@player_1.name} wins!"
```
silvabox: Why are you using `p` here?  You should be returning the result from the method.  Also, these results are presentation strings, which is not the game's concern.

lib/player.rb
```diff
@@ -0,0 +1,8 @@
+class Player
```
silvabox: `Player` has no behaviour.  Unless the player is also managing the weapon selection, you might as well just use a string to represent the name, rather than a whole class.

PR #528, opened 2016-10-17
MSabatino - RPS - Challenge 

ManuCiao:


app.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  post '/paper' do
+    @game.paper
+    @game.computer_move
+    redirect '/move'
+  end
+
+  post '/scissors' do
+    @game.scissors
+    @game.computer_move
```
maryrosecook: It feels like the code for the `/rock`, `/paper` and `/scissors` routes are quite similar.  Could you abstract them so that one route can handle all three cases?

lib/game.rb
```diff
@@ -0,0 +1,60 @@
...
+
+class Game
+
+attr_reader :player, :pc_move, :outcome, :player_move
+
+  RULES = {
+    rock: {rock: :draw, paper: :lose, scissors: :win},
+    paper: {paper: :draw, scissors: :lose, rock: :win},
+    scissors: {scissors: :draw, rock: :lose, paper: :win}
+    }
```
maryrosecook: Love that you've used data to drive the logic!

lib/game.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def initialize(player)
+    @player = player
+  end
+
+  MOVE = [ :rock ,:paper ,:scissors ]
+
+  def computer_move
+    @pc_move = MOVE.sample
```
maryrosecook: It feels like this method is an extra responsibility.  Could it be moved to a different class?

lib/game.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def rock
```
maryrosecook: These `rock`, `paper` and `scissors` methods may not be necessary once the `/rock`, `/paper` and `/scissors` routes are refactored out.

lib/game.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def paper
+    @player_move = MOVE[1]
+  end
+
+  def scissors
+    @player_move = MOVE[2]
+  end
+
+  def draw?
```
maryrosecook: I like how these methods make the code more readable!

views/move.erb
```diff
@@ -0,0 +1,84 @@
+
+<div class="intro">
+  <h1>You chose <%= @game.player_move.to_s.capitalize %>!</h1><br>
```
maryrosecook: It feels like `@game.player_move` makes the view know too much about the model.  Can the controller pass some variables into the view? eg in the controller: `@player_move = @game.player_move`.

views/move.erb
```diff
@@ -0,0 +1,84 @@
...
+  </form>
+</div>
+
+<div>
+  <footer>Made by Manuela Sabatino</footer>
+</div>
+
+<style>
+
+  body {
```
maryrosecook: Is there a way to avoid repeating this CSS in different templates?

maryrosecook:
Nice! Way to go with doing more refactoring!

PR #530, opened 2016-10-17
Struggling to stub out random behaviour in computer_spec

stejgregory:
Logic layer completed. Need to work on controller next.


lib/game.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def self.instance
+    @game
+  end
+
+  def initialize(player, computer)
+    @player = player
+    @computer = computer
+    @player_score = 0
```
silvabox: `player_` is a clue that this should be in `Player`.  Same with `computer_`.

lib/player.rb
```diff
@@ -0,0 +1,17 @@
...
+class Player
+
+attr_reader :name, :weapon
+
+  def initialize(name)
+    @name = name
+    @weapon = nil
+  end
+
+  def weapon_select(weapon)
```
silvabox: `weapon_` is probably redundant here as the player doesn't do anything else, just `select` would be better.

lib/player.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'game'
+
+class Player
+
+attr_reader :name, :weapon
+
+  def initialize(name)
+    @name = name
+    @weapon = nil
```
silvabox: instance vars are `nil` implicitly - no need to initialize.

lib/computer.rb
```diff
@@ -0,0 +1,15 @@
...
+
+class Computer
+
+attr_reader :weapon
+
+  def initialize
+    @weapon = nil
+  end
+
+  def weapon_select
```
silvabox: Is this needed?  I shouldn't have to tell the computer to make its selection - it should just do it implicitly

PR #532, opened 2016-10-17
Play a simple Rock Paper Scissor Game

souljuse:
All tests pass and your name is on glitter

### DIFF
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -12,7 +12,7 @@ class Game
 
   def initialize(player)
     @player = player
-    @rbs = [:rock, :paper, :scissors]
+    @rps = [:rock, :paper, :scissors]
     @computer_choice
     @winner
   end
@@ -27,10 +27,8 @@ class Game
     end
   end
 
-  def rbs
-    @computer_choice = @rbs.sample
+  def rps
+    @computer_choice = @rps.sample
   end
-
-  private
 
 end
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -1,0 +1,36 @@
+class Game
+
+  def self.create(player_name)
+    @game = Game.new(player_name)
+  end
+
+  def self.instance
+    @game
+  end
+
+  attr_reader :winner, :player, :computer_choice
+
+  def initialize(player)
+    @player = player
+    @rbs = [:rock, :paper, :scissors]
+    @computer_choice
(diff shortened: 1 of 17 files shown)


lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def self.instance
+    @game
+  end
+
+  attr_reader :winner, :player, :computer_choice
+
+  def initialize(player)
+    @player = player
+    @rps = [:rock, :paper, :scissors]
```
silvabox: use a constant here to avoid recreating the array for every game.

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+    @computer_choice
```
silvabox: these two lines are redundant.  You do not need to initialize instance variables until you have a value to put in them (they will be `nil` by default).

rps.rb
```diff
@@ -0,0 +1,71 @@
...
+    redirect '/choose_your_weapon'
+  end
+
+  get '/choose_your_weapon' do
+    erb(:choose_your_weapon)
+  end
+
+  post '/rock' do
+    computer_chooses
+    @game.who_wins(:rock, @game.computer_choice)
```
silvabox: I shouldn't need to pass the computer choice here - the game should know how to handle that implicitly.

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+    @winner
+  end
+
+  def who_wins(user_input, computer_choice)
```
silvabox: this method name implies a question; the answer to which would be the player who won.  Instead it returns `true`, `false` or `nil` which doesn't answer the question.  Avoid assigning special significance to `nil` other than unassigned.

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+    if (user_input == computer_choice)
+      @winner = nil
+    elsif (user_input == :rock && computer_choice == :scissors) || (user_input == :paper && computer_choice == :rock) || (user_input == :scissors && computer_choice == :paper)
```
silvabox: this is messy and too difficult to read; avoid long conditionals like this.

silvabox: this is messy and too difficult to read; avoid long conditionals like this.

PR #533, opened 2016-10-17
Works but controller needs changing

James-SteelX:


lib/computer.rb
```diff
@@ -0,0 +1,22 @@
+class Computer
+
+  MOVES = [:rock, :paper, :scissors, :lizard, :spock]
+
+  attr_reader :move_choice, :name
+
+  def initialize(name = 'RPSLS-oid')
+   @move_choice = nil
```
silvabox: there's no need to initialize instance variables to `nil` - they are nil implicitly

lib/game.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  attr_reader :player, :winner, :computer
+
+  def initialize(player, computer = Computer.new)
+   @player = player
+   @computer = computer
+   @winner = nil
+  end
+
+  def result?
```
silvabox: methods with `?` suffix should only return `true` or `false` and not alter state.

lib/game.rb
```diff
@@ -0,0 +1,38 @@
...
+   @computer.computer_select
```
silvabox: Is this needed?  The computer should make its selection implicitly

lib/player.rb
```diff
@@ -0,0 +1,31 @@
+class Player
+
+ attr_reader :name, :player_choice
```
silvabox: `player_` is a redundant prefix for any methods in `Player`

spec/game_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+describe Game do
+
+  let(:comp2  ) { double(:computer, :name => :comp, :move_choice => :rock )}
+  let(:comp   ) { double(:computer, :name => :comp, :move_choice => :scissors )}
+  let(:player1) { double(:player, :name => :fred, :player_choice => :rock )}
+  let(:player2) { double(:player, :name => :fred, :player_choice => :paper )}
+  let(:player3) { double(:player, :name => :fred, :player_choice => :scissors )}
+  subject(:game) { described_class.new(:fred) }
+  subject(:draw) { described_class.new(player1, comp2) }
```
silvabox: defining multiple subjects like this is probably not defined in RSpec (although possible - it may have unexpected side effects).  This suggests you should be using `describe` or `context` blocks to separate out those tests that use different subject configurations.

lib/player.rb
```diff
@@ -0,0 +1,31 @@
...
+
+ def initialize(name)
+   @name = name
+   @player_choice = nil
+ end
+
+ def choose_rock
```
silvabox: although this creates a nice interface, it's actually not good design practice in the model layer.  It would be better to have one `choose` method and pass the choice as a parameter to that method.

app.rb
```diff
@@ -0,0 +1,65 @@
...
+    erb :index
+  end
+
+  post '/names' do
+   @player_1 = Player.new(params[:player_1_name])
+   @game = Game.new_game(@player_1)
+   redirect '/play'
+  end
+
+  before do
```
silvabox: Note, this will run before _every_ route, including those above, regardless of where you define it in this class.

PR #534, opened 2016-10-17
Rock Paper Scissors Challenge

james1968:
Please note the application is not complete.  Further work is required on unit and feature tests as well as issues computing a results once weapons are selected.


lib/game.rb
```diff
@@ -0,0 +1,55 @@
...
+  attr_reader :player, :player_selection, :computer_selection, :name
+
+  RULES = {
+    rock:     { rock: 'Draw', paper: 'Lose', scissors: 'Win' },
+    paper:    { rock: 'Win', paper: 'Draw', scissors: 'Lose' },
+    scissors: { rock: 'Lose', paper: 'Win', scissors: 'Draw'}
+  }
+
+  def initialize(name)
+    @computer_selection = nil
```
silvabox: you don't need to initialize instance variables to `nil`, they are nil implicitly

lib/game.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def loser?
+    result == 'Lose'
+  end
+
+  def draw?
+    result == 'Draw'
+  end
+
+  def computer_selection
```
silvabox: this method is confusing - it looks like a reader, but it's actually setting the value of `@computer_selection`.

`@computer_selection` is unnecessary anyway - the `Computer` instance should manage it's own selection

lib/game.rb
```diff
@@ -0,0 +1,55 @@
...
+    @computer_selection = @computer.selection
+  end
+
+  def player_selection(weapon)
```
silvabox: again, this method looks like a reader, but it's a mutator.  `@player_selection` is unnecessary as the player already stores it's selection.

lib/computer.rb
```diff
@@ -0,0 +1,8 @@
+class Computer
+  OPTIONS = [:rock, :paper, :scissors]
+
+  def selection
```
silvabox: this should be `weapon` so it is consistent with `Player`.  You can take advantage of duck typing later on then as the game can interact with either object without knowing if it's a `Player` or a `Computer`

spec/game_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  #let(:computer_selection) { double :paper }
+  subject { described_class.new(name)}
+
+
+  describe '#player_selection' do
+    it 'returns the players selection' do
+      expect(subject.player_selection(:scissors)).to eq(:scissors)
+    end
+  end
+  let(:computer_selection) { double :paper }
```
silvabox: what is this for?  It's not used anywhere

lib/game.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  attr_reader :player, :player_selection, :computer_selection, :name
+
+  RULES = {
+    rock:     { rock: 'Draw', paper: 'Lose', scissors: 'Win' },
+    paper:    { rock: 'Win', paper: 'Draw', scissors: 'Lose' },
+    scissors: { rock: 'Lose', paper: 'Win', scissors: 'Draw'}
+  }
+
+  def initialize(name)
```
silvabox: shouldn't this parameter be called `player`?

PR #535, opened 2016-10-17
Basic working model

DavidAMAnderson:


lib/game.rb
```diff
@@ -0,0 +1,47 @@
...
+   @player = player
+   @game_options = [:Rock, :Paper, :Scissors]
+   @winning_combos = [[:Scissors, :Paper], [:Paper, :Rock], [:Rock, :Scissors]]
+  end
+
+  def player_choice
+    @player_choice
+  end
+
+  def player_choice=(choice)
```
silvabox: lots of methods prefixed `player_` are a strong indication that this functionality belongs in `Player`

lib/game.rb
```diff
@@ -0,0 +1,47 @@
...
+  def computer_choice
+    @computer_choice = @game_options.sample
+  end
+
+  def player_wins?
+    fail "No winner" if draw?
+    @winning_combos.include? @outcome
+  end
+
+  def outcome
```
silvabox: this method should be private

lib/game.rb
```diff
@@ -0,0 +1,47 @@
...
+    @outcome = [@player_choice, @computer_choice]
```
silvabox: you don't need an instance variable here - just return the value from the method.

lib/game.rb
```diff
@@ -0,0 +1,47 @@
...
+    fail "Not possible option" unless @game_options.include? choice
+    @player_choice = choice.to_sym
+  end
+
+  def computer_choice
+    @computer_choice = @game_options.sample
+  end
+
+  def player_wins?
+    fail "No winner" if draw?
```
silvabox: Avoid raising exceptions in non exceptional circumstances.  This method should only return `true` or `false`.  If it's a draw, the player hasn't won, so it should be false.

lib/game.rb
```diff
@@ -0,0 +1,47 @@
...
+    @winning_combos.include? @outcome
+  end
+
+  def outcome
+    @outcome = [@player_choice, @computer_choice]
+  end
+
+  def draw?
+    outcome[0] == outcome[1]
```
silvabox: this logic is a bit cryptic

PR #574, opened 2017-01-21
Submission pull request

KatHicks:
Note that at the time of this pull request, the code is not yet finished!


KatHicks:
Started again (with backup of work) to try and resolve nokogiri bundle error which is stopping my tests from passing.

PR #583, opened 2017-01-21
KateLoschinina: Week3 Weekend challenge, RPS

kateloschinina:
Pull request contains: User story 1 and 2; Bonus Level 1 and 2


views/index.erb
```diff
@@ -0,0 +1,29 @@
...
+    <h1> Rock, Paper and Scissors </h1>
+
+    <h3> Welcome to the game!</h3>
+
+    <form action = "/game_choice" method = "post">
+      <label for = "how_many_players">
+        Do you want to play with computer or with some of your colleagues?</br>
+        <input type = "radio" name = "how_many_players" value = "one" id = "one_player" checked>Play with computer<br>
+        <input type = "radio" name = "how_many_players" value = "two" id = "two_players">Play with colleagues<br>
+      </label>
```
neoeno: I'd take another look at the `label` tag here, maybe compare it to some docs or references, maybe look at checkboxes together with label tags as there are a few styles specifically for that. It's supposed to be that you can click 'Play with computer' and the checkbox is activated, but here you have to click the checkbox itself.

lib/game.rb
```diff
@@ -0,0 +1,21 @@
...
+    @choice_one = choice_one ? choice_one.to_sym : :rock
+    @choice_two = choice_two ? choice_two.to_sym : :rock
+    @options = [:rock, :paper, :scissors, :spock, :lizard]
+  end
+
+  def won?
+    return :tie if @choice_one == @choice_two
+    return :won if RULES[@choice_one].include? @choice_two
+    :lose
+  end
```
neoeno: You mentioned the multiple `return`s — no issue with that to me. These are good idiomatic ruby "guard clauses". Some programmers object to them and prefer to have a single return per method, but it's generally accepted in ruby.

lib/game.rb
```diff
@@ -0,0 +1,21 @@
...
+  RULES = { rock: [:scissors, :lizard],
+          paper: [:rock, :spock],
+          scissors: [:paper, :lizard],
+          lizard: [:paper, :spock],
+          spock: [:rock, :scissors] }
+
+  def initialize(choice_one, choice_two)
+    @choice_one = choice_one ? choice_one.to_sym : :rock
+    @choice_two = choice_two ? choice_two.to_sym : :rock
+    @options = [:rock, :paper, :scissors, :spock, :lizard]
```
neoeno: Does this line do anything? Out of curiosity. If it does, it could be made a constant.

views/game_two_players_final.erb
```diff
@@ -0,0 +1,24 @@
...
+      <h3>The game is tied. We need to replay to break the tie.</h3>
+    <% end %>
+
+    <form action = "/replay_game_two_players" method = "post">
+      <input type = "submit" value = "Play again!">
+    </form>
+
+</body>
+
+<!-- templates that go into another templates -->
```
neoeno: Don't forget to remove commented out notes or code before you commit. If you forget though, don't worry — people will pick up on it in every code review you get for the rest of your life until you stop doing it! ha

app.rb
```diff
@@ -0,0 +1,91 @@
...
+    session[:choice_two] = params[:choice]
+    redirect '/game_two_players_final'
+  end
+
+  get '/game_two_players_final' do
+    @name_one = session[:name_one]
+    @name_two = session[:name_two]
+    @choice_one = session[:choice_one]
+    @choice_two = session[:choice_two]
+    @won = (Game.new(@choice_one, @choice_two)).won?
```
neoeno: These brackets make me think you've programmed before! Java, perhaps? Where you have to do `Boolean has_won = (new Game()).has_won();`?

In either case, it's a decent instinct to surround stuff with brackets but in Ruby people will find that surprising, and it isn't required in these cases.

views/game_one_player.erb
```diff
@@ -0,0 +1,35 @@
...
+<body>
+
+  <h1>Welcome, <%= @name%>!</h1>
+
+  <h3>Make your choice:</h3>
+  <form action = "/game_one_player" method = "post">
+    <input type = "radio" name = "choice" value = "rock" id = "rock" checked>Rock<br>
+    <input type = "radio" name = "choice" value = "paper" id = "paper">Paper<br>
+    <input type = "radio" name = "choice" value = "scissors" id = "scissors">Scissors<br>
+    <% if @which_game == "rpssl" %>
```
neoeno: Can you think of a name for this that a reader might be able to figure out at a glance? Without needing to think about it?

views/game_one_player.erb
```diff
@@ -0,0 +1,35 @@
...
+    <p>Computer has thrown <%=@thrown%>.</p>
+    <!-- <p><%=@won%>.</p> -->
+    <% if @won == :won %>
+      <h3>Congratulations, you won!</h3>
+    <% elsif @won == :lose %>
+      <h3>Unfortunately you lost. Try again, and maybe you will be lucky next time!</h3>
+    <% else %>
+      <h3>The game is tied. We need to replay to break the tie.</h3>
+    <% end %>
+  <% end %>
```
neoeno: This is how I'd indent my `if`s in ERB templates — rather than as it is in lines 15-18.

spec/game_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  subject(:game_scissors_rock) { described_class.new(scissors, rock)}
+  subject(:game_rock_paper) { described_class.new(rock, paper)}
+  subject(:game_paper_scissors) { described_class.new(paper, scissors)}
+  subject(:game_rock_rock) { described_class.new(rock, rock)}
+  let(:rock) {:rock}
+  let(:paper) {:paper}
+  let(:scissors) {:scissors}
+  let(:spock) {:spock}
+  let(:lizard) {:lizard}
```
neoeno: Perhaps you're finding it awkward to have this many `subject`s already. Good instinct here — though I think there are some more readable techniques. Here's how I'd do it:

```ruby
require 'game'

describe Game do
  def play_game_with_moves(move_1, move_2)
    described_class.new(move_1, move_2)
  end

  describe '#won?' do
    context 'first player wins' do
      it 'rock beats scissors' do
        expect(play_game_with_moves(:rock, :scissors).won?).to eq(:won)
      end
    end
  end
end
```

But with many more tests, obviously, as you have here. There are many more ways to make that more concise but that's as far as I'd go (mostly because people shout at me if I make my tests too clever).

Note my version doesn't use `subject` — it's not always compulsory, just usually a good standard to follow.

Good instinct setting up the `let`s in lines 22-26, though personally I'd leave them out as I think the symbols are as readable. Your version does guard against silly mistakes though (because if you typo'd `scissrs` it'd throw an undefined method error, whereas if you typo `:scissrs` it wouldn't), so I think it's reasonable.

spec/coumputer_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'computer'
+
+describe Computer do
+  subject(:computer) { described_class.new }
+  let(:options) { [:rock, :paper, :scissors, :spock, :lizard] }
+
+  it 'returns random variable within class' do
+    expect(options).to include(computer.weapon)
```
neoeno: If `computer.weapon` also returned `:evil_weapon` — how often would this tests fail? How many times do you have to run the tests to be 100% certain that you've covered all the options?

spec/features/registration_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+    scenario 'I want two players' do
+      register_two_players_rps
+      expect(page).to have_content "Welcome, Katerina! It's your turn now!"
+    end
+  end
+  context 'pick up which game you want to play' do
+    scenario 'I want one player' do
+      register_one_player_rpssl
+      expect(page).to have_content 'Welcome, Katerina!'
```
neoeno: I really like the way you've extracted out repetitive feature test logic into methods here. Though I'd give this a read: https://robots.thoughtbot.com/mystery-guest (it uses some techniques we haven't covered yet, but hopefully you'll get the idea)

lib/game.rb
kateloschinina: You are right, it's just a left-over =)

PR #595, opened 2017-01-22
Rick Clark pull request

rkclark:


rkclark:
Known issues based on code review:

- Not all game outcomes explicitly tested in unit tests, but I felt risk of the methods returning wrong results was low.
- Hound doesn't like some of my CSS syntax, but that was all done in a rush and not the major focus on the project anyhow :)
- I didn't need to stub randomness out of any feature tests as they test the screen output based on a two player game where I can control both moves in my web_helper methods. Maybe I should have added additional feature tests to cover this vs the computer opponent.
- I have presentation strings in business logic.... but I have encapsulated these within a MessageHandler class so they are there intentionally. The MessageHandler passes the messages into the MessageLog. I'd appreciate feedback on this approach!
-  There is some light logic in the views layer, for example, in game.erb the players' move icons are hidden/shown based on whether there is currently a Round in progress. e.g. `class="<%= 'hidden-xs-up' unless !@round.nil? %>"`. This seemed like an OK way to do it but I'd appreciate any feedback.
- The view also extracts messages from the message log and turns them into list items with `<%= @message_log.messages.reverse.map { |message| "<li>#{message}</li>" }.join %>`. Is this bad?
- I'm not sure what to think of my controller. It, I think by its very nature, ends up orchestrating the creation of many objects, and prompts those objects to complete actions when the game routing requires it. I'm not sure where it is on the "skinny scale" so, again, feedback would be great!
- I didn't have time to cover many edge cases, so the app crashes if you press to Choose Move without first selecting a move icon, for example.

Gemfile
```diff
@@ -1,15 +1,20 @@
...
   gem 'rubocop-rspec'
   gem 'rubocop'
   gem 'coveralls', require: false
   gem 'capybara'
   gem 'byebug'
 end
+
+group :development do
+  gem 'sass'
+end
```
neoeno: Ah, so you've used rails before perhaps? Correct me if I'm wrong, but there's no information on how to compile sass -> css here? Do correct me, I might have just missed it!

README.md
```diff
@@ -28,65 +20,49 @@ So that I can enjoy myself away from the daily grind
...
+I chose to pursue the bonus objectives from the start, these were:
 
+- Allow the choice of playing one player vs the computer or two player
+- Extend the game to use the 'special rules': Rock, Paper, Scissors, Spock, Lizard
 
-## Bonus level 1: Multiplayer
+I designed the following domain model to guide me in my implementation:
 
-Change the game so that two marketeers can play against each other ( _yes there are two of them_ ).
+[Domain Model](https://app.ardoq.com/presentation/makersacademyjan2017/587e15df72fa6d218cea11a3)
```
neoeno: :+1:!

README.md
```diff
@@ -28,65 +20,49 @@ So that I can enjoy myself away from the daily grind
...
-## Basic Rules
+**Testing**
+- RSpec
+- Capybara
 
-- Rock beats Scissors
-- Scissors beats Paper
-- Paper beats Rock
+**Front-end**
+- Bootstrap 4
```
neoeno: How you finding bootstrap 4, out of interest?

app.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  post '/names' do
+    player_1 = Player.new(name: params[:p1_name])
+    player_2 = Player.new(name: params[:p2_name], human: (params[:player_2_type] == 'human'))
+    Game.new(player_1: player_1, player_2: player_2)
+    message_handler = MessageHandler.new(message_log: MessageLog.new, game: Game.game_instance)
+    message_handler.enter_game_message
+    Round.round_instance = nil
+    redirect '/play'
+  end
```
neoeno: It's an interesting point you make about the controller having a bit much going on here. I'd probably consider using the façade pattern to encapsulate the whole game. In fact that might also surface some of the presentation/app logic mixing that's lurking around. There are probably other ways to do it though.

lib/message_handler.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def add_message_to_log(message)
+    message_log.add_message(message)
+  end
+
+  def find_loser(winning_move)
+    winning_move[0] == :player_1 ? game.players[:player_2].name : game.players[:player_1].name
+  end
+
+  def find_winner(winning_move)
```
neoeno: These two methods started to make me a bit nervous about the OO design here. I'm surprised that `MessageHandler`, which I suspected to be just receiving 'log' messages (e.g. "X joined the game", "X won over Y"), knew what a winner was and was being asked to find it.

Seems like the line in the views `@round.result_message` touches much more of the application than I expected. The domain could do with some refactoring I think.

That said, if I break `#find_winner` only one unit test fails — that's a really good sign of unit test isolation :+1:

lib/round.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def draw?
+    player_1_move == player_2_move
+  end
+
+  private
+
+  def handle_draw
+    @result_message = message_handler.draw_result_message(player_1_move)
```
neoeno: So the `MessageHandler` is responsible for generating messages? That makes more sense — I supposed it was responsible for receiving messages initially.

sass/main.scss
```diff
@@ -0,0 +1,167 @@
...
+//FOOTER STYLES
+
+#footer {
+  background-color: $color5;
+  font-size: 0.8rem;
+  color: $color1;
+  a {
+    color: $color2;
+  }
+}
```
neoeno: Sadly, though some Rails developers don't like it, CSS is still part of the code and we still have to apply best practices to it if we want to do good work! In this case that means consistent formatting, readable variable names. Good call on extracting the colours as variables though!

spec/unit_message_handler_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'message_handler.rb'
+require 'message_log.rb'
+require 'game.rb'
```
neoeno: Not 100% but I think we can drop the `.rb` here. Might be wrong.

Also suggest dropping the `unit_` prefix on the filenames unless you're really dead set on it and would argue the case in an interview situation, I can see what you're trying to do, but I think most devs would find it surprising.

spec/unit_message_handler_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    before(:each) do
+      allow(game).to receive(:players) { { player_1: player_1, player_2: player_2 } }
+      allow(player_1).to receive(:name) { "Beast" }
+      allow(player_2).to receive(:name) { "Magneto" }
+    end
+
+    def message_log_can_return(message)
+      allow(message_log).to receive(:add_message)
+      allow(message_log).to receive(:messages) { [message] }
+    end
```
neoeno: Nice :+1:

spec/unit_round_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  let(:round_class) { described_class }
+  let(:player_1_move) { :lizard }
+  let(:player_2_move) { :spock }
+  let(:rules_handler) { class_double("RulesHandler") }
+  let(:message_handler) { instance_double("MessageHandler") }
+  let(:game) { instance_double("Game") }
+  let(:winning_move) { [:player_1, [:lizard, :spock, "poisons"]] }
+
+  before do
+    MessageHandler.instance_variable_set("@message_handler_instance", message_handler)
```
neoeno: Is there a reason we can't pass this in? We've got all the dependency injection stuff set up so it seems odd not to use it. Reason I ask is because `#instance_variable_set` is more commonly used as an 'escape hatch' for when you really can't do anything else (so people will often question it when they see it).

views/game.erb
```diff
@@ -0,0 +1,141 @@
...
+                </label>
+              </div>
+              <div class="col-12">
+                <div class="row">
+                  <div class="col-8 offset-2">
+                    <div class="row justify-content-around">
+                      <div class="col-5">
+                        <label class="move-label">
+                          <input name="p1_choice" type="radio" class="hidden-input" value="lizard">
+                          <img src="/img/lizard.svg" class="move-icon">
```
neoeno: This much nesting/indentation in HTMl is a code smell for me. It points either to the views doing too much (in which case I'd break them up into partials or layouts), or the HTML structure itself being overly complex.

In this case it might be an over-reliance on Bootstrap. I suspect that because I've been the person writing this HTML before! It's really helpful but perhaps try omitting it in a project and working putting it together with CSS. It might take longer but it'll probably save your `<` & `>` keys a bit of wear!

views/game.erb
```diff
@@ -0,0 +1,141 @@
...
+            <div class="col-12">
+              <img src="/img/padlock.svg" alt="padlock icon" class="move-hider-icon">
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+    <div class="col-12 text-center pb-4" id="message_log_wrapper">
+      <ul>
+        <%= @message_log.messages.map { |message| "<li>#{message}</li>" }.join %>
```
neoeno:
```erb
<% @message_log.messages.each do |message| %>
  <li><%= message %></li>
<% end %>
```

Maybe.

views/layout.erb
```diff
@@ -0,0 +1,33 @@
...
+  </head>
+  <body>
+    <div id="main-wrapper">
+      <nav class="navbar pb-3 text-center">
+        <a href="/" class="navbar-brand">Rock, Paper, Scissors, Lizard, Spock!</a>
+      </nav>
+      <%= yield %>
+      <footer id="footer" class="py-2 px-3">
+        <div>Icons made by <a href="http://www.flaticon.com/authors/maxim-basinski" title="Maxim Basinski">Maxim Basinski</a>, <a href="http://www.flaticon.com/authors/pixel-perfect" title="Pixel perfect">Pixel perfect</a>, <a href="http://www.flaticon.com/authors/popcorns-arts" title="Popcorns Arts">Popcorns Arts</a>, <a href="http://www.flaticon.com/authors/vectors-market" title="Vectors Market">Vectors Market</a> and <a href="http://www.freepik.com" title="Freepik">Freepik</a> from <a href="http://www.flaticon.com" title="Flaticon">www.flaticon.com</a>, licensed by <a href="http://creativecommons.org/licenses/by/3.0/" title="Creative Commons BY 3.0" target="_blank">CC 3.0 BY</a></div>
+      </footer>
```
neoeno: I just want to credit *you* for your honourable crediting!

views/layout.erb
```diff
@@ -0,0 +1,33 @@
...
+    </div>
+    <script src="https://code.jquery.com/jquery-3.1.1.slim.min.js" integrity="sha384-A7FZj7v+d/sdmMqp/nOQwliLvUsJfDHW+k9Omg/a/EheAdgtzNs3hpfag6Ed950n" crossorigin="anonymous"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/tether/1.4.0/js/tether.min.js" integrity="sha384-DztdAPBWPRXSA/3eYEEUWrWCy7G5KFbe8fFjk5JAIxUYHKkDx6Qin1DkWx51bBrb" crossorigin="anonymous"></script>
+    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.6/js/bootstrap.min.js" integrity="sha384-vBWWzlZJ8ea9aCX4pEW3rVHjgjt7zpkNpZk+02D9phzyeVkE+jo0ieGizqPLForn" crossorigin="anonymous"></script>
```
neoeno: If you remove these script tags, does it still work?

views/results.erb
```diff
@@ -0,0 +1,58 @@
...
+        <div class="col-6 offset-3">
+          <h3 id="result_message"><%="#{@round.result_message}"%></h3>
+        </div>
+        <div class="col-4">
+          <div class="px-5">
+            <img class="results-move-icon img-fluid" src="/img/<%= @round.player_1_move.to_s %>.<%= @round.player_1_move == :spock ? 'jpg' : 'svg' %>" alt="player 1 move">
+          </div>
+        </div>
+        <div class="col-4 align-self-center">
+          <%= "<img id=\"win-indicator-arrow\" src=\"/img/right-arrow.svg\" class=\"img-fluid px-5 #{'flip-horizontal' if @round.winning_move[0] == :player_2}\" alt=\"arrow\">" unless @round.draw? %>
```
neoeno: There's quite a lot going on in this line — can you see any way it could be made more readable?

lib/round.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def handle_win
+    @winning_move = rules_handler.decide_winner(player_1_move, player_2_move)
+    game.update_score(winning_move[0])
+    @result_message = message_handler.battle_result_message(winning_move)
```
neoeno: This seems to me like the class exceeding its responsibilities. This reduces its *mobility* — the ability of a class to be reused in another context because it ends up tied up (coupled) with other concerns of the application.

Gemfile
rkclark: I've not used Rails, but have been using Sass for a while after learning about it on Treehouse (their course on it is great, its taught by the guy who wrote it!). I've used both the ruby gem version and the node version (via Gulp). To compile sass -> css in my project you would just run `sass sass/main.scss public/css/main.css`, or to have sass watch the scss file and auto-compile any changes, run `sass --watch sass/main.scss:public/css/main.css` :)

README.md
rkclark: It's awesome. The flexbox powered grid is super useful and they've made some quality of life improvements with the column class names. I also like the new padding and margin classes, i.e px-1 for left/right padding of scale 1 (i think its 1 rem, not sure).

app.rb
rkclark: Just looked up the facade pattern, which is really helpful. My struggle so far with OOD is the fact that I always feel I need to make bad design decisions somewhere, as something needs to be the 'orchestrator' and I haven't been able to figure out a way to avoid that. I guess the facade pattern is kind of recognising that something needs to do be responsible for creating the required objects, and is a good way to keep it out of the controller. I will try to use it!

lib/message_handler.rb
rkclark: So I created the MessageHandler as I was trying to separate concerns so that Round wasn't trying to construct messages. The Message handler constructs all of the app's messages and puts them in the MessageLog. I realised that I needed to reference the player names, and this required me to add a dependency with Game+Player somewhere. I didn't want to add it to Round as that class does a lot already, so I decided to do it in MessageHandler. I didn't feel good about it though! I also think that the winning_move should be accessed via a Struct so that the methods know less about the data structure.

lib/round.rb
rkclark: Yup

lib/round.rb
rkclark: Yes I agree!

sass/main.scss
rkclark: The colours intentionally have generic names, as the idea of using the variable-based colour pallette at the top of the main.scss file is that you can easily switch out the colours at any time to change the design. If I'd called $color2 "$lightblue" or something like that, then it would have been difficult to change it later to light red! However, I would agree that the names could be improved and could have names like $main-bkgrnd-color and $primary-tint or something like that, the same way that Bootstrap does if you look in the scss source files.

In terms of the CSS formatting, it was definitely rushed so some of the Hound comments I agree with, but some I don't. I understand the pros/cons of using id selectors and I don't see a problem in using them sparingly in the right circumstance, so I wouldn't agree with Hound that you should never have them!

spec/unit_message_handler_spec.rb
rkclark: Yes I was trying to find a way to distinguish between the unit and feature test files!

spec/unit_round_spec.rb
rkclark: So this bit is complicated, and I don't fully understand what is going on. However, at a certain point I started to get errors from rspec saying that a Game instance double was leaking from one test into another and this caused the test to fail. It took me ages to find a solution, which is the line you have queried above where I used instance_variable_set. It's something to do with the use of the class instance variables and that somewhere a long the line in the tests there is a  message handler instance that contains a Game instance double and it doesn't get refreshed between tests. By adding the line above I guess it forces a refresh of the class instance variable which fixed the leaking double.

views/game.erb
rkclark: Yep I realised the HTML structure got quite complex. I have done projects with and without bootstrap before :) I think breaking down the views as you suggest would be the best way forward. I also am not 100% up to speed with Bootstrap 4 yet (including it was a way of learning more, I don't want to neglect front end stuff too much during the course) so I've probably not done things in the most efficient way. A lot of this was done late on Sunday!

views/game.erb
rkclark: Yeah... so I didn't realise you could "break up' the erb in that way at the time. But I've got it now!

views/layout.erb
rkclark: Those are the Bootstrap 4 pre-requisites, so I guess it will depend on which bits of Bootstrap I am/am not using!

views/results.erb
rkclark: Yep, now I understand that you can break apart the erb bits this could be much clearer :)

views/layout.erb
neoeno: Might be worth finding out! Unneeded dependencies will often raise questions in a review.

spec/unit_round_spec.rb
neoeno: Aha — I guess that's another of the problems with global/pseudoglobal variables like class-level variables!

sass/main.scss
neoeno: I'd agree with what you've said about CSS variables.

Great you've got your own ideas on CSS & the way it should be done. I know I've got lots of opinions that differ with the style guides I've had to use. In general organisations expect people to follow to the set style guide & linting rules when they're writing code, and debate the actual rule as a separate matter. Give this a read for some context: https://blog.codinghorror.com/death-to-the-space-infidels/

Hard to anticipate in this case though since hound just pops in and checks CSS too without you having known it would! (Or at least I didn't know it would...)

lib/message_handler.rb
neoeno: Good thought process. Maybe `MessageGenerator` would be a better name? I guess coming from JS, 'handlers' always make me think of event handlers so subconsciously that's what I'm expecting. Could just be me though.

I'll leave the OOD thinking to you — think it'll just take working these problems through a few more times and it'll start to come together.

app.rb
neoeno: :+1: Good observation you had there! Maybe good to bear in mind the reasons why we do OOD — not just because we leave mean comments on your PRs if you don't! For me the main thing I try to bear in mind is flexibility — that is, the ability of the software to respond to change. It's not that OOD & patterns guarantee this, but that we can carefully apply them to get there.

In this case, I'd reason loosening the coupling between the controller and game logic by adding a façade (or interface) makes it easier to change one without changing the other. That said, the proof of the OOD pudding is in the eating, so if you try it and it seems bad, there might be a good reason & time to try something different.

PR #599, opened 2017-01-22
Weekend Challenge pull request

nenoch:


docs/personal_review.md
```diff
@@ -0,0 +1,16 @@
+# Self-Review - What can be done better.
+
+* Include how to download and run the app.
+* I didn't test all possible outcomes of the game, neither at feature nor at unit test level.
+* Long if/elsif trees and no Weapon class. However, I chose radio buttons to limit choices so there was no need to check if the weapon chosen by the user was available.
+* I created a Computer class (thinking about introducing a 2nd player later but run out of time) but completely forgot about duck-typing.
```
neoeno: Good catch!

docs/personal_review.md
```diff
@@ -0,0 +1,16 @@
...
+* I didn't test all possible outcomes of the game, neither at feature nor at unit test level.
+* Long if/elsif trees and no Weapon class. However, I chose radio buttons to limit choices so there was no need to check if the weapon chosen by the user was available.
+
+
+I wrote down questions here and there as comments marked with the friendly symbol:
+#::???::
+
+:)
+
+I also have a final question: why my coveralls went from 94% to 92% when I refactor getting rid of the global variable?
```
neoeno: Not totally sure on that one — could be that Coveralls is only covering unit tests... but that would surprise me. I'd start by running `coveralls report` and then maybe re-doing my change slowly, running coveralls on each change to see where the decrease occurs.

spec/computer_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'computer'
+
+describe Computer do
+  subject(:computer) { described_class.new }
+
+  describe "#select_weapon" do
+    it 'allows computer to select a weapon randomly' do
+      expect(computer.weapons).to include(computer.select_weapon)
+
+    end
```
neoeno: How many times would we have to run this test before we were sure `#select_weapon` didn't return something that wasn't in `#weapons`?

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      choose('rock')
+      click_button 'PLAY'
+      expect(page).to have_text 'You chose ROCK!'
+    end
+
+    scenario 'a player would like to play against the computer' do
+      # ::???::
+      # Is there a better way to test for randomness?
+      #
+      allow_any_instance_of(Array).to receive(:sample).and_return("paper")
```
neoeno: This is serviceable in this context! In more complex contexts we might run into trouble — for instance, if on line 32 run this code:

```ruby
p "Roll a 6 to win a million dollars: ", [1, 2, 3, 4, 5, 6].sample
```

The result would be:

```
"Roll a 6 to win a million dollars: "
"paper"
```

If any other code has to use "Array#sample" somewhere else before our test finishes, they might be disappointed not to win the million dollars.

But perhaps you're already convinced and that's why you asked! We could narrow down our stub to something like:

```ruby
allow_any_instance_of(Computer).to receive(:select_weapon).and_return("paper")
```

Still not ideal, but at least that way it's only our objects we're fiddling with.

spec/game_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'game'
+
+describe Game do
+  let(:player) { double :player }
+  let(:computer) { double :computer }
+  let(:name) { double :name }
+  subject(:game) {described_class.new(player, computer)}
```
neoeno: Nailing the doubles usage here! :+1:

lib/computer.rb
```diff
@@ -0,0 +1,23 @@
...
+  # I don't understand This
+
+  # Then *four* new objects will be created *every time you call `weapons`*  (what are the four objects?).  Use a constant with symbols instead:
+  #
+  # ```ruby
+  # WEAPONS = [:rock, :paper, :scissors]
+  # ```
+
+  def initialize
+    @weapons = ["rock", "paper", "scissors"]
```
neoeno: Take a look at the `#__id__` method below:

```ruby
2.3.3 :002 > "rock".__id__
 => 70154692093060
2.3.3 :003 > "rock".__id__
 => 70154679776700
2.3.3 :004 > "rock".__id__
 => 70154679757900
2.3.3 :005 > "rock".__id__
 => 70154679740100
2.3.3 :006 > "paper".__id__
 => 70154679722520
2.3.3 :007 > "paper".__id__
 => 70154679671260
```

You can use IRB and `#__id__` to figure out where new objects are being created every time, and where the old objects are reused. Contrast:

```ruby
2.3.3 :008 > :rock.__id__
 => 1155228
2.3.3 :009 > :rock.__id__
 => 1155228
2.3.3 :010 > :rock.__id__
 => 1155228
2.3.3 :011 > :paper.__id__
 => 1155548
2.3.3 :012 > :paper.__id__
 => 1155548
```

lib/game.rb
```diff
@@ -0,0 +1,48 @@
+require_relative 'computer'
+
+class Game
+
+  # ::???::
+  #
+  # What should be private?
+  # Not just in this class but in all classes of this app.
+  #
```
neoeno: Good question — the rule is: if a method needs to be used by any other bit of the application, it should be public.

Where you have to use your discernment is: which methods *should* be used by the rest of the application (public), and which should be hidden away (private)?

I usually find that I create methods in two situations:

1. I want the class to do a new thing
2. I want to simplify a method that already exists

Usually case 1 methods are public and case 2 methods are private.

E.g.:

```ruby
# Case 1
class Computer
  def insult_opponent
    adjective = ["terrible", "awful"].sample
    "You're terrible!"
  end
end

# Case 2
class Computer
  def insult_opponent
    "You're #{insulting_adjective}!"
  end

  private

  def insulting_adjective
    ["terrible", "awful"].sample
  end
end
```

The rest of the application has no business calling `#insulting_adjective` — so that's private.

(Obviously usually the method would be more complex before you thought about extracting a private method to simplify it)

rps.rb
```diff
@@ -0,0 +1,49 @@
...
+require './lib/game'
+require './lib/computer'
+
+#::???::
+#
+# Is the succession of GET / POST requests correct?
+# How could it be better?
+#
+# I am not sure about the use of variables in the controller and across POST and GET requests.
+# What is the best practice? How could it be better?
```
neoeno: I'm happy with the order (is that what you mean?) of get & post requests here. There are only four anyway so if we needed to add one it wouldn't take long to refactor!

I'd probably set up the variables in the controller just the way you've done. I might create a presenter object (worth googling), but that's probably overkill for here. `/challenge` is bordering on too much — could be a case for a sinatra helper: http://www.sinatrarb.com/intro.html#Helpers

PR #613, opened 2017-01-24
Konrad remerged PR by kay

neoeno:
@schlattk's code — added it here for ease of review


spec/game_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+describe Game do
+
+#subject(:game) {decribed_class.new(choice)}
+#let(:randomizer) {double :randomizer}
+
+  it "should generate a random choice between either rock, paper or scissors" do
+
+    game = Game.new("scissors")
+    expect (game.randomizer) == "rock" || "paper" || "scissors"
```
neoeno: Might be worth trying this line out in [rspec_playground](https://github.com/makersacademy/rspec_playground). Try these variants:

```ruby
expect ("cat") == "cat" || "mouse" || "dog"
expect ("giraffe") == "cat" || "mouse" || "harry" 
expect("draw").to eq("draw")
```

Have an explore and see which ones do what you want and which don't.

schlattk: Kay, thanks for your comments, really helpful. I am using Atom and I know I need to make more effort with my formatting. play ground is great and I will make use of it in the future, I realize that the or statement was going wrong and I should have changed it earlier. I am usually ok with git and did not have problems making pull requests on the two earlier challenges but with this one for some reason.. best Konrad

PR #623, opened 2017-03-05
Rock Paper Scissors game

FloraHarvey:


views/multi_player_result.erb
```diff
@@ -0,0 +1,25 @@
+<link rel="stylesheet" type="text/css" href="/style.css" />
```
maryrosecook: Could the `single_player_result.erb` and `multi_player_result.erb` be combined into one template? This would require several abstractions.  Happy to talk it through!

lib/computer.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'game'
+
+class Computer
+
+  attr_reader :options, :name, :score
+
+  def initialize
+    @options = Game::OPTIONS
+    @name = "The Computer"
+    @score = 0
```
maryrosecook: It feels like keeping score may not be in the user stories.

PR #624, opened 2017-03-05
RPS Challenge, Weekend 3

JohnChangUK:


app.rb
```diff
@@ -0,0 +1,38 @@
...
+    @game = Game.create(player, computer)
+    erb :play
+  end
+
+  before do
+    @game = Game.instance
+  end
+
+  post "/selection" do
+    player_choice = (@game.player.pick(params[:choose_your_weapon]))
```
maryrosecook: Could this code be rewritten to avoid violating the law of Demeter? eg could the game be given the information it needs (the player move) and then calculate the result internally? Or to put it another way, could you change the Game so that the /selection route can be just: 

```
post "/selection" do
  @game.set_player_move(params[:choose_your_weapon])
  @winner = @game.winner
  redirect '/result'
end
```

lib/computer.rb
```diff
@@ -0,0 +1,19 @@
+class Computer
+  attr_reader :name, :computer_selection
+
+  def initialize
+    @name = "AIcomp"
+  end
+
+  def selection
+    random = rand(1..3)
```
maryrosecook: Could you do this in a more extensible way that wouldn't require adding more logic like `return "Lizard" if random == 4`) if you add new moves?

lib/game.rb
```diff
@@ -0,0 +1,49 @@
...
+  def self.instance
+    @game
+  end
+
+  def selections(player_choice, computer_choice)
+    @games_array << player_choice
+    @games_array << computer_choice
+  end
+
+  def wins?
```
maryrosecook: Could the pluralisation of `wins?` be consistent with the pluralisation of `draw`?

lib/game.rb
```diff
@@ -0,0 +1,49 @@
...
+    WIN.include?(@games_array)
+  end
+
+  def draw?
+    DRAW.include?(@games_array)
+  end
+
+  def winner
```
maryrosecook: It feels like, in a subtle way, this method has two responsibilities.  The first is to say no one one.  The second is to say the name of the winner.  Could these be separated out?

PR #662, opened 2017-04-29
ROCK|PAPER|SCISSORS|LIZARD|SPOCK

adamjohnsnow:


lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+    [:scissors, :paper],
+    [:scissors, :lizard],
+    [:paper, :rock],
+    [:paper, :spock],
+    [:lizard, :spock],
+    [:lizard, :paper],
+    [:spock, :rock],
+    [:spock, :scissors]
+  ]
+  LOSING_HANDS = [
```
neoeno: You could ditch this one right? Or `WINNING_HANDS`? Remember no. 4 of [the 4 rules of simple design](https://martinfowler.com/bliki/BeckDesignRules.html)

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+    [:scissors, :spock],
+    [:paper, :scissors],
+    [:paper, :lizard],
+    [:lizard, :scissors],
+    [:lizard, :rock],
+    [:spock, :paper],
+    [:spock, :lizard]
+  ]
+  BEST_OF_DEFAULT = 5
+  WIN_STRING = ' WON THAT ROUND'
```
neoeno: Spend some time thinking about what 'Separation of Concerns' means as a software design principle.

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def play
+    p1win if WINNING_HANDS.include?([@player1.choice, @player2.choice])
+    p2win if LOSING_HANDS.include?([@player1.choice, @player2.choice])
+    draw if @player1.choice == @player2.choice
+  end
+
+  def declare_winner
+    @player1.wins > (@best_of / 2) ? @winner = @player1 : @winner = @player2
+    @player1.wins > (@best_of / 2) ? @loser = @player2 : @loser = @player1
```
neoeno: Read this method out loud to yourself — as if you were talking it through with a pair. Pay attention to the language you use.

What do we mean by 'readability'? How do we judge if code is 'readable'? Why is it important?

lib/game.rb
```diff
@@ -0,0 +1,81 @@
+class Game
+  attr_reader :player1, :player2, :p1_choice, :p2_choice,
+  :result, :best_of, :order, :winner, :loser
```
neoeno: Are all of these needed?

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+  def p2win
+    @result = [:p2win, "#{@player2.name.upcase}#{WIN_STRING}", "is beaten by"]
+    @player2.win
+  end
+
+  def draw
+    @result = [:draw, "NO ONE #{WIN_STRING}", "draws with"]
+  end
+
+end
```
neoeno: What is the responsibility of this class?

Too easy? Write down each of the methods and their responsibilities. Does each method have one responsibility? Then try to group them. Do they fit into one group? Or does it seem like there are separate groups inside them?

rps.rb
```diff
@@ -0,0 +1,82 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    @game = Game.instance
+    erb(:play)
+  end
+
+  post '/choice' do
+    Game.instance.order[0].log(params[:choice].to_sym)
```
neoeno: If you hadn't read `game.rb`, what would you be able to tell me about this line by reading it?

spec/features/feature_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    sign_in_2_players_and_play
+    find("#rock").click
+    expect(page).to_not have_button('Next')
+  end
+  scenario 'play 2 player game' do
+    sign_in_2_players_and_play
+    play_2_player
+    expect(page).to have_content("PLAYERTWO WINS! 3-0")
+  end
+end
```
neoeno: I'd leave a space between the scenarios — helps the eye pick them out as separate

views/index.erb
```diff
@@ -0,0 +1,26 @@
...
+    text-align: center;
+}
+</style>
+</head>
+<body>
+  <img src="RockPaperScissorsLizardSpock.jpg" height="580">
+  <h2>ROCK|PAPER|SCISSORS|LIZARD|SPOCK</h2>
+  <div class="container">
+  <p class="text-center">
+<form action ="/names" method='post' style="leftmargin:50">
```
neoeno: Indentation still matters in HTML unfortunately! Also, can you find a reference for what `leftmargin` does?

views/play.erb
```diff
@@ -0,0 +1,29 @@
...
+<form action ="/choice" method='post'>
+
+    <input type="image" id="rock" src="rock.png" value="rock" name="choice" height="180">
+    <input type="image" id="paper" src="paper.png" value="paper" name="choice" height="180">
+    <input type="image" src="scissors.png" value="scissors" name="choice" height="180">
+    <input type="image" src="lizard.png" value="lizard" name="choice" height="180">
+    <input type="image" src="spock.png" value="spock" name="choice" height="180">
+
+</form>
+<h1><%= @game.player1.name.upcase %> (Wins: <%= @game.player1.wins %>) vs. <%= @game.player2.name.upcase %> (Wins: <%= @game.player2.wins %>)</h1>
```
neoeno: I wonder if the views have too much control here — they can even play the game! It might be nice if we could just give them the information they needed to do their job and nothing else — then we'd be able to reason better about any bugs we encountered (e.g. this bug can't be coming from the view — it can only display information, not alter the game's state)

lib/player.rb
```diff
@@ -0,0 +1,21 @@
+class Player
+  attr_reader :name, :wins, :type, :choice
+
+  def initialize(name, type)
+    @name = name
+    @wins = 0
+    @type = type
```
neoeno: Whenever you find yourself writing a class with a `type` attribute or method — think 'is there a polymorphic relationship here'.

Is there a way you might have solved this problem with that technique?

PR #724, opened 2017-06-24
Initial pull request

jinimcoroneo:


app.rb
```diff
@@ -0,0 +1,37 @@
+require 'sinatra/base'
+require './lib/player'
+require './lib/game'
+
+class RPS < Sinatra::Base
+  enable :sessions
```
maryrosecook: Do sessions need to be enabled?

app.rb
```diff
@@ -0,0 +1,37 @@
...
+    @game = Game.create(Player.new(params[:player_name]))
+    redirect '/welcome'
+  end
+
+  get '/welcome' do
+    erb :welcome
+  end
+
+  post '/play' do
+    @game.player.choose(params[:choice]).to_sym
```
maryrosecook: Could the player's move be set without breaking the encapsulation of `game`?

app.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  before do
+    @game = Game.instance
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
```
maryrosecook: Love these clearly named routes and correspondingly clearly named templates: names, welcome, play

lib/game.rb
```diff
@@ -0,0 +1,44 @@
...
+  def self.instance
+    @game
+  end
+
+  def initialize(player, opponent_choice = nil)
+    @player = player
+    @opponent_choice = opponent_choice
+  end
+
+  def choose
```
maryrosecook: Does the name `choose` reveal the intention of this method?

lib/game.rb
```diff
@@ -0,0 +1,44 @@
+require "./lib/player.rb"
+
+class Game
+
+  GAME_RULES = {
```
maryrosecook: Does the name `GAME_RULES` reveal the intention of this hash? What data is it storing? What name would help the most for someone unfamiliar with the code?

lib/game.rb
maryrosecook: Is this behaviour in the right place? Does a game of RPS know how to generate a random move? Or could that behaviour be moved somewhere else?

lib/game.rb
```diff
@@ -0,0 +1,44 @@
...
+  def initialize(player, opponent_choice = nil)
+    @player = player
+    @opponent_choice = opponent_choice
+  end
+
+  def choose
+    @opponent_choice = [:rock, :paper, :scissors].sample
+  end
+
+  def win?
```
maryrosecook: Love these domain-revealing helper methods!

lib/player.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  attr_reader :name, :choice
+
+  def initialize(name, choice = nil)
+    @name = name
+    @choice = choice
+  end
+
+  def choose(choice)
+    @choice = choice.to_sym
```
maryrosecook: This is a nitpick, but should this method call `to_sym`, or should that be done in the controller, given that the controller understands how the view and model message each other?

spec/features/welcome_spec.rb
```diff
@@ -0,0 +1,50 @@
+# As a marketeer
+# So that I can see my name in lights
+# I would like to register my name before playing an online game
+
+feature 'see my name on welcome page' do
```
maryrosecook: Love all these clear tests!

views/play.erb
```diff
@@ -0,0 +1,63 @@
...
+      <form action="/welcome">
+      <input type="submit" class="button buttonblack" value="Play again?">
+      </form>
+    </div>
+
+  </body>
+</html>
+
+
+<style>
```
maryrosecook: Is there a way to DRY up this styling code?

app.rb
jinimcoroneo: I'm a bit confused with regards to this.  I can only access the player via the game, is there another way to do this using dependency injection?

maryrosecook: One option is to add a method to `Game` that will make the player choice. Another is to only create the game once the player's choice has been set.

PR #735, opened 2017-06-26
RPS weekend challenge attempt

alistairkung:


README.md
```diff
@@ -28,59 +28,6 @@ So that I can enjoy myself away from the daily grind
...
-  # Want a nice code coverage website? Uncomment this next line!
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
-
-You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
+To do if I had more time
+------------------------
+I would like to refactor my controller which currently looks too large. I would like to attempt to move some of the logic into the model layer.
```
maryrosecook: Good idea!

app.rb
```diff
@@ -0,0 +1,33 @@
...
+    redirect('/play')
+  end
+
+  before do
+    @game = Game.instance
+  end
+
+  post '/attack' do
+    move = params[:move].to_sym
+    @game.player.select_move(move)
```
maryrosecook: Is there a way to set the player's move without breaking the encapsulation of game?

app.rb
```diff
@@ -0,0 +1,33 @@
...
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    @game = Game.create(Player.new(params[:player_1_name]), Cpu.new)
+    redirect('/play')
+  end
+
+  before do
```
maryrosecook: Could this before block be moved to a place that is less surprising for the reader of the code?

app.rb
```diff
@@ -0,0 +1,33 @@
...
+    @game = Game.instance
+  end
+
+  post '/attack' do
+    move = params[:move].to_sym
+    @game.player.select_move(move)
+    @game.cpu.roll_attack
```
maryrosecook: Is there a way to set the CPU's move without breaking the encapsulation of game?

maryrosecook: It feels like the controller is driving the running of the game - it tells the `cpu` to `roll_attack` and then tells the game the `generate_result`.  Should the controller be doing both of these things?

lib/cpu.rb
```diff
@@ -0,0 +1,9 @@
+class Cpu
+
+  attr_reader :move
+
+  def roll_attack
```
maryrosecook: Does an external unit (in this case, the controller) need to trigger the calculation of the move?

lib/player.rb
```diff
@@ -0,0 +1,13 @@
+class Player
+
+  attr_reader :name, :move
```
maryrosecook: Really like this code in the way that the `Player` and `Cpu` have the same interface for getting the move (`move`).

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+  def self.instance
+    @game
+  end
+
+  def initialize(player = Player.new(name), cpu = Cpu.new)
+    @player = player
+    @cpu = cpu
+  end
+
+  def generate_result
```
maryrosecook: Definitely agree with what you said in the group code review - `generate_result` and `result` could be combined into one method.  Could you laser in on what was stopping you from managing this refactor? Happy to reflect with you on this!

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def sign_in_and_play
```
maryrosecook: Love this helper!

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+
+class Game
+
+  attr_reader :player, :cpu, :game_result
+
+  RULES = {
+    :rock     => {:rock => :draw, :paper => :lose, :scissors => :win},
+    :paper    => {:rock => :win, :paper => :draw, :scissors => :lose},
+    :scissors => {:rock => :lose, :paper => :win, :scissors => :draw}
+  }
```
maryrosecook: Go through and fix all the hound style violations that make sense. It's important to have cleanly formatted code.

spec/game_spec.rb
```diff
@@ -0,0 +1,74 @@
+require "game.rb"
+
+describe Game do
```
maryrosecook: Love these tests!

views/play.erb
```diff
@@ -0,0 +1,135 @@
...
+.ascii-art {
+    font-family: monospace;
+    white-space: pre;
+}
+
+</style>
+
+<body>
+
+  <div1 class="ascii-art">
```
maryrosecook: Is there a way to use partials to render the ASCII hands?

views/play.erb
```diff
@@ -0,0 +1,135 @@
...
+
+    <% end %>
+  </div2>
+
+  <h1 align ="center">
+    <%= @game.player.name %> vs CPU
+  </h1>
+
+  <div align="center">
+    <% if @game.game_result == :win %>
```
maryrosecook: Is there a way to use a helper method to automatically turn `game_result` symbols like `:win` into strings like `You won!`

PR #805, opened 2017-10-15
My submission

jonsanders101:


jonsanders101:
CODE REVIEW REQUEST

I'd like to request a code review for this weekend challenge. I've included some notes below on some points I've identified as parts of the codebase I need to work on with varying levels of "I know how I would go about this" confidence. I'd still like to talk through my code so I can challenge some of the flaws in my approach to problem solving.

Thanks!

**Self review notes**

* For some reason, my web_helpers.rb methods weren’t recognised by my feature_spec.rb and I ran out of time to investigate deeply why this should be, so I’ve left comments where these helper methods would go.
* Testing Game#play:
This test feels vacuous because it requires multiple stubs. It only tests that the value of `@result` changes, though this is the main behaviour of that method. It relies on the `Player#weapon_choice`and `Weapon#versus`, but these should be tested in their respective class specs (right?). Is the changing value of result what I should be testing? Is the difficulty in testing down to poor design in the method?
* I should use symbols instead of strings when evaluating the weapon choices. I can do this by converting the strings which are the values to the HTML buttons to symbols.
* My ‘/choice’ controlled in ‘app.rb’ is hideous and should delegate some of its logic to to the model, somehow.
* The ‘Game’ class is too large and has too many responsibilities, including:
- Being an interface which coordinates the other classes.
- Handling presentation logic with the Game#result_string method
- Handling turn-taking between players with Game#soft_reset
* The Weapon class might be better implemented as a module than a totally separate class.
* The division of objects into separate classes feels like overkill

PR #828, opened 2017-11-12
RPB Weekend challenge submission

gabrielabud:


gabrielabud:
to change ruby version

PR #860, opened 2018-01-21
Week3 RPS Challenge 

domvernon:


app/views/single_player.erb
```diff
@@ -0,0 +1,20 @@
...
+<h1>Rock-Paper-Scissors-Lizard-Spock<h1>
+
+<h3>Enter your name to begin</h3>
+
+<form action='/single-player' method='post'>
+  <label for="player1_name">
+    Player name:
+    <input type="text" name="player1_name">
+  </label>
+  <br><br>
```
EdwardAndress: Some thoughts on using <br> like this: https://stackoverflow.com/questions/1726073/is-it-sometimes-bad-to-use-br

app/main.rb
```diff
@@ -0,0 +1,62 @@
...
+    Game.start(player1, player2)
+    redirect '/play'
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post '/play' do
+    @game.player1.move = params[:player1_attack]
```
EdwardAndress: There are a few places where you 'reach through' one object to get to a method.  This contravenes something called the 'Law of Demeter', which (I find) a little tough to fully grasp but here's a summary: https://rails-bestpractices.com/posts/2010/07/24/the-law-of-demeter/

I.e. The guidance is to prefer `@game.player_one_move = ` rather than `@game.player_one.move =`.  This means the controller is coupled (needs to know about) to the game and not to the player as well.

Let me know if that doesn't make sense.

lib/game.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+  end
+
+  def initialize(player1, player2)
+    @player1 = player1
+    @player2 = player2
+  end
+
+  attr_reader :player1, :player2, :result
+
```
EdwardAndress: What do you think about this alternative to setting result in '#compare'?

```ruby
def result
  return :draw if draw?
  return :win  if player_one_win?
  return :lose if player_two_win?
end

def player_one_win?
  RULES[player1.move].include? player2.move
end

def player_two_win?
  RULES[player2.move].include? player1.move
end

def draw?
  player1.move == player2.move
end
```

lib/computer.rb
```diff
@@ -0,0 +1,11 @@
+class Computer
+  attr_reader :move
+
+  def initialize
+    @move = Game::RULES.keys.sample
+  end
+
+  def name
```
EdwardAndress: Since you were wondering where you could have tested more - this method is untested (in computer_spec.rb)

spec/features/helpers.rb
```diff
@@ -0,0 +1,12 @@
+def submit_single_name
```
EdwardAndress: nice use of helpers 👍

spec/features/multiplayer_spec.rb
```diff
@@ -0,0 +1,22 @@
+feature 'Multiplayer' do
+  scenario 'Player1 wins' do
+    submit_two_names
```
EdwardAndress: you could have a `before(:each)` block for `submit_two_names`

spec/game_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it 'returns the player object' do
+      expect(game.player1).to be player1
+    end
+  end
+
+  describe '#compare' do
+    before do
+      allow(computer).to receive(:move).and_return :Scissors
+    end
+
```
EdwardAndress: this would be a good place to use `context`

```ruby
context "player one wins" do
  it "returns :win" do
    <blah blah blah>
  end
end
```

spec/player_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+describe Player do
+  subject(:player) { described_class.new('Dom') }
+
+  describe '#name' do
+    it 'should return the name' do
+      expect(player.name).to eq 'Dom'
+    end
+  end
+
+  describe '#move' do
```
EdwardAndress: `#move` and `#move=` are two different methods so you should have a describe block for each.

EdwardAndress:
Nice work Dom.  I've tried to answer the questions that you raised, inline against the relevant bits of code.  When I checked out the repo the test coverage was 100% but one or two methods were only tested indirectly (i.e. some tests rely on them).

With regards to duplication in the templates, you're right to be concerned and it would be cool to find a way of DRYing that up.  In the review session we discussed a few approaches such as passing an array of players, iterating through that to build a form with the correct number of fields.  If you wanted to post a form to different routes depending on the state of the game you could try building the `action` property based on some value passed from the controller.

```
# I don't know if this would work but it the idea is that you set @player_number in the controller and # then re-use the same view for both player one and player two
<form action="/play#{@player_number}" method='post'>
```

EdwardAndress:
There are some other strategies which allow you to reduce the number of controller actions.  At some point one must weight up being DRY with

lib/game.rb
EdwardAndress: This might also be interesting: https://refactoring.guru/refactoring/techniques/simplifying-conditional-expressions

Thanks to @mdsomerfield

domvernon: I like this a lot. Very neat!

PR #861, opened 2018-01-21
RPS PR

lewmoore:


spec/features/feature_tests_spec.rb
```diff
@@ -0,0 +1,73 @@
+
+feature 'A player can start a game' do
+  scenario 'A player can enter their name to start a game' do
+    start_game
```
EdwardAndress: nice use of a helper

spec/game_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'game'
+
+describe Game do
+
+  subject(:game) { described_class.new(player1, player2) }
+  let(:player1) { double :player }
+  let(:player2) { double :player }
+
+  describe '#player1' do
```
EdwardAndress: Very minor point: I'd say "returns player one" rather than "knows".  Sometimes small things will grate on whoever is reviewing your code (in the professional environment).

spec/game_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    it 'Knows player 2' do
+      expect(game.player2).to eq player2
+    end
+  end
+
+  describe '#moves' do
+    context 'when players both choose paper' do
+      it 'Is a draw' do
+        expect(game.moves('Paper', 'Paper')).to eq "You both suck!"
+      end
```
EdwardAndress: the first context should close here

EdwardAndress: But the use of context is correct!

spec/person_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'player'
+
+describe Player do
+
+  subject(:player1) { described_class.new('Lewis') }
+  subject(:player2) { described_class.new('Frank') }
+
+  describe '#initialize' do
```
EdwardAndress: `#initialize` assigns the parameter to '#name' (rather than 'knows name').

views/play.erb
```diff
@@ -0,0 +1,40 @@
...
+  <label for="player1_move">
+    <%= @game.player1.name %> what's your move?
+  <select name="player1_move">
+    <option value="Rock">Rock</option>
+    <option value="Paper">Paper</option>
+    <option value="Scissors">Scissors</option>
+  </label>
+  </select>
+
+  <br></br>
```
EdwardAndress: You might find this interesting, with regards to using "<br>": https://stackoverflow.com/questions/1726073/is-it-sometimes-bad-to-use-br

spec/features/feature_tests_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    select('Rock', :from => 'player2_move')
+    click_button('Submit')
+    expect(page).to have_content "Frank chose: Rock"
+  end
+end
+
+feature 'Game results displayed on page' do
+  scenario 'Players both choose Rock' do
+    start_game
+    select('Rock', :from => 'player1_move')
```
EdwardAndress: How about another helper for selecting weapons?

```ruby
select_weapons('Rock', 'Rock')

def select_weapons(player_one_weapon, player_two_weapon)
  select player_one_weapon, from: 'player1_move'
  select player_two_weapon, from: 'player2_move'
end
```

lib/game.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def player1
+    @players.first
+  end
+
+  def player2
+    @players.last
+  end
+
+  def moves(player1_move, player2_move)
```
EdwardAndress: To me, the name of this method implies it will return some moves.  How about something like `#result` instead?

lib/player.rb
```diff
@@ -0,0 +1,8 @@
+class Player
+
+  attr_reader :name, :move
```
EdwardAndress: Is the reader for '@move' ever used?

spec/person_spec.rb
EdwardAndress: Did you change the name of this class at some point? The file is called person_spec.rb but the class is Player.

views/index.erb
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
```
EdwardAndress: Once you feel comfortable writing .erb files you could try another format such as .haml or .slim, which are more concise.  It's good to start with .erb though becuase it's closer to .html and a basic understanding of HTML is important.

views/play.erb
```diff
@@ -0,0 +1,40 @@
+<html>
+<body>
+<center>
+  <div class="header">
+  <h1>Lets play!</h1>
+</div>
+
+<%= @game.player1.name %> vs. <%= @game.player2.name %>
```
EdwardAndress: Check this out in relation to line 8: https://rails-bestpractices.com/posts/2010/07/24/the-law-of-demeter/

Let me know if you have any questions!

EdwardAndress: OK it's a post targeted at Rails but the first few points are still relevant.

EdwardAndress:
Hi Lewis, this is generally great.  You mentioned in the request for review that you were unsure about test coverage.  It seems fine to me.  Coverage is reported as 100%.  Looks like each method on all your 'models' are tested and you have feature tests that cover the normal paths through the app.  You could extend the feature tests to cover some 'unhappy' paths, such as when the user/s forget to enter their name/s.  What should happen then?  These tests would drive more coding rather than just increase coverage.

Were there any particular areas that you felt were not tested properly?  Are there some tests that you have written but feel uncomfortable about?  If so, please DM me as I might miss replies here.

PR #871, opened 2018-01-21
Tom Grand rps

Tagrand:


README.md
```diff
@@ -1,86 +1,68 @@
...
+## Approach <a name= "approach"></a>
 
+I set myself 4 targets starting this challenge 
+  
+   - *Never* use a global variable 
+   - Build one feature at a time 
+   - After passing tests, make sure they also fail for the **right reasons** 
+   - Focus on extendability in design.
+   
+I've focused alot on TDD and BDD over the past few weeks, and this week I found it much more natural. I've started seeing the benefits - debugging is easier as I'm working in smaller cycles, theres less issues to spot and naming tests carefully indicates clearly whats gone wrong. I also found doing one feature at a time helped structure my appraoch and keep my focus tight. I started with a site that could hold your name, then a site where you could pick rock, then one with three options, then one which told you win or lose. I found this helped structure my thoughts, and focus on just passing my tests.
```
EdwardAndress: I love the fact that you're taking time to reflect on, and document, your process.

app.rb
```diff
@@ -0,0 +1,44 @@
...
+    @game = Game.current
+  end
+
+  get "/" do
+    erb :index
+  end
+
+  post "/start" do
+    player_1 = Player.new(params["Enter name"])
+    if params["Player two"] == ""
```
EdwardAndress: How about `if params["Player two].empty?`?  It's a tiny bit more expressive.

app.rb
```diff
@@ -0,0 +1,44 @@
...
+    redirect "/play"
+  end
+
+  get "/play" do
+    erb :play
+  end
+
+  post "/choices" do
+    @game.player_1.move(params["attack"])
+    @game.player_2.move(params["attack_2"])
```
EdwardAndress: I understand correctly, line 33 will not do anything if the game is a one player game.  Getting to that point of understanding requires me to look back and forth between this and the `game.rb` file.  Given that, you could push this into the `Game` class and have all that stuff in one place.  So you'd end up with a method like `Game#choices` into which you could either pass the whole params hash or just the info required.

I'd also just use a getter and setter for `#choice` rather than set choice using `#move`.  But that would probaly happen inside the `Game` class.

EdwardAndress: Then you could think about whether there are any other places where you can push more code into the `Game` class.

lib/game.rb
```diff
@@ -0,0 +1,31 @@
+require_relative "player"
+require_relative "computer"
+
+class Game
+
+  RESULTS = { "rock" => "paper", "paper" => "scissors", "scissors" => "rock" }
+
+  def self.create(player_1, player_2 = Computer.new, two_player = false)
```
EdwardAndress: Would you ever want to set `two_player = true` when one of the opponents is a computer?  If not, you could have a method which sets this 'flag' (http://whatis.techtarget.com/definition/flag) dependent on whether one of the players `#is_a?(Computer)`

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+    @player_1 = player_1
+    @player_2 = player_2
+    @two_player = two_player
+  end
+
+  def winner
+    first = player_1.choice
+    second = player_2.choice
+    return "You drew" if first == second
+    RESULTS[first] == second ? "#{player_2.name} won!" : "#{player_1.name} won!"
```
EdwardAndress: This is a nice neat way of finding a winner without having to list out all possible combinations.

spec/game_spec.rb
```diff
@@ -0,0 +1,77 @@
+require "game"
+
+describe Game do
+  let(:player) { double :player, name: "Derek" }
+  let(:computer) { double :player, name: "Clement" }
+  let(:game) { described_class.new(player, computer) }
+
+  def set_choice(a, b)
```
EdwardAndress: This is a really nice abstraction that makes your tests super readable.  Nice work.  I saw that you also asked a question, in the review request, about testing random stuff.  One approach is to set the 'seed' (http://ruby-doc.org/core-2.2.0/Random.html#method-i-seed).  TL;DR: ruby only mimics randomness and you can, kind of, switch it off.

Taking this approach ensures that ruby's `Random` class is actually used.  Whereas stubbing `#choice` means you could pretty much write anything in that method body and the test would still pass.

views/play.erb
```diff
@@ -0,0 +1,28 @@
...
+</head>
+
+<body>
+ <h1> Welcome to the game, <%=@game.player_1.name%><h1>
+ <h2> Player 1 is the great  <%=@game.player_1.name%>  </h2>
+ <h2> Player 2 is the mighty  <%=@game.player_2.name%> </h2>
+   <form action = "/choices" method = "POST" >
+     <p> Player 1 </p>
+
+     <input type= "radio" id= "rock" name="attack" value = "rock" checked = "checked"> Rock
```
EdwardAndress: Are the IDs used?

EdwardAndress:
Really nice, overall.  There are a couple of particularly nice touches, which I have commented on.  I've tried to address two of your specific questions with comments.  To your final question about what goes where, I don't see any particular sins in your code, other than that perhaps you could be delegating more to your `Game` class, from the controller.

EdwardAndress:
Drop me a line on Slack if you have any questions!

PR #872, opened 2018-01-21
RPS Challenge

MissDove:


app.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  before do
+    @game = Game.instance
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post '/rock' do
```
mdsomerfield: It feels like you have too many routes here. Your routes for rock/paper/scissors all look suspiciously similar, as do the methods for these routes in the game class. 

Think DRY. Any time you see code that looks similar expect for one or two minor details (in this case, whether it is rock, paper or scissors) you should be asking 'should those details be state'? 

In your game class, you have made rock, paper and scissors all strings in a collection, so you've already identified them as specific instances of type of thing. We should treat them like this all the way down the stack. 

e.g.

```
post '/weapon' do
  @weapon = params[:weapon];
  @game.select_weapon(@weapon)
  @game.computer
  redirect '/selection'
end
```

lib/game.rb
```diff
@@ -0,0 +1,46 @@
+class Game
+
+  attr_reader :player, :selection, :computer_selection, :end_result
+
+  CHOICE = ['Rock', 'Paper', 'Scissors']
```
mdsomerfield: You've got two different collections CHOICE and BEATS that contain the same sets of strings. This means that you have to maintain an index state independently of the two collections, which feels wasteful and could also lead to bugs. If we're designing an OO system, then we should be able to interrogate the weapon itself for information about what it beats. 

The problem is that your weapons here are strings. This is a common reflex, but it's often the wrong one - read https://refactoring.guru/smells/primitive-obsession for a bit more on that. There are several ways we can refactor this; one way would be to encapsulate the comparison:

```
class Weapon

  attr_reader :name

  def initialize(name, beats)
    @name = name
    @beats = beats
  end

  def beats(opponent_weapon)
    @beats == opponent_weapon.name
  end 

end
```

Then we can initialise our weapons as:

```
WEAPONS = [
  new Weapon('Rock', 'Scissors'), 
  new Weapon('Scissors', 'Paper'), 
  new Weapon('Paper', 'Rock')
]
```

and our result becomes:

```
@selection.beats(@opponent_weapon)
```

You may wonder if this is better - it is probably more code. It is also probably more robust and extensible, because we're encapsulating the game behaviour better, so if we wanted to add additional weapons (I believe a version of this challenge calls for implementing rock, paper, scissors, lizard, spock), then we can keep the refactor simple.

lib/game.rb
```diff
@@ -0,0 +1,46 @@
...
+  def scissors
+    @selection = CHOICE[2]
+  end
+
+  def computer
+    @computer_selection = CHOICE[rand(3)]
+  end
+
+  def result
+    @end_result = 'lost'
```
mdsomerfield: In the example above, we encapsulated the weapon comparison. That example didn't include logic for handling a draw. What if we could initialise our weapons as follows:

```
WEAPONS = [
  new Weapon(:rock,        [:rock => :draw, :paper => :lose, :scissors => :win]), 
  new Weapon(:scissors', [:rock => :lose, :paper => :win, :scissors => :draw]), 
  new Weapon(:paper,      [:rock => :win, :paper => :draw, :scissors => :lose])
]
```

and in our weapon we returned

```
def result(@opponent_weapon)
  @outcomes[@opponent_weapon]
end
```

We've just described our game in an extensible way in a few lines of code.

PR #873, opened 2018-01-21
Code Review Request - Ed Goold

Gleoman:
Hi - I would really appreciate a code review of my weekend challenge attempt.  I am aware that it may not make for the most sophisticated code but any thoughts would be greatly appreciated.  Broader context is provided in my README-EG file.

Thanks!

Ed

### DIFF
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -7,7 +7,7 @@ class Game
   RULES = { rock: :scissors, paper: :rock, scissors: :paper }
 
   attr_reader :player
-  attr_accessor :opposition
+  attr_accessor :opposition, :result
 
   def initialize(player = Player)
     @player = player
@@ -16,11 +16,11 @@ class Game
 
   def winner
     if RULES[self.player.weapon] == @opposition
-      return "Player Wins!"
+      "Player Wins!"
     elsif RULES[@opposition] == self.player.weapon
-      return "Player Loses!"
+      "Player Loses!"
     elsif @opposition == self.player.weapon
-      return "A draw..."
+      "A draw..."
     end
   end
 
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -4,14 +4,27 @@ require_relative 'weapon'
 
 class Game
 
-  attr_reader
+  RULES = { rock: :scissors, paper: :rock, scissors: :paper }
 
-  def initialize
+  attr_reader :player
+  attr_accessor :opposition
 
+  def initialize(player = Player)
+    @player = player
+    @opposition = nil
   end
 
-  def method_name
-
+  def winner
+    if RULES[self.player.weapon] == @opposition
+      return "Player Wins!"
+    elsif RULES[@opposition] == self.player.weapon
+      return "Player Loses!"
+    elsif @opposition == self.player.weapon
+      return "A draw..."
+    end
   end
 
+  def computer
+    @opposition = [:rock, :paper, :scissors].sample
+  end
 end
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -1,0 +1,17 @@
+require_relative 'player'
+require_relative 'computer'
+require_relative 'weapon'
+
+class Game
+
+  attr_reader
+
+  def initialize
+
+  end
+
+  def method_name
+
+  end
+
+end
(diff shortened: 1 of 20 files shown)


lib/game.rb
```diff
@@ -0,0 +1,30 @@
+require_relative 'player'
```
mdsomerfield: You mention in your code review that extracting out the computer class feels like overkill - how would you define overkill? 

A better question to ask is - what is the responsibility of this class?

It might be easy to answer that with 'model the game', but that's kind of like saying 'to manage the application' - i.e. its general and non-specific, which makes it very hard to think usefully about what code should belong there and what code should not.

A better approach might be 'the responsibility of the game class is to know what kinds of weapons you can choose'. Or you might decide it is to 'decide whether one weapon beats another'. 

Either way, you can then look at the class and cross out any lines that don't match that responsibility. In either of the above examples, it feels like the computer selection is not the responsibility of this class.

mdsomerfield: Another way to think about this is - should the game care _how_ the computer chooses a weapon? It might be some super complex machine learning algorithm. Or it might just be a random sample. The game doesn't care and doesn't want to be concerned with such details.

lib/game.rb
```diff
@@ -0,0 +1,30 @@
...
+  attr_reader :player
+  attr_accessor :opposition, :result
+
+  def initialize(player = Player)
+    @player = player
+    @opposition = nil
+  end
+
+  def winner
+    if RULES[self.player.weapon] == @opposition
```
mdsomerfield: Whenever you see multiple 'if' statements, think 'can I refactor that into a class or a data structure'? There is a whole set of techniques for this here: 

https://refactoring.guru/refactoring/techniques/simplifying-conditional-expressions

The technique I would go with here is to move the logic into a hash map, something like

```
rules = {
    rock: {rock: :draw, paper: :lose, scissors: :win},
    paper: {rock: :win, paper: :draw, scissors: :lose},
    scissors: {rock: :lose, paper: :win, scissors: :draw}
}
```

and when you want to get the winner, you can then just assess:

`rules[@player_weapon][@opponent_weapon]`

spec/game_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  let(:player) { double :player, name: "name" }
+  let(:rock) { double :rock }
+
+    it 'records the player as an instance variable' do
+      expect(subject.player).to eq :player
+    end
+
+  describe '#computer' do
+    it 'chooses a weapon at random' do
+      expect([:rock, :paper, :scissors]).to include(subject.computer)
```
mdsomerfield: This doesn't really look like you're testing anything here. If it chooses the same weapon every time (i.e. not at random) your test will still pass.

Do you think you can effectively test random behaviour? I would suggest that what you are really testing here is that 'the computer selects a weapon' and you want to stub the randomness.

spec/game_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    it 'assigns that weapon to the opposition instance variable' do
+      example = subject.computer
+      expect(subject.opposition).to eq example
```
mdsomerfield: This doesn't feel like a useful test - you're effectively testing that attr_accessor works. If you test behaviour that makes use of subject.computer then you can implicitly test this. 

Test behaviour, not state.

lib/game.rb
```diff
@@ -0,0 +1,30 @@
+require_relative 'player'
+require_relative 'computer'
+require_relative 'weapon'
+
+class Game
+
+  RULES = { rock: :scissors, paper: :rock, scissors: :paper }
+
+  attr_reader :player
+  attr_accessor :opposition, :result
```
mdsomerfield: Whenever you find yourself using attr_accessor, that's a sign that you haven't encapsulated your code properly. Or to put it another way, if one class feels the need to mess around with the state of another class, is the state in the right place?

Why do we need other classes to have write access to the result and opposition - is this just for test purposes? Always prefer dependency injection in that case - it's very rare that there's a good reason to use attr_accessor, unless the class that you're writing is specifically designed as a representation of data (e.g a model in a database).

PR #882, opened 2018-01-30
For giada! :)

EdwardAndress:


app.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  post "/name" do
+    session[:player] = Player.new(params[:player])
+    redirect "/start"
+  end
+
+  get "/start" do
+    @player = session[:player]
+    erb :play
+  end
```
EdwardAndress: I'd expect there to be an action like `post '/play'` which is used when the weapon choice form is submitted, because you're setting a value on the player object.

EdwardAndress: Actually, all I changed was the name of one action and the method of another.

lib/ai.rb
```diff
@@ -0,0 +1,5 @@
+class Ai
+  def choice
+    ["rock", "paper", "scissors"].sample
+  end
+end
```
EdwardAndress: I like this implementation because it's really simple and readable.

lib/game.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'player'
+require_relative 'ai'
+
+class Game
+
```
EdwardAndress: This long method totally works and is quite a natural way to build methods at first, _but_, it is really hard to read and also hard to change / maintain.  You'd almost certainly see bugs creeping in here.

It's possible to clean this up by using various forms of hash.  For example, one that you 'key into' using the player and Ai weapon choices like this:

OUTCOME_HASH[:player_one_weapon][:ai_weapon]

Can you imagine what that hash might look like?

Alternatively, you could try to get rid of hashes: https://github.com/EdwardAndress/rps-weapons

NOTE: My weapons are probably a pain to extend because every time I add a new weapon to the game I'll have to update all existing weapons.

EdwardAndress: You might also find this interesting: https://refactoring.guru/refactoring/techniques/simplifying-conditional-expressions

spec/ai_spec.rb
```diff
@@ -0,0 +1,8 @@
+describe Ai do
+
```
EdwardAndress: Testing random things is a bit tricky.  Have a play with the `srand()`.  Try doing this:

```ruby
10.times do
 srand(1); puts [1,2,3,4,5,6,7,8,9,10].sample
end
```

What happens?  Can you use this in your code somewhere? :)

spec/features/testing_infrastructure_spec.rb
```diff
@@ -0,0 +1,52 @@
+# require "pry"
+
+feature "Entering player name" do
+  scenario "Player can enter a name at the start of the game" do
+    sign_in_and_play
```
EdwardAndress: nice use of a helper method here to dry up the code

spec/features/testing_infrastructure_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    choose("scissors")
+    click_button("Play!")
+    expect(page).to have_content("You played: scissors")
+  end
+
+  scenario "Player can see ai's move" do
+    sign_in_and_play
+    choose("scissors")
+    click_button("Play!")
+    expect(page).to have_content("The computer played:")
```
EdwardAndress: You might also be able to use srand here to control what the computer chooses and then actually test that the weapon choice appears on the screen.  Alternatively you could use a regex since you'll have already tested that the AI chooses a weapon in the correct way.

```ruby
expect(page).to have_content(/The computer played: (rock|paper|scissors/))
```

Have you come across regular expressions?  They are a bit intimidating at first but super useful.  If you want to play with them and see how to match various things http://rubular.com/ is great.

spec/features/testing_infrastructure_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+end
+
+feature "Display the winner" do
+  scenario "Player can see if they win" do
+    sign_in_and_play
+    choose("scissors")
+    allow_any_instance_of(Array).to receive(:sample).and_return("paper")
```
EdwardAndress: OK, nice stubbing here with `allow_any_instance_of`

spec/game_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  describe "#play_match" do
+
+    context "rock - rock" do
+      it "works out the result" do
+        expect(game.play_match("rock", "rock")).to eq("It's a draw!")
+      end
+    end
+
+    context "rock - paper" do
+      it "works out the result" do
```
EdwardAndress: The `describe` / `context` and `it` blocks form part of your documentation.  For `it` blocks it's helpful if they state the expected outcome.  So, for example, `it "declares player_one the winner" do`

EdwardAndress: Otherwise good use of `context`

EdwardAndress:
Hi @GiadaSimonetti nice work generally.  I think the main point is to think about how you can remove that big if/else statement.  There are a few other comments too.  I hope it's helpful.

PR #883, opened 2018-02-11
Full personal reflection/ code review in 'README.md' file

marcusfgardiner:


README.md
```diff
@@ -1,3 +1,40 @@
...
+* I used the TDD cycle throughout development, and stuck to using this technique solidly. For each stage/feature: I wrote feature tests, then units tests, got the unit tests passing then the feature tests, then did a live test of the site. I found when modelling specifically that process diagraming was helpful to collect my thoughts
+* After I satisfied all user stories, I provided a link to return to the start of the game and gave basic CSS a go as I had never done this before (e.g. adding images to the background of the first two pages)
+
+
+Mini personal code review: My appraisal of my attempt at the weekend challenge
+---------
+* Overall I was pleased with my approach and outcome for this challenge. I feel like I applied a lot of what I had learned doing last weekends challenge, particularly driving development through testing and breaking things down using algorithmic thinking at each point
+* I found that really letting the tests DRIVE development, constantly running RSpec and having that give the next direction, was really helpful and enabled me to build everything in a lean way
```
EdwardAndress: 👍

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def initialize(player_1, player_2)
+    @player_1 = player_1
+    @player_2 = player_2
+  end
+
+  def self.instance
+    @game
+  end
+
```
EdwardAndress: Since you already passed in both players on instantiation, it feels weird to pass a player into `#choose_move`.  Did you choose to do this so that the method can be used for either player?  If so, is there definitely a scenario where you'd want to automatically choose the move for player 1?

app.rb
```diff
@@ -0,0 +1,35 @@
...
+  get '/play' do
+    erb :play
+  end
+
+  post '/make_choice' do
+    @game.player_1.set_move(params[:choice])
+    redirect 'game_over'
+  end
+
+  get '/game_over' do
```
EdwardAndress: I'd be inclined to send one message to the game object in this action `#calculate_winner` and let the game take care of sending the message to set player 2's move.

```ruby
get '/game_over' do
  @game.calculate_winner
  erb :game_over
end
```

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+    if player_1.move == player_2.move
+    elsif (player_1.move == 'rock' && player_2.move == 'scissors') ||
+        (player_1.move == 'scissors' && player_2.move == 'paper') ||
+        (player_1.move == 'paper' && player_2.move == 'rock')
+      @winner = player_1
+    else
+      @winner = player_2
+    end
+  end
+
```
EdwardAndress: There are a few different ways to get rid of this if statement.  Step one would be to abstract the rules into an attribute or constant.  Think about how you might do that and, also, the code you want inside this method.  For example could you end up with something like this?

```ruby
def calculate_winner
  @winner = RULES[player_one_move][player_two_move]
end
```

EdwardAndress: Here's my attempt to make RPS battles readable https://github.com/EdwardAndress/rps-weapons

There's one big flaw in terms of maintainability though.  Can you work out what it is?

lib/player.rb
```diff
@@ -0,0 +1,14 @@
+class Player
+
+  attr_reader :name
+  attr_accessor :move
+
+  def initialize(name)
+    @name = name
+  end
+
+  def set_move(move)
```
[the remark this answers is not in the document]
EdwardAndress: Listen to the cries of houndci

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def sign_in_and_play
```
EdwardAndress: Is it signing in and playing or just signing in?

spec/game_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  let(:alternate_game) { game_class.create(first_player_loses, second_player) }
+  let(:first_player_loses) { double('Player 1', name: 'Marcus', move: 'paper') }
+
+  let(:first_player_draw) { double('Player 1', name: 'Marcus', move: 'rock') }
+  let(:draw_game) { game_class.create(first_player_draw, second_player) }
+
+  describe '#initialize' do
+    it 'Instantiates game with player' do
+      expect(game.player_1.name).to eq 'Marcus'
```
EdwardAndress: Accessing the name of player one like this breaks the 'law of demeter'.

Check this out.

https://rails-bestpractices.com/posts/2010/07/24/the-law-of-demeter/

spec/game_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+    it 'Instantiates game with second player' do
+      expect(game.player_2.name).to eq 'Gamblore'
+    end
+  end
+
+  it 'Stores an instance of game in a class variable' do
+    game
+    expect(game_class.instance).to be_instance_of game_class
```
EdwardAndress: Would this work?
```ruby
expect(game.class.instance).to be_instance_of Game
```

EdwardAndress: -> it removes the need for line 26

spec/game_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    describe '#random_move' do
+      it 'assigns a random move if none is given' do
+        allow(first_player).to receive(:set_move)
+        expect(game).to receive(:random_move)
+        game.choose_move(first_player)
+      end
+    end
+  end
+
+  describe '#calculate_winner' do
```
EdwardAndress: This could be made more readable by using contexts and moving some of the let blocks down here.  For example:

```ruby

let(:player_one)  { double :player, move: player_one_move }
let(:player_two)  { double :player, move: player_two_move }
subject { described_class.new(player_one, player_two) }

describe '#calculate_winner' do
  context 'player one chooses rock player two chooses scissors' do

    let(:player_one_move) { "rock"        }
    let(:player_two_move) { "scissors" }

    it 'returns player one' do
        expect(game.calculate_winner).to eq player_one
    end
  end
end
```

EdwardAndress: ^ I realise this also implies a slight change to the functionality of game

spec/player_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'player'
+
+describe Player do
+
+  subject(:player) { Player.new('Marcus') }
+
+  describe '#initialize' do
```
EdwardAndress: Are you testing '#initialize', '#name' or both?  I think this reads better as a test of '#name'

lib/player.rb
```diff
@@ -0,0 +1,14 @@
+class Player
+
+  attr_reader :name
+  attr_accessor :move
```
EdwardAndress: This accessor for `#move` means your `#set_move` method is not necessary.

spec/player_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+  end
+
+  describe '#set_move' do
+    it 'sets a players move' do
+      player.set_move('rock')
+      expect(player.move).to eql('rock')
+    end
+  end
+end
```
EdwardAndress: You could also test `Player#move` (the attribute reader).

views/game_over.erb
```diff
@@ -0,0 +1,28 @@
...
+    <%= @game.player_1.name %> chose
+      <%= @game.player_1.move %>
+  </div>
+  <div>
+    <%= @game.player_2.name %> chose
+      <%= @game.player_2.move %>
+  </div>
+</h2>
+<h1>
+  <div>
```
EdwardAndress: In the group code review we talked about how you can push some logic into the controller and then into the model.

You could try a implementing a `Game#display_result` method which handles all the logic and just returns the correct string.  Or, a bit more of a challenge, you could try implementing a `ResultPresenter` class, to be used in the view, wiyj a method that takes a game object and returns the correct string.

views/index.erb
```diff
@@ -0,0 +1,21 @@
+<html>
+  <head>
+    <link rel='stylesheet' type='text/css' href='/index.css'>
+  </head>
+  <div>
+    <h1>Launch control initiated.
```
EdwardAndress: I think that nesting a <div> inside an <h1> is not semantically correct.  I.e. an <h1> element is the most important heading on the page.  If you include a <div> in the <h1> that meaning is also applied to the <div>.  It's mostly an accessibility issue (i.e. think screen readers).

README.md
```diff
@@ -1,3 +1,40 @@
...
+* There were a few situations where being new to web apps made tests hard to write e.g. ensuring a random outcome was fixed for Capybara testing
+
+What I would do next
+---------
+* The app is in a basic form: I would like to add a lot more style to it as well as more features. Given its scope, I am happy with it
+* I could also add more tests and refer to specific parts of webpages to cover every possible choice combination and make the testing more robust
+* I am keen to gain a lot more practice in HTML (tags and attributes) and CSS syntax, as well as creating more complex web apps using a Model View Controller structure
+
+Questions from the exercise:
```
EdwardAndress: Great questions here! :)

EdwardAndress: It's a nice reflection on the whole too.

README.md
```diff
@@ -1,3 +1,40 @@
...
+* in the model: Research a better way to calculate RPS winner (or statement very long). Could a hash of potential choices be used?
+* Did I need to use a post/redirect/get loop for the 3 RPS button choices? Was there another way/ best practice way of doing this?
+* Research how you know which part of array is returned by srand
+* Confirm whether stubbing should definitely NOT be done as part of feature tests (I used srand to fix a variable outcome)
+* Check whether git repo clones work with images saved to public folder - do the images download and work on someone else's computer?
+* In label tag, what does 'for' attribute do?
+* Why does continous integration not pass on pull request - error is that ruby version is not correct - gemfile correctly specifies ruby version to use.
+
```
EdwardAndress: With regards to the failing build.  This is a really good learning opportunity, I think ;) How will you find the answer?  I think the answer lies in understanding a bit more about Travis.

EdwardAndress:
Hi Marcus,

I made a ton of comments, which I hope are helpful.  Generally a really good effort.  The reflection that you do will really help a lot and you seem to have a good idea about where the code could have been improved, which is awesome.  Let me know if you have any questions.

I've tried to point you in the right direction with regards to the failing build but if you don't crack it come and chat.

Eddie

app.rb
marcusfgardiner: Makes sense

lib/game.rb
marcusfgardiner: As you expand out the number of permutations, you would need a huge number of classes and methods. If you used a hash instead, you could just add to that hash

spec/features/web_helpers.rb
marcusfgardiner: Good point, only signing in!

spec/game_spec.rb
marcusfgardiner: True :)

spec/game_spec.rb
marcusfgardiner: Bit uncertain about this one. Going to add it to me question bank. I am guessing the wrapper method would do the calling of the other method for you directly from game e.g. game.player_name ?

spec/player_spec.rb
marcusfgardiner: Makes sense

lib/player.rb
marcusfgardiner: I was using set_move to try and provide a wrapper method for setting move, but didn't know if this was neccesary

views/game_over.erb
marcusfgardiner: Sounds good - yep I understood the logic extraction from the group discussion, so you could have e.g. @game.win? in the view and ask the model without doing any string comparison

README.md
marcusfgardiner: Travis yaml updated! :)

PR #892, opened 2018-02-11
Submit RPS Challenge

MatthewBurstein:


app.rb
```diff
@@ -0,0 +1,46 @@
+require 'sinatra/base'
+require './lib/human_player'
+require './lib/computer_player'
+require './lib/game'
+
+class RockPaperScissors < Sinatra::Base
+
+  before(/(\/game)\/?\w*/) do
+    @game = Game.current_game
+  end
```
soph-g: Remove code if it's not being used - consider checking out onto another branch if you want to experiment

views/loss.erb
```diff
@@ -0,0 +1,13 @@
...
+You chose <%= @game.results[:loser] %>.
+</h3>
+<h3>
+The computer chose <%= @game.results[:winner] %>.
+</h3>
+<h1>You lose!!</h1>
+
+<form action="/game" method="get">
+  <button type="submit" name="play again">Play again!</button>
+</form>
```
soph-g: Repetition in the Win/Lose/Draw views - does it make sense to have three different pages? Try to DRY out displaying the result

spec/features/play_and_win_and_start_new_game_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'win a game and click new game button' do
```
soph-g: Lots of feature tests which is great, but consider collating these to make the test files easier to navigate. For example, you could have feature test for 'play a game' and win/lose/draw scenarios within that feature test.

lib/game.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def self.current_game
+    raise "No current game" unless @current_game
+    @current_game
+  end
+
+  attr_reader :player_1, :player_2, :moves, :results
+
+  DEFAULT_MOVE_LIST = %w(rock paper scissors)
```
soph-g: Is this used within Game?

lib/game.rb
```diff
@@ -0,0 +1,54 @@
...
+    raise "You must choose a move" unless player_1_move
+    draw_result(player_1_move) if player_1_move == player_2_move
+    win_result(player_1_move, player_2_move) if @rules[player_1_move.to_sym] == player_2_move
+    loss_result(player_1_move, player_2_move) if @rules[player_2_move.to_sym] == player_1_move
+    @results
+  end
+
+  private
+
+  def draw_result(player_1_move)
```
soph-g: Repetition within the different result methods - could these be simplified and DRYed out? For example, the win/lose/draw message is for the user, so is it necessary to store which player has won/lost?

lib/human_player.rb
```diff
@@ -0,0 +1,9 @@
+class HumanPlayer
```
soph-g: Could this class be made more similar to ComputerPlayer, so Game could use them in the same way? Perhaps look into duck typing :)

soph-g:
Hi Matthew,

I've added a few comments to your code, overall I think this is a good attempt at the challenge and you've clearly understood how to build a Sinatra app. And, I enjoyed your README!

The main things I would recommend focussing on are reducing repetition in your code, and looking for opportunities to simplify the logic wherever possible. I think spending time on refactoring could help with both of these.

Let me know if any of my comments aren't clear, or you'd like to discuss further.

Thanks,

Sophie

PR #895, opened 2018-02-11
Serena Hathi - RPS game (week 3)

serenahathi:


lib/opponent.rb
```diff
@@ -0,0 +1,11 @@
+require_relative 'game'
+require_relative 'player'
+
+class Opponent
+
+  attr_reader :opponent_option
+
+  def option
+    @opponent_option = Game::OPTIONS.sample
```
soph-g: Couple of things here:
1) Should the Opponent class depend on information from the Game class?
2) Could the opponent_option and player.move methods be made more similar from the Game's perspective? Maybe look into duck typing.

lib/game.rb
```diff
@@ -0,0 +1,36 @@
...
+    elsif @player.option == "Rock" && @opponent.opponent_option == "Scissors"
+      "#{@player.name} wins"
+    elsif @player.option == "Scissors" && @opponent.opponent_option == "Paper"
+      "#{@player.name} wins"
+    elsif @player.option == "Paper" && @opponent.opponent_option == "Rock"
+      "#{@player.name} wins"
+    else
+      "Opponent wins"
+    end
+  end
```
soph-g: Agree that this method could be simplified. Maybe look at ways you could store the rules and then use the rules to determine the outcome.
Also agree that it would be good to look at moving the string values into the view. Maybe the view could determine which string to display, based on the game result?

soph-g:
Hi Serena,

I've just done a code review and left a few comments. Overall it seems that you understand how to build a Sinatra app which is great! Also I really liked your README.

There are a few opportunities for improvement, I think spending a bit more time refactoring might help you with simplifying the logic in the Game, for example. With regards to testing, your feature tests are comprehensive, however you might find feature tests more useful if they flow more naturally with the user stories and features you are implementing. This would definitely be easier to discuss so feel free to grab me for a chat.

If there's anything in my comments that's unclear, or you'd like to discuss further, please let me know.

Thanks,

Sophie

PR #896, opened 2018-02-11
Rock Paper Scissors PR

lwkchan:


lib/computer_player.rb
```diff
@@ -0,0 +1,15 @@
+class ComputerPlayer
```
soph-g: Not sure this class is being used anywhere - if not then it's good practice to remove, or use a different branch if you want to experiment. Can always merge it back onto master if the experiment turns into code you want to use.

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def set_player_weapon(weapon)
+    @player_weapon = weapon.to_sym
+  end
+
+  def result
+    case player_win?
+    when true
+      "#{@player.name} wins this game"
```
soph-g: The result strings should live in the view, and the view should render the appropriate string based on the game result. However, like the use of true/false/nil to determine win/lose/draw

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+    when false
+      "Computer wins this game"
+    when nil
+      "It's a draw"
+    end
+  end
+
+  private
+
+  def player_win?
```
soph-g: Is there a simpler way to store the 'rules' of the game and then determine the win/lose/draw outcome. Maybe think about an appropriate data structure for the rules.

lib/game.rb
```diff
@@ -0,0 +1,68 @@
+class Game
+
+  attr_reader :player, :player_weapon, :pc_weapon
+
+  def initialize(player)
+    @player = player
+    @pc_weapon = random_weapon
+    @result = result
```
soph-g: Does 'result' need to be set on initialize?

lib/game.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def self.create(player)
+    @game = Game.new(player)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def set_player_weapon(weapon)
```
soph-g: Should the player weapon live in Game or in Player? At the moment, Player class is very small, consider moving all information about the player into this class or perhaps you could just use Game.

spec/features/gameplay_spec.rb
```diff
@@ -0,0 +1,54 @@
+feature "Gameplay" do
+
+  it "gives player options of rock, paper and scissors" do
+    sign_in_and_play
+    expect(page).to have_button('Rock')
+    expect(page).to have_button('Paper')
+    expect(page).to have_button('Scissors')
+  end
+
+  describe "Announcing player's chosen weapon" do
```
soph-g: Really good use of describe blocks within these tests.

soph-g:
Hi Laura,

I've added some comments to your code, overall it's looking good - I appreciated the README and feature tests in particular :)

If you want to go further with this exercise then I think there are two main things you could look at:
- implementing sessions in Sinatra
- Refactoring the result calculation to simplify this code and reduce repetition.

Please let me know if there's anything you'd like to discuss further.

Thanks,

Sophie

PR #902, opened 2018-02-12
new pull request - travis problems 

Kotauror:


app.rb
```diff
@@ -0,0 +1,71 @@
...
+  get '/play' do
+    @single_game = SingleGame.instance
+    erb(:play)
+  end
+
+  post '/attack' do
+    @single_game = SingleGame.instance
+    @computer_weapon = @single_game.computer_weapon
+    @weapon = params[:weapon]
+    @result = @single_game.calculate_result(@computer_weapon, @weapon)
```
maryrosecook: Encapsulation.  This code is not really respecting the encapsulation of SingleGame.  It's feeding data in (the moves) and pulling it out (the result).  The code could instead focus on what object is needed.  To decide that, have a look at attack.erb and see what data its displaying.  What concept does all this data centre on?

lib/multi_game.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'player'
+
+class MultiGame
```
maryrosecook: DRY. To DRY up MultiGame and SingleGame, think about what is similar and what is different.  Try and extract one tiny similarity to a place both can share.  Try to extract another similarity.  Keep on doing that.  Try to avoid inheritance.  A more versatile way to share code is a module.

PR #989, opened 2018-06-04
Updates Rock, paper, scissors

chrisrusselldigital:


lib/rps_game.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def score
+    @scores = []
+    @scores << [@choice, @rps]
+  end
+
+  def referee
+    if @choice == @rps
+      draw
```
mdsomerfield: Try and avoid complex if/else statements. Houndci has already picked up on this (cyclomatic complexity) - here are some strategies for refactoring that you could try: https://refactoring.guru/refactoring/techniques/simplifying-conditional-expressions. The relevant ones are either 'consolidate conditional expression' or 'replace conditional with polymorphism'. In the latter case, you don't need to create actual classes, you can achieve the same result with a hash or an array. 

For example, consider how you might write this method if you had:

```
[ 
  { :choice => :rock, :beats => :scissors },
  { :choice => :scissors, :beats => :paper },
  { :choice => :paper, :beats => :rock }
]
```

Now imagine that you needed to add 4 more options to the game - is it going to be easier to extend the array above, or the if statement you have written?

spec/rps_game_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rps_game'
+
+describe RpsGame do
+
+  it 'the computer selects rock, paper or scissors' do
+    # setup
```
mdsomerfield: I like this setup, exercise, verify - I use arrange, act, assert (http://defragdev.com/blog/?p=783) but I think it amounts to the same thing.

spec/rps_game_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    # exercise
+    game.rps_selector
+    # verify
+    expect(game.rps_selector).to match 'Rock|Paper|Scissors'
+  end
+
+  it 'takes the input from a user' do
+    # setup
+    game = RpsGame.new
+    # exercise
```
mdsomerfield: One reason using a framework like setup/exercise/verify is useful is that it can highlight problems with tests. One thing to look out for - if there is no 'exercise', then the test probably isn't a useful test, as it's probably not testing behaviour.

PR #1004, opened 2018-07-29
Tim's RPS

tblanchard01:
It works, but I would like to: 

1) update formatting
2) get RSpec working for dropdown list.


app.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  post '/names' do
+     session[:player_1_name] = params[:player_1_name]
+    redirect '/play'
+  end
+
+  post '/weapon' do
+    session[:player_1_choice] = params[:choice]
+    session[:player_2_choice] = ['rock', 'paper', 'scissors'].sample
```
alicelieutier: This looks like logic that could move out of the controller.

views/result.erb
```diff
@@ -0,0 +1,6 @@
+
+<h2> You chose: <%= session[:player_1_choice] %> </h2> 
+
+<h2> Your opponent chose: <%= session[:player_2_choice] %> <h2> 
+
+<h1> <%=Logic.find_winner(session[:player_1_choice], session[:player_2_choice]) %> </h1>
```
alicelieutier: This looks like too much logic for the view. You might want to extract it.

PR #1008, opened 2018-07-29
Marek's RPS(SL) Challenge

Mschmidt19:
My entire project is in one commit because when I finished the original app, I forked my own repo and continued the bonus objectives on the new fork. Then I realized I could no longer create a pull request here because of this. I tried a few options, one of which was to force push everything. That worked, however, it made me 68 commits behind the original fork and I again couldn't make a pull request. I finally had to delete my repo again and manually copy and paste each file after cloning in order to keep the old commit history. Luckily it worked, but my commit history was lost on this repo. It is saved locally, however.

Sorry if this hurts the reviewing process!


alicelieutier:
Could you maybe copy/paste the `git log` output? Just to get an idea of the order you took things in.

app.rb
```diff
@@ -0,0 +1,54 @@
...
+    @game = Game.instance
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post '/result' do
+    @game.player1.make_choice(params[:attack_choice])
+    @game.player2.make_random_choice
```
alicelieutier: This looks like a lot of logic for the controller.
Could there be a method in game to "play a turn"?

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+  def winner
+    p1_index = Player::CHOICES.index(@player1.choice)
+    p2_index = Player::CHOICES.index(@player2.choice)
+    return nil if p1_index == p2_index
+    return @player2 if ((p1_index + 1) % 5) == (p2_index % 5)
+    return @player2 if ((p1_index + 3) % 5) == (p2_index % 5)
+    @player1
+  end
+
+  def loser
```
alicelieutier: Isn't the loser just the player who did not win? This method looks like you are repeating a bunch of logic from the other one on top.

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+    @player1 = player1
+    @player2 = player2
+  end
+
+  attr_reader :player1, :player2
+
+  # This method looks at the indexes of player choices
+  # in relation to the CHOICES array in Player class
+  # You lose if your opponent's choice is to the right or 3 right of yours
+  # And you win if your opponent's choice is to the left or 3 left of yours
```
alicelieutier: The impulse to create shorter methods is good, but shouldn't come at a cost to readability, and this method does not seem very readable (which is why sometimes you will hear "if you need a comment, your code could probably be written to be more transparent").

Also, what if you suddenly added two more choices, or changed the order of choices in Player? Your game class at the moment seems heavily coupled to your Player class.

spec/features/play_again_spec.rb
```diff
@@ -0,0 +1,18 @@
+feature "User can play again" do
+  context "At end of game, with a winner" do
+    it "Redirects to play page with same player name" do
+      allow(Kernel).to receive(:rand).and_return(2)
```
alicelieutier: Can you comment on what this line does, for future code readers?

views/result.erb
```diff
@@ -0,0 +1,95 @@
...
+  <head>
+    <title>Result</title>
+    <link href="<%= url('/styles.css') %>" rel="stylesheet" type="text/css" />
+    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.2.0/css/all.css" integrity="sha384-hWVjflwFxL6sNzntih27bfxkr27PmbbK/iSvJ+a4+0owXq79v+lsFkW54bOGbiDQ" crossorigin="anonymous">
+  </head>
+  <body>
+    <div class="column centered">
+      <p class="centered subtitle white_text"><%= @game.player1.name %> chose <%= @game.player1.choice %></p>
+      <div class="centered">
+        <% if @game.player1.choice == "rock" %>
```
alicelieutier: Could this logic be placed somewhere else? (a `get_image` method maybe?)

PR #1036, opened 2018-09-23
rps challenge

m-rcd:


lib/computer.rb
```diff
@@ -0,0 +1,12 @@
+class Computer
```
samjones1001: Computer feels like a potentially unnecessary class here. Ideally, the application shouldn't care if a player in the game is a human or a computer - the Player class could allow for an option to be picked, or simply generate one at random if not option is chosen.

spec/features/homepage_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'spec_helper'
+require './app'
+
+RSpec.feature 'Testing homepage' do
+  scenario 'have game title' do
+    visit '/'
+    expect(page).to have_content 'ROCK PAPER SCISSORS LIZARD SPOCK!'
+  end
+
+  scenario 'user can choose 1 player' do
```
samjones1001: This test and the one after it feel a little redundant.  If the button doesn't exist, then other feature tests which rely on the user being able to click the button will fail, so you test the existence of the button implicitly.

I think these types of tests are a great stepping stone to use as part of the TDD process, but they can - and should - be deleted once you have better quality tests in place.

spec/unit/computer_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+describe Computer do
+  subject(:computer) { described_class.new }
+
+  context '#initialize' do
+    it 'set the name as Skynet' do
+      expect(computer.name).to eq 'Skynet'
+    end
+  end
+
+  context '#choice' do
```
samjones1001: These tests aren't really doing anything. 

You're explicitly telling your array to return a specific value, and then checking if that value has been returned. 🙂

spec/unit/game_spec.rb
```diff
@@ -0,0 +1,260 @@
...
+    end
+
+    it 'has a default' do
+      game = Game.new(player1)
+      expect(game.player2).to be_an_instance_of Computer
+    end
+  end
+
+  context '#play versus computer' do
+    context 'computer choice is paper' do
```
samjones1001: While this is definitely super well tested, I don't think you need to test every possible outcome of the game - if you later wanted to add further options this could quickly get to be unmanageable 🙂 

I think a test for a draw, a test for a player one win, and a test for a player two win would be enough.

PR #1037, opened 2018-09-23
rsp challenge

ShinyVerse:


app.rb
```diff
@@ -0,0 +1,54 @@
...
+  get '/multiplayer' do
+    @game = session[:game]
+    erb :multiplayer
+  end
+
+  post '/selection' do
+    @game = session[:game]
+    choice = params[:choice]
+    p2_choice = params[:p2_choice]
+    @game.player1.choice_setter(choice)
```
mdsomerfield: There's a development principle known as the law of demeter, with a converse anti-pattern known as a 'trainwreck' which is probably worth searching and reading about. Essentially, rather than the above, better would be:
```
@game.set_choice(choice, :player1)
```
Why? Because we're letting @game decide how that gets implemented, so if I change the way in which choices are set it doesn't break the code here because it is encapsulated by the Game class. You probably covered some of this in a workshop on delegation in week 2 (I don't actually know if that workshop definitely happened, so if not and that's a new concept then please come chat).

mdsomerfield: Also, on the naming - notice I have used 'set_choice' rather than 'choice_setter'. Try to prefer method names that are verbs which describe the intent of the thing that is being done, it makes it easier to read code and clearer what the code is doing.

lib/game.rb
```diff
@@ -0,0 +1,26 @@
+class Game
+  attr_reader :winner, :player1, :player2
+  def initialize(player1, player2)
+    @player1 = player1
+    @player2 = player2
+    @winner = nil
+    @player_one_wins = [
+      "rock-paper",
```
mdsomerfield: Rather than encoding the logic in strings, I would prefer something like { :winner => 'rock', :loser => 'paper' }. It makes it clearer what's going on and also easier to extend. We want the structure of the data to reflect the structure of the logic.

lib/game.rb
```diff
@@ -0,0 +1,26 @@
...
+    if player_one_wins.include?("#{player1.choice}-#{player2.choice}")
+      @winner = player1.name
+    else
+      @winner = player2.name
+    end
+  end
+
+  private
+
+  attr_reader :player_one_wins
```
mdsomerfield: Why do you have an attr_reader for this? It feels like this is logic that should be internal to the class. I think there are a few other attr_readers you could probably remove. Remember, we care about keeping things private and encapsulation.

PR #1039, opened 2018-09-23
Weekend Challenge 3: RPS

cristinaocanamanzano:


README.md
```diff
@@ -1,24 +1,44 @@
-# RPS Challenge
+RPS Challenge
+==================
+This is my proposal to build the Makers Academy Weekend 3 Challenge: Rock, Paper, Scissors
 
-Instructions
+Setup
```
soph-g: Good guide on how to run the app locally. Should also include how to run tests.

app.rb
```diff
@@ -0,0 +1,44 @@
...
+  post '/options' do
+    @game = Game.instance
+    @game.choose_player_1_option(params[:chosen_option])
+    @game.choose_player_2_option(@game.computer_option)
+    redirect '/result'
+  end
+
+  get '/result' do
+    @game = Game.instance
+    @result = @game.result(@game.player_1_option, @game.player_2_option)
```
soph-g: Do you need to pass in arguments to this method? Seems like the game should be able to access player_1_option and player_2_option when the result method is called, as you are calling these on game!

app.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  post '/names' do
+    player_1 = Player.new(params[:player_1])
+    player_2 = Computer.new("The Machine")
+    @game = Game.create(player_1, player_2)
+    redirect '/play'
+  end
+
+  get '/play' do
+    @game = Game.instance
```
soph-g: Good use of class methods

lib/game.rb
```diff
@@ -0,0 +1,49 @@
...
+  def self.instance
+    @game
+  end
+
+  def result(option_1, option_2)
+    return "tie" if option_1 == option_2
+    RULES[option_1.to_sym] == option_2.to_sym ? player_1 : player_2
+  end
+
+  def computer_option
```
soph-g: Lines 30 - 48 - I'm not clear why these different methods are needed. They are all doing similar things so I think these could be refactored.

lib/player.rb
```diff
@@ -0,0 +1,15 @@
+class Player
+  attr_reader :name
+  attr_accessor :chosen_option
+
+  OPTIONS = ["rock", "paper", "scissors"]
```
soph-g: This constant is only used in the Computer class, not sure it should be defined here.

lib/player.rb
```diff
@@ -0,0 +1,15 @@
+class Player
+  attr_reader :name
+  attr_accessor :chosen_option
```
soph-g: Do you need to have attr_accessor for chosen_option, and a setter method called choose_option?

spec/features/result_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  scenario "Shows player's 2 option" do
+    allow_any_instance_of(Array).to receive(:sample).and_return('paper')
+    sign_in_and_play
+    click_button "Rock"
+    expect(page).to have_content "The Machine chose paper"
+  end
+
+  scenario "Shows winner if not a tie" do
```
soph-g: Would be good to show both user and computer winning

PR #1043, opened 2018-09-23
First Pull Request

afaraone:
Review in forthcoming request

### DIFF
--- a/app/app.rb
+++ b/app/app.rb
@@ -27,12 +27,12 @@ class RPS < Sinatra::Base
   post '/play' do
     @game = session[:game]
     if @game.on_turn == session[:p1]
-      session[:choice1] = params['choice'].to_sym
+      session[:choices] = [params['choice'].to_sym]
       redirect '/result' unless @game.multiplayer?
       @game.switch
       redirect '/play'
     else
-      session[:choice2] = params['choice'].to_sym
+      session[:choices] << params['choice'].to_sym
       @game.switch
     end
     redirect '/result'
@@ -40,7 +40,7 @@ class RPS < Sinatra::Base
 
   get '/result' do
     @game = session[:game]
-    @game.make_move(session[:choice1], session[:choice2])
+    @game.make_move(session[:choices])
     @game.calculate_winner
     @game.tie? ? erb(:tie) : erb(:winner)
   end
--- a/app/app.rb
+++ b/app/app.rb
@@ -9,9 +9,12 @@ class RPS < Sinatra::Base
   end
 
   post '/' do
-    p params
     session[:p1] = Player.new(params[:name])
-    params[:name2].empty? ? session[:p2] = Computer.new : session[:p2] = Player.new(params[:name2])
+    if params[:name2].empty?
+      session[:p2] = Computer.new
+    else
+      session[:p2] = Player.new(params[:name2])
+    end
     session[:game] = Game.new(session[:p1], session[:p2])
     redirect('/play')
   end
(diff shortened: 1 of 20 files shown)


app/app.rb
```diff
@@ -0,0 +1,49 @@
...
+
+class RPS < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb(:welcome)
+  end
+
+  post '/' do
+    session[:p1] = Player.new(params[:name])
```
soph-g: Try to avoid using sessions, instead look at the later stages of the Battle App challenge to see how you can use Class methods and variables instead of sessions.

app/models/game.rb
```diff
@@ -0,0 +1,69 @@
+require_relative 'computer'
+require_relative 'player'
+
+class Game
+  MOVES   = [:rock, :paper, :scissors, :spock, :lizard]
+  P1_WIN  = [[:scissors, :paper], [:paper, :rock], [:rock, :lizard],
+            [:lizard, :spock], [:spock, :scissors], [:scissors, :lizard],
+            [:lizard, :paper], [:paper, :spock], [:spock, :rock],
+            [:rock, :scissors]]
+  P2_WIN  = P1_WIN.map { |combo| combo.reverse }
```
soph-g: How could you avoid this hidden duplication?

app/models/game.rb
```diff
@@ -0,0 +1,69 @@
+require_relative 'computer'
+require_relative 'player'
+
+class Game
+  MOVES   = [:rock, :paper, :scissors, :spock, :lizard]
+  P1_WIN  = [[:scissors, :paper], [:paper, :rock], [:rock, :lizard],
```
soph-g: How could you simplify this code and make it easier to understand?

app/models/game.rb
```diff
@@ -0,0 +1,69 @@
+require_relative 'computer'
+require_relative 'player'
+
+class Game
+  MOVES   = [:rock, :paper, :scissors, :spock, :lizard]
```
soph-g: How could you remove the duplication between moves and P1_WIN?

app/models/game.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def tie?
+    self.winner == :tie
+  end
+
+  def make_move(choices)
+    players.each_with_index do |player, index|
+      if player.computer?
+        player.move = Game::MOVES
```
soph-g: Should aim to call the same methods on all players, rather than needing to change behaviour based on the type of player. What if the computer class had a move method that returned a random choice...

app/models/game.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def switch
+    self.on_turn = opposite_of(on_turn)
+  end
+
+  def loser
+    opposite_of(winner)
+  end
+
+  def tie?
```
soph-g: Is this method used?

app/models/player.rb
```diff
@@ -0,0 +1,12 @@
+class Player
+  attr_reader   :name
+  attr_accessor :move
+
+  def initialize(name)
+    @name = name
+  end
+
+  def computer?
```
soph-g: Is this really required? Would be best not to have to check what something is.

spec/features/game_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  scenario 'tie' do
+    sign_in_multiplayer
+    make_choice('paper', 'paper')
+    expect(page).to have_content('The game is tied!')
+  end
+end
+
+# feature 'single-player' do
```
soph-g: Best to fix or remove commented out tests

spec/units/game_spec.rb
```diff
@@ -0,0 +1,147 @@
+require './app/models/game'
+
+describe Game do
+  let(:foo) { double(:mockplayer1) }
```
soph-g: remember to name your doubles something that makes sense in the context of the tests e.g. player1, player2

PR #1044, opened 2018-09-23
RPS challenge

AntonioDaria:
I will need to complete this challenge the user story where we declare the winner is still missing


app.rb
```diff
@@ -0,0 +1,30 @@
...
+class RpsWeb < Sinatra::Base
+
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    session[:name] = params[:name]
```
soph-g: Try to avoid using sessions and store the player name in an appropriate class instance instead.

app.rb
```diff
@@ -0,0 +1,30 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    @name = session[:name]
+    @shape = session[:shape]
```
soph-g: Make sure that your controller is interacting with the Game model to get information as much as possible, rather than using sessions

capybara-201809222039387627995796.html
```diff
@@ -0,0 +1,10 @@
+<h2>GAME TIME!!!!</h2>
```
soph-g: Are these views duplicates? Looks like these have been created by Capybara but I haven't seen this behaviour before. Worth checking how/why these views were created and removing them if they are not required.

PR #1056, opened 2018-09-24
pos

djstephenson7:


app.rb
```diff
@@ -0,0 +1,25 @@
...
+
+class RPS < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    session[:name] = params[:name]
```
soph-g: Try storing the player name in the Game, so avoiding the use of sessions.

app.rb
```diff
@@ -0,0 +1,25 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    @name = session[:name]
+    erb :play
+  end
+
+  get '/result' do
+    @decision = params[:decision]
```
soph-g: Try using the Game class to calculate the result and give a winner.

app.rb
```diff
@@ -0,0 +1,25 @@
...
+    erb @game.result
```
soph-g: erb needs to be given a path to the view it should render.

lib/computer.rb
```diff
@@ -0,0 +1,8 @@
+require './lib/game'
+
+class Computer
+
+  def computer_choice
```
soph-g: Could you rename this method to make the Computer and Player classes more similar? Game then wouldn't need to know what type of players player1 and player2, it could interact with all players in the same way.

lib/game.rb
```diff
@@ -0,0 +1,16 @@
...
+class Game
+
+  OPTIONS = [:Rock, :Paper, :Scissors]
+  RULES = {
+    Rock: :Scissors,
+    Paper: :Rock,
+    Scissors: :Paper
+  }
+
+  def player_move(weapon)
```
soph-g: Try updating the weapon in Player class when the player makes a move.

lib/player.rb
```diff
@@ -0,0 +1,8 @@
+class Player
+  attr_reader :weapon
+
+  def weapon(weapon)
```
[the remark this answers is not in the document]
soph-g: Try to resolve this error, you could try creating a 'getter' method that shows the player weapon, and a 'setter' method that sets the player weapon. If you're not sure what I mean try researching the difference between getter and setter methods in Ruby.

spec/features/computer_spec.rb
```diff
@@ -0,0 +1,12 @@
+require './lib/computer'
+
+describe Computer do
```
soph-g: Make sure only feature tests live in the 'features' folder - any test that is checking behaviour of a class is a unit test

spec/features/player_spec.rb
```diff
@@ -0,0 +1,10 @@
+require './lib/player'
+
+describe Player do
+  subject(:player) { described_class.new }
+  describe '#weapon' do
+    it "raises error" do
+      expect{ (Game::OPTIONS).exclude? player.weapon(Game::OPTIONS) }.to raise_error('not a possible weapon')
```
soph-g: What is this test testing? Is there a simpler way to check that you can't pass in a String that isn't a valid option?

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,10 @@
+def player_sign_in
```
soph-g: Good use of web helpers

views/play.erb
```diff
@@ -0,0 +1,14 @@
+<center>
+<h1>Welcome, <%= @name %></h1>
+
+<div>
+  <form action="/result" method="get">
+    <label for="player_move"><h1>Choose: Rock, paper or scissors.</h1></label>
+    <input type="text" name="player_move">
```
soph-g: Is 'text' the best input type for choosing a weapon? If you're not sure, try investigating the different types of input you can use in html forms and try changing this to use an input well suited for a user picking from a defined number of options.

views/result.erb
```diff
@@ -0,0 +1,7 @@
+<%= @game.result %>
```
soph-g: Try updating the app.rb to use this view when get `/result` is called

PR #1063, opened 2018-10-20
RPS

JL-J:


app.rb
```diff
@@ -0,0 +1,39 @@
...
+    redirect '/results'
+  end
+
+  get '/results' do
+    erb :results
+  end
+
+  run! if app_file == $0
+
+end
```
runticle: These routes are nice and "Skinny" :) 
I'm not sure if it's good or not to have a Get and Post method with the same name as it could get confusing

lib/game.rb
```diff
@@ -0,0 +1,48 @@
...
+    answer = evaluate
+    winner(answer)
+  end
+
+  private
+
+  def winner(answer)
+    return "The winner is #{@player_1}" if answer == 'player_1_win'
+    return "The winner is the computer" if answer == 'player_2_win'
+    return "It's a draw" if answer == 'draw'
```
runticle: Apparently it is good practise to have the text in the winner method somewhere in the "presentation" layer. ie, winner should return ":win, :lose or :draw" and then for example the app.rb file may have three routes, "/win", "/lose" and "/draw" etc with the congratulations messages in each one.

runticle: this is better for the open/closed principle, whereby it is easy to change the presentation code without having to change 'low-level' stuff

spec/computer_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      expect(computer.choice).to eq 'Paper'
+    end
+    it 'it returns scissors' do
+      allow(Kernel).to receive(:rand).and_return 3
+      computer = Computer.new
+      expect(computer.choice).to eq 'Scissors'
+    end
+  end
+
+end
```
runticle: nicely stubbed!

spec/evaluate_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+    it 'returns player 2 as the winner' do
+      expect(subject.evaluate(["Scissors", "Rock"])).to eq "player_2_win"
+    end
+    it 'returns its a draw' do
+      expect(subject.evaluate(["Scissors", "Scissors"])).to eq "draw"
+    end
+  end
+
+end
```
runticle: good, but apparently we should be testing eeeevery result

lib/game.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def result
+    answer = evaluate
+    winner(answer)
+  end
+
+  private
+
+  def winner(answer)
+    return "The winner is #{@player_1}" if answer == 'player_1_win'
```
runticle: if you have `@player1.name` instead, the results page will show it!

PR #1064, opened 2018-10-20
Completed RPS Challenge

MHUS25:


README.md
```diff
@@ -1,86 +1,68 @@
...
+* The computer/bot will choose a random option
+* A winner will be declared
+* User can click 'Play again' and start a new game
 
-## Bonus level 1: Multiplayer
+## Approach
+* 3 Classes: Player, Game & Computer
+* 3 Views: /index, /play & /result
+* Aimed to use the MVC pattern to improve the maintainability of the app
+* Encapsulated Computer in a separate class to take advantage of duck-typing in the Game class - the game does not need to know if it's comparing two players or one player and a computer!
```
soph-g: Great!

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def result
+    winner(@player_1, @player_2)
+  end
+
+  private
+
+  def winner(player_1, player_2)
+    if RULES.include?([player_1.shape, player_2.shape])
```
soph-g: It's not clear how the rules work for the game. For example, if I wanted to add in new weapons it's not obvious what changes I would have to make to the RULES to make the game work. Can you use a different data structure to make the rules easier to understand, without adding comments?

rps_web.rb
```diff
@@ -0,0 +1,41 @@
...
+    @game = Game.instance
+    @player_1 = Player.instance
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post '/shape' do
+    @game.player_1.shape = params[:shape]
```
soph-g: In this line you're reaching quite far into the game instance, having to chain a method onto Game then the Player. Instead could you call a method on Game that will delegate to the Player?

rps_web.rb
```diff
@@ -0,0 +1,41 @@
...
+    redirect('/result')
+  end
+
+  get '/result' do
+    @game.player_2.shape = Computer.new.bot_shape
```
soph-g: As above, can you instead call a method on Game that delegates to the player?
Also, why not set the Computer choice on initialize, rather than having to chain the method call?

spec/features/result_spec.rb
```diff
@@ -0,0 +1,76 @@
+feature 'Winner declared' do
```
soph-g: Good coverage in your tests but they are hard to read due to the repetition. Can you find a way to make these tests easier to read by grouping them?

spec/game_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'game'
+
+describe Game do
+  let(:player_1) { double(:player_1, name: "Cinderella") }
+  let(:player_2) { double(:player_2, name: "Bot") }
+  subject(:game) { Game.new(player_1, player_2) }
```
soph-g: Good use of dependency injection

spec/game_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  it "stores player names" do
+    expect(subject.player_1.name).to eq "Cinderella"
+    expect(subject.player_2.name).to eq "Bot"
+  end
+
+  it "stores player moves" do
+    allow(player_1).to receive(:shape).and_return("Rock")
+    allow(player_2).to receive(:shape).and_return("Scissors")
+    expect(subject.player_1.shape).to eq "Rock"
```
soph-g: Not sure these tests are meaningful - you're stubbing the behaviour under test! Also, as noted previously, it would be better to use delegation rather than method chaining here.

PR #1066, opened 2018-10-21
Rock, Paper, Scissors, Lizard, Spock

Caitlin-cooling:


README.md
```diff
@@ -1,86 +1,27 @@
-# RPS Challenge
+# Rock, Paper, Scissors
+This is a game that let's you play a game of rock paper scissors against the game.
 
-Instructions
--------
+### How do I Use it?
+Clone this repo. Run the ruby file `app.rb`, this will start your server. Then in your browser go to localhost:4567.
```
soph-g: Try to give exact commands, and don't forget to include installing gems and running tests in the instructions.

README.md
```diff
@@ -1,86 +1,27 @@
...
-I would like to register my name before playing an online game
+### Running the tests
+The tests for this program use rspec and capybara, so simply run rspec from within this directory.
 
-As a marketeer
-So that I can enjoy myself away from the daily grind
-I would like to be able to play rock/paper/scissors
-```
+### Known Issues
+* Line 6 in `winner.rb` is too long. I plan to use a hash instead of an array and access the values of that the see player 2 move
```
soph-g: These are all good changes to identify. If you can spot opportunities for improvement it's good to actually make those changes before submitting for a coach review.

app.rb
```diff
@@ -0,0 +1,84 @@
...
+require_relative './lib/winner'
+require_relative './lib/computer'
+
+class RPS < Sinatra::Base
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/two_player' do
```
soph-g: Not passing any data here, should this be a post route?

app.rb
```diff
@@ -0,0 +1,84 @@
...
+    erb(:two_player)
+  end
+
+  post '/one_player' do
```
soph-g: Not passing any data, should this be a post route?

app.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  get '/multiplayer_game' do
+    @players = Players.instance
+    @name1 = @players.name1
+    @name2 = @players.name2
+    erb(:multiplayer_game)
+  end
+
+  post '/multiplayer_game2_names' do
```
soph-g: Not sure what this route does, naming is not clear.

app.rb
```diff
@@ -0,0 +1,84 @@
...
+  post '/game/determine' do
+    @players = Players.instance
+    @players.move1 = params["move1"] if @players.move1.nil?
+    @players.move2 = params["move2"] if @players.move2.nil?
+    redirect '/game/result'
+  end
+
+  get '/game/result' do
+    @players = Players.instance
+    @name1 = @players.name1
```
soph-g: Lots of variables being set here - is there a way to avoid this and use a single variable that the view can access the moves and player names on?

capybara-201810200910053459868959.html
```diff
@@ -0,0 +1,6366 @@
+<!DOCTYPE html>
```
soph-g: Don't think this file should be commited - looks like it's been generated by capybara.

lib/winner.rb
```diff
@@ -0,0 +1,32 @@
+require_relative "messages"
+
+class Winner
+  include Messages
+
+  WIN_COMBOS = [[:rock, :scissors], [:paper, :rock], [:scissors, :paper], [:scissors, :lizard], [:rock, :lizard], [:lizard, :paper], [:paper, :spock], [:lizard, :spock], [:spock, :rock], [:spock, :scissors]]
```
soph-g: Find this structure unclear - what do the nested arrays represent? Difficult to work out what the rules are using this information.

lib/winner.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize(players)
+    @players = players
+    convert_to_sym
+  end
+
+  def determine
+    WIN_COMBOS.each do |combo|
```
soph-g: Could you simplify this to avoid using if statements and iteration?

lib/winner.rb
```diff
@@ -0,0 +1,32 @@
...
+        return player1_won
+      end
+    end
+    return draw if @player1_move == @player2_move
+    return player2_won
+  end
+
+  private
+
+  def convert_to_sym
```
soph-g: Could this method be simplified to remove repetition?

spec/players_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'players'
+
+describe Players do
+
+  let(:players) { Players.new("Caitlin", "James") }
+
+  it 'name1 returns the name of the first player' do
```
soph-g: Recommend grouping unit tests by the method under test using 'describe' blocks.

lib/players.rb
```diff
@@ -0,0 +1,18 @@
+class Players
```
soph-g: Feels like this class has multiple responsibilities. It's holding the names and moves for both player 1 and player 2. Is there a Player object that could be extracted here?

lib/computer.rb
```diff
@@ -0,0 +1,9 @@
+class Computer
+
+  attr_reader :move2, :name2
+
+  def initialize(options)
+    @move2 = options.sample
```
soph-g: Why move2 and name2?

README.md
Caitlin-cooling: Updated how do I use. Running tests is in a separate section below.

app.rb
Caitlin-cooling: Changed

capybara-201810200910053459868959.html
Caitlin-cooling: Deleted

app.rb
Caitlin-cooling: Have since learnt that you should post to the current page then redirect, have implemented this.

PR #1084, opened 2018-10-21
RPS Challenge

rsmith88:


app.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  post '/single_player_name' do
+    player_1 = Player.new(params[:player_1_name])
+    player_2 = Computer.new
+    @game = Game.create(player_1, player_2)
+    redirect '/single_player_enter_move'
+  end
+
+  post '/two_player_name' do
```
soph-g: Could you reduce repetition and handle single and mulitplayer in a single route?

app.rb
```diff
@@ -0,0 +1,73 @@
...
+  post '/two_player_move1' do
+    @game.player1.move = params[:choice]
+    redirect '/two_player_enter_move2'
+  end
+
+  get '/two_player_enter_move2' do
+    erb :tp_enter_move2
+  end
+
+  post '/two_player_move2' do
```
soph-g: Again, can you reduce repetition in the routes? This is very similar the to single player route?

lib/computer.rb
```diff
@@ -0,0 +1,16 @@
...
+  WEAPONS = ["Rock", "Paper", "Scissors", "Lizard", "Spock"]
+
+  def initialize
+    @name = "The Computer"
+    @move = nil
+  end
+
+  attr_accessor :name, :move
+
+  def random_move
```
soph-g: Does the Controller need to call the random_move method on the computer or could the random move be set on initialize?

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def self.create(player1, player2)
+    @game = Game.new(player1, player2)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def initialize(player1, player2)
```
soph-g: Could you use dependency injection to set player2 to be the Computer by default?

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+    @player1 = player1
+    @player2 = player2
+    @winner = "unassigned"
+  end
+
+  attr_reader :player1, :player2, :winner
+
+  def winning_combo
+    @win_combo = {
```
soph-g: Good structure, however the naming could be clearer. I think the key is the winning weapon, with the values being the things it beats, but I'm not sure!

soph-g: Also, is the variable assignment necessary? What if the method returned the hash?

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+      "Rock" => ["Scissors", "Lizard"],
+      "Scissors" => ["Paper", "Lizard"],
+      "Paper" => ["Rock", "Spock"],
+      "Lizard" => ["Paper", "Spock"],
+      "Spock" => ["Rock", "Scissors"]
+      }
+  end
+
+  def evaluate(player1, player2)
+    @winner = player1.name if player1_wins?
```
soph-g: Could you reduce repetition in the winner determination?

spec/features/infrastructure_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    start_game_sp
+    click_button "Lizard"
+    expect(page).to have_content("You played Lizard.")
+  end
+  scenario 'it allows user to choose Spock' do
+    start_game_sp
+    click_button "Spock"
+    expect(page).to have_content("You played Spock.")
+  end
+  scenario 'it outputs the computers move & winner' do
```
soph-g: Try to check that the computer can win/lose and the user can win/lose, and there can be a draw. Your tests currently don't check that the correct winner/loser/draw result is presented to the user.

spec/game_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'game'
+
+describe Game do
+  describe "#initialize" do
+    it 'initializes a game with one player' do
+      player1 = Player.new("Ray")
```
soph-g: Test for game is dependent on the Player here - remove the dependency by using a double for Player.

PR #1094, opened 2018-10-22
Rock Paper Scissors 

Edy1988:


capybara-201810191540552379363805.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
```
soph-g: Be careful not to commit test output files!

app.rb
```diff
@@ -9,17 +9,50 @@ class RPSapp < Sinatra::Base
...
+    elsif both_played
+      display_result
     else
-      @current_player_name = @player_2_name
+      store_player_1_choice
+      request_player_2_choice
     end
+  end
+
+  def request_player_1_choice
```
soph-g: Would expect to see methods like these in a Class, rather than in the app.rb. Should be aiming to move logic out of the Controller and into a Model (Class) as much as possible.

app.rb
```diff
@@ -0,0 +1,76 @@
+require 'sinatra/base'
+
+class RPSapp < Sinatra::Base
+
+  get '/' do
+    erb :index
+  end
+
+  post '/player_1_turn' do
+    @player_1_name = params[:player_1_name]
```
soph-g: Rather than relying on posting params for every request, could you use a Game model to store the player names?

app.rb
```diff
@@ -0,0 +1,76 @@
...
+  def draw?
+    @player_1_move == @player_2_move
+  end
+
+  def render_draw_for(move)
+    @draw_move = move
+    erb :draw
+  end
+
+  def player_1_won?
```
soph-g: Would expect to see logic like this in a Class (Model), rather than in the Controller. Should aim to have as little logic as possible in the views and controller.

game.rb
```diff
@@ -0,0 +1,18 @@
+class RockPaperScissorsGame
```
soph-g: Good start on a Game model - can you integrate this with the Controller?

PR #1103, opened 2018-10-22
Final pull request - please review

PreetiSekhon2:


app.rb
```diff
@@ -0,0 +1,64 @@
...
+    @noofplayers = params[:noofplayers]
+    erb :playernames
+  end
+
+  post '/players' do
+    @game = Game.create.assign_players([params[:player1],params[:player2]])
+    redirect "/play"
+  end
+
+  get '/play' do
```
samjones1001: There are some pretty big routes in this controller 🙂 

Generally speaking if a route contains more than a couple of lines of code it indicates one of two things: 
- some of the logic in the route could be pushed back to an appropriate model
- the route is doing too much and has become over complicated.

app.rb
```diff
@@ -0,0 +1,64 @@
...
+  rescue NoWinnerError => err
+    @err = err
+    erb :playrps
+  end
+
+  post '/play' do
+    if params[:submit] == "submit"
+      @game.start_round
+      @game.player_plays(@game.players[0],params[:move1])
+      @game.players[1].get_name == "Computer" ? @game.player_plays(@game.players[1],["rock","paper","scissors"].sample) : @game.player_plays(@game.players[1],params[:move2])
```
samjones1001: This feels like a good example of something that could benefit from being pushed back to the model layer.  You have some reasonably complex business logic, figuring out what should be passed into the player_plays method - could this just be put into a method in the game class.

lib/judge.rb
```diff
@@ -0,0 +1,54 @@
...
+class Judge
+  attr_reader :moves
+  def initialize()
+    @moves = Array.new(0)
+    File.open(File.dirname(__FILE__) + '/moves.txt').each {|line|
+      @moves.push({(line.strip).to_sym => 0})
+    }
+  end
+
+  def decide_round_winner(players)
```
samjones1001: This is a huge method!

General consensus in the ruby community is that if a method is longer than 5 lines then it's too long.
https://robots.thoughtbot.com/sandi-metz-rules-for-developers

A useful exercise if you find yourself with a very long method might be to go through the method and make a list of everything that it does - I'm almost certain that you will find that it has multiple responsibilities. 

Once you've determined the responsibilities, you should be able to extract some of them into private methods.

PR #1116, opened 2018-11-18
RPS challenge

AgaDufrat:


app.rb
```diff
@@ -0,0 +1,97 @@
...
+  # single player methods:
+
+  get '/singleplayer' do
+    @version = session[:version]
+    erb(:singleplayer)
+  end
+
+  post '/name' do
+    player1 = Player.new(params[:name])
+    @game = Game.create(player1, \
```
alicelieutier: You don't actually need that `\` here, as you have a return inside parentheses. This could be written:
```
@game = Game.create(
  player1,
  Player.new("Computer", true),
  @version 
)
```

AgaDufrat: ah.. of course! Thank you for pointing that out

PR #1128, opened 2018-11-19
challenge completed

EllyChanx:
This repo only has 1 time commit since I performed the task without forking this original repo.
For previous commit and change record, please see:
https://github.com/EllyChanx/RPS_Challenge


app.rb
```diff
@@ -0,0 +1,49 @@
...
+    erb :play2
+  end
+
+  post "/choice" do
+    session[:player_2_action] = @game_1.player_2_action(params[:player_2_action])
+    redirect "/result"
+  end
+
+  get "/result" do
+    @xx = session[:player_1_action]
```
ShinyVerse: xx and yy aren't very descriptive variable names. Best to go for transparency for future devs reading your code than an easy fix. In the Rps challenge this wasn't such a big deal, but best to break the habit before it forms.

lib/game_1.rb
```diff
@@ -0,0 +1,49 @@
...
+  def player_1
+    @players[0]
+  end
+
+  def player_2
+    @players[1]
+  end
+
+  def player_1_action(choice)
+    @result = []
```
ShinyVerse: I felt that @result belonged in the initialize method.

This would make player_1_action(choice) and player_2_action(choice) pretty much identical, they could be converted into a single method that could be used for both.

spec/features/name_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  scenario 'submitting names' do
+    visit('/')
+    fill_in :player_1_name, with: 'Dave'
+    fill_in :player_2_name, with: 'Mittens'
+    click_button 'Go!'
+    click_button 'Confirm!'
+    expect(page).to have_content 'Mittens'
+  end
+
```
ShinyVerse: Use of the web_helper function sign_in_and_play could be applied here to dry up the code.

lib/game_1.rb
```diff
@@ -0,0 +1,49 @@
...
+    @result << choice
+    choice
+  end
+
+  def check_tie
+    return true if (@result[0] == @result[1])
+    false
+  end
+
+  def result
```
ShinyVerse: Result function was easy to read for where it's at now, but could probably not be expanded upon easily.

PR #1136, opened 2018-11-19
My submission

willmcb:
I didn't really get on to adding the extra functionality, but I was spending time trying to make the code as clean as I could. Would love some feedback. Thanks.


lib/game.rb
```diff
@@ -0,0 +1,34 @@
+class Game
+  attr_reader :player1, :player2
+  def initialize(player1, player2)
+    @player1 = player1
+    @player2 = player2
+  end
+
+  def result
+    # player 1 lookup left, player two right.
+    rules = {
```
alicelieutier: make line shorter like this:
```
rules = {
      rock: {
          rock: draw,
          paper: player_two_wins,
          scissors: player_one_wins
      },
      paper: { 
etc.
```

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,32 @@
+require 'sinatra/base'
```
alicelieutier: This controller is super clean, that is really nice to see.

PR #1141, opened 2018-12-16
Submission

matharotheelf:


app.rb
```diff
@@ -0,0 +1,38 @@
...
+  @player = Player.instance
+  @name = @player.name
+  erb(:game)
+  end
+
+  post '/choice' do
+  @player = Player.instance
+  @player.randomchoice
+  @player.choose(params[:choice])
+  @player = Player.create(@player.name, @player.choice, @player.computer_choice)
```
joaoag: Would you consider having a separate class for Computer?

lib/player.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def choose(choice)
+    @choice = choice
+  end
+
+  def randomchoice
+    @computer_choice = ["rock", "paper", "scissors"].sample
+  end
+
+  def game_result
```
joaoag: Would you consider breaking down game_result into a series of smaller methods? This might be more adherent to SRP

lib/player.rb
```diff
@@ -0,0 +1,49 @@
...
+    if computer_choice == choice
+      "draw"
+    elsif lose1 || lose2 || lose3
+      "lose"
+    elsif win1 || win2 || win3
+      "win"
+    end
+  end
+
+  def self.create(name, choice = nil, computer_choice = nil)
```
joaoag: Nice work getting round the global variable. I sadly did not do this

spec/player_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "player"
```
joaoag: Good work making feature tests going through whole process and also seperate unit testing the Player class

PR #1152, opened 2018-12-16
:(

tomlovesgithub:


lib/result.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def playerbeatscomp
+    @game_result = ['You Win']
+  end
+
+  def draw
+    @game_result
+  end
+
+end
```
jaywayawyaj: Amazing use of emoji's

lib/result.rb
```diff
@@ -0,0 +1,43 @@
...
+    elsif @weapon == '✌️' && @comp_hand == '🤚'
+      playerbeatscomp
+    elsif @weapon == '✌️' && @comp_hand == '👊'
+      compbeatsplayer
+    elsif @weapon == '🤚' && @comp_hand == '👊'
+      playerbeatscomp
+    elsif @weapon == '🤚' && @comp_hand == '✌️'
+      compbeatsplayer
+    end
+  end
```
jaywayawyaj: Could this be slimmed down? Consider using values in a hash and comparing that way - saves on the mega if/elsif space

app.rb
```diff
@@ -0,0 +1,44 @@
...
+    session[:name_colour] = params[:name_colour]
+    session[:weapon] = params[:weapon]
+    @name_colour = session[:name_colour]
+    @player_name = session[:player_name]
+    @weapon = session[:weapon]
+    @result = session[:result]
+    @result_cp = @result.comp_hand
+    erb(:winner)
+  end
+
```
jaywayawyaj: Could some of this be extracted into a player class (for instance) where you store attributes? Feels like there's a lot going on in the controller layer

lib/result.rb
jaywayawyaj: Could this class be split into more than one? It feels like it's got multiple responsibilities

lib/result.rb
```diff
@@ -0,0 +1,43 @@
...
+    elsif @weapon == '👊' && @comp_hand == '🤚'
+      compbeatsplayer
+    elsif @weapon == '✌️' && @comp_hand == '🤚'
+      playerbeatscomp
+    elsif @weapon == '✌️' && @comp_hand == '👊'
+      compbeatsplayer
+    elsif @weapon == '🤚' && @comp_hand == '👊'
+      playerbeatscomp
+    elsif @weapon == '🤚' && @comp_hand == '✌️'
+      compbeatsplayer
```
stonefarmer9: would using a simple win, lose or draw make writing your code easier later on?

lib/result.rb
stonefarmer9: This class does a lot, what is its main responsibility and how would you refactor it?

views/weapon_select.erb
```diff
@@ -0,0 +1,18 @@
...
+</center>
+
+<center>
+
+
+  <form action='/result' method='post'>
+
+    <input name='weapon' type='Submit' value='👊'>
+    <input name='weapon' type='Submit' value='🤚'>
+    <input name='weapon' type='Submit' value='✌️'>
```
stonefarmer9: really cool use of emoji that you have to show me! and love the use of the the individual buttons rather thasn type input like i Have

lib/result.rb
```diff
@@ -0,0 +1,43 @@
...
+    @comp_hand = hands_arr.sample
+    @weapon = weapon
+    @game_result = ['Draw']
+    eval
+  end
+
+  def eval
+    if @weapon == @comp_hand
+      draw
+    elsif @weapon == '👊' && @comp_hand == '✌️'
```
joaoag: sweet emojis

lib/result.rb
```diff
@@ -0,0 +1,43 @@
...
+      playerbeatscomp
+    elsif @weapon == '👊' && @comp_hand == '🤚'
+      compbeatsplayer
+    elsif @weapon == '✌️' && @comp_hand == '🤚'
```
joaoag: Because you've outlined all the playerbeatscomp scenarios and the draw scenario, the compbeatsplayer outcome could be covered by a final 'else' clause at the end. So you'd only need five 'if' statements: three for player win, one for player draw and one for player lose. Hit me up if this doesn't make sense, sorry if explained poorly

joaoag:
Also forgot to mention excellent work with the visuals. Really love the colour selection option, the emoji displays too. 💯

PR #1157, opened 2018-12-17
Wkend 3 - rps challenge

OSSahota:
Not complete - the game doesn't meet the functionality of declaring a winner.
The code also needs tidying up and refactoring in many places. The views are very basic and need to present the output in a nicer format when rendered.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -12,14 +12,22 @@ class RPS < Sinatra::Base
   post '/name' do
     # Extrating values in param hash to session hash.
     session[:player_1_name] = params[:player_1_name]
+    p params
     # Redirect to '/play' route transferring responsibility of rendering.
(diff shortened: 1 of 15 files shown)


app.rb
```diff
@@ -0,0 +1,41 @@
...
+    @computer_choice_made = Computer.new.make_random_choice
+    p @computer_choice_made
+    erb(:choice_made)
+  end
+
+  # run! starts a server.
+  # Only start a server if the file has been executed directly with __FILE__
+  # being the current file and $0 being the executed file.
+  run! if app_file == $0
+end
```
CKKH: Nice work Onkar! Could it be worth moving the player attributes to its own model class like computer?

PR #1184, opened 2019-01-27
Completed the rock, paper scissors week 3 weekend challenge with basic functionalities.

JPProgrammingTime:
**Goals** 

- I can TDD anything
- I can program fluently
- I can refactor anything
- I can debug anything
- I can model anything
- I write code that is easy to change

**Evidence**

I can TDD anything and model anything

Every time I added a completely new functionality or feature for the user stories of the project, I stuck to TDD by writing the test first. This coincided with the modelling goal as I was forced to draw out class diagrams (my favourite) and process models to first get an idea of how my web app should operate, before I could write a test. The only situations where I deviated from strict TDD was when I realised I would have to slightly modify previous functions to make my current and new core tests to pass. 

As evidence of my testing, I have unit specs for my Game, Player and Computer classes with a feature spec for the web app itself. I achieved 100% coverage but there is one particular feature test that I could not solve, which I think may have increased the quality of my tests significantly. I tried to create a feature test that would stub the computer's move (stored in an instance variable in the controller file) to scissors so that when the user chooses rock twice it would take them to the winners page. This test would effectively show that the core gameplay works. I got close but unfortunately I could not figure out how to stub the controller instance variable for a particular post route after having tried various Capybara/RSpec statements.

I can debug anything 

This being the first time I have developed a program with front-end features, most of, if not all, of the test failure statements were quite new to me. However, I found that because the structure of my tests, due to having modelled the necessary parts before writing the tests, were logically sound (at least in terms of message steps) the errors were mostly trivial and the solution could easily be found on stack overflow or without even the need to look at an external resource. 

Unfortunately, as mentioned above, I could not create a fully interactive, from sign in to end of game feature test, but I could tell that the missing factor is related to a particular controller instance variable. I used the save_and_open_page phrase to view the output page and identify what was not being controlled properly in the test.

I can program fluently,  I write code that is easy to change and I can refactor anything

This weekend I realised just how comfortable I have become, not only with Ruby, but also with program design and understanding/researching on how to use gems. Whenever I needed to specify more statements to use certain features of gems I found myself being able to read support documents much more fluently than I did before. Examples are including the Capybara::DSL module for using phrases like 'visit' for test-support classes like Web Helpers and using Sinatra keywords like set :public_folder and :view to set the paths of my public folders where most of my static files are like css stylesheets, images and audio files) and the view folder (where my html files are).

I found making modifications and refactoring much easier now, as I seem to stick to the single responsibility principle and don't repeat yourself, much more strictly than before. Some of my controller route methods need to be further refactored but due to me having assigned core game responsibilities to the Game, Player and Computer classes as such:

- The Game's responsibility was to assign a point to the winner of the round, find the winner, and also check to see if the
game should end as the player or computer has won two rounds (best of three).

- Both Player and Computer would be responsible for choose their hand: rock, paper or scissors.

- The app/controller would pass the class methods and attributes through posts and redirect to relevant pages. I made the controller responsible for starting a new game through the homepage sign in but I was unsure whether to assign this responsibility to the Game class.

Whenever I needed to make modifications, I ended up either just adding/deleting a method to a single class or add/remove a line to a method within a particular class, rather than having to restructure each component of the app every time there was a change. I also barely had to touch my controller methods with the exception of adding user story features.


Gemfile
```diff
@@ -11,4 +11,7 @@ group :test do
   gem 'rubocop', '0.56.0'
   gem 'simplecov', require: false
   gem 'simplecov-console', require: false
+  gem 'selenium-webdriver'
+  gem 'pry'
```
vaithschmitz: Why are you requiring pry and readline in your file? You should be able to use them both to interact with your app without having them in your Gemfile.

PR #1193, opened 2019-01-28
Rock Paper Scissors - week 3 challenge

dippedincode:
The challenge is to create a web-app of the game Rock Paper Scissors. The first version is a one-player against the computer. The second version is a two-player, human vs human.
A goal was to see if I could build a web-app using TDD.
The program needed to be easy to read and fulfil its requirements in an elegant way.
I gave myself another goal of making it look good with some images.


app.rb
```diff
@@ -0,0 +1,78 @@
+require 'sinatra/base'
+require 'shotgun'
+
+class RockPaperScissors < Sinatra::Base
+  enable :sessions
```
AdrienFabre: Using sessions can help to reduce the number of params.

PR #1197, opened 2019-01-28
Failing tests - completed solution

JakePEGG:


lib/computer.rb
```diff
@@ -1,7 +1,7 @@
 class Computer
 
-attr_reader :choice 
-  def initialize(choice)
+attr_reader :choice
+  def initialize
```
brookewoolley: Your test is expecting to pass an argument to computer so I think if you update your test to use doubles you should be able to get it to pass.

lib/player.rb
```diff
@@ -1,8 +1,8 @@
 class Player
-  
-attr_reader :choice
+  attr_reader :choice
+
   def initialize(choice)
-    @choice = choice.to_sym
+    @choice = choice.to_s
```
brookewoolley: Update your test to accept a string ("rock") and it should pass.

PR #1205, opened 2019-01-28
Pull Request - RPS Challenge

ebk45:
I can TDD anything:

Evidence shown through commits - they show a clear process of testing, writing passing code and refactoring. It's not a perfect process but we're getting there.


app.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  get '/game' do
+    @player_name = session[:player_name]
+    erb(:game)
+  end
+
+  post '/cpu_choice' do
+    session[:player_choice] = params[:player_choice]
+    redirect '/result'
```
punchcafe: why does this redirect to a post?

lib/cpu.rb
```diff
@@ -0,0 +1,9 @@
+class CPU
+
+  attr_reader :option
+
+  def initialize
+    @option = ['🤚', '✌️', '✊'].sample
```
punchcafe: loool this is great

lib/result.rb
```diff
@@ -0,0 +1,28 @@
...
+      'You win!'
+    elsif draw?
+      "It's a draw!"
+    else
+      'You lost!'
+    end
+  end
+
+  def win?
+    @player_choice == "✊" && @cpu_choice == "✌️" ||
```
punchcafe: this is a cool way of doing it, but I'm not sure how good practice it is to have such long conditionals

spec/features/testing_infrastructure_spec.rb
```diff
@@ -0,0 +1,8 @@
+# require './app'
```
punchcafe: if you're not using it, delete it maybe

spec/units/cpu_spec.rb
```diff
@@ -0,0 +1,18 @@
+require './lib/cpu'
+
+RSpec.describe 'CPU:' do
```
punchcafe: this is interesting. I did the same thing, since it's such a basic thing. I want to know if this kind of testing state is ok.

PR #1207, opened 2019-02-23
Clare Pinder: rock, paper scissors, 1 player.

clarepins:


mmar8:
Code is easy to read. Classes and file names are well named and thought through. 
In my opinion, it would be great to commit more often.

PR #1212, opened 2019-02-24
Jules rps-challenge

Jimjule:


spec/game_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "game"
+
+describe Game do
+  it 'allows the player to choose scissors' do
+    game = Game.new('Spidey')
+    game.play('Scissors')
```
simon-norman: Could be good to have an expect statement here to make it clear what you are testing. Might be good to test other scenarios - what happens eg if user selects rock and computer selects paper?

views/end.erb
```diff
@@ -0,0 +1,28 @@
...
+
+    body {
+      display: flex;
+      text-color: white;
+      flex-direction: column;
+      text-align: center;
+      align-items: center;
+      background-image: url('https://inhdw.com/wallpaper/8/skull-and-bones-pirate-battle-814.jpg');
+      background-color: #333fff;
+    }
```
simon-norman: Nice! Possible addition would be to break up these CSS rules to reduce duplication - e.g. you could have h1, body { display: flex etc. }

views/index.erb
```diff
@@ -0,0 +1,29 @@
+<head>
+  <style>
+    h1 {
+      display: flex;
+      text-align: center;
```
simon-norman: Same CSS across multiple files, which is totally fair as we haven't learned how to use CSS files yet :)

app.rb
```diff
@@ -0,0 +1,39 @@
...
+    redirect('/game')
+  end
+
+  get '/game' do
+    @player_one = @game.player_one
+    erb(:game)
+  end
+
+  post '/end' do
+    @choose = @game.play(params[:choose])
```
simon-norman: I don't think this choose variable is doing anything and can be removed

views/end.erb
Jimjule: Thanks, implemented!

views/index.erb
Jimjule: Quite right!

PR #1216, opened 2019-02-24
Valiant solution of valour.

Schofield88:
Welcome, warrior. We've been expecting you.

Men search their entire lives for even the merest hint of purpose. To die knowing one has missed the chance to meet their potential head on at the charge wakes many from the blackest depths of restless sleep in rivers of their own sweat, their dream-leaked screams still echoing from the walls. 

But you may rest now, warrior. What you have set out to seek, what you feared you may never find has found you. Rest now, knowing that you will finally plumb the depths of your courage and face the flayed centre of your very existence. You will stare your soul in the face and it will stare back at you. Don't blink, for you have already been judged.

Fear is a reflex, courage is a choice.

Welcome to Rock, Paper, Scissors.

### DIFF
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -1,0 +1,17 @@
+class Game
+
+  def initialize(player_move, ai_move)
+    @player_move = player_move
+    @ai_move = ai_move
+  end
+
+  def winner
+    win = "You win at this skillful game of chance"
+    lose = "HAL wins. Another time, McCloud"
+    draw = "This one's a draw"
+    @player_move == "Rock" && @ai_move == "Scissors" ? win : @player_move == "Rock" && @ai_move == "Paper" ? lose : draw
+    @player_move == "Paper" && @ai_move == "Rock" ? win : @player_move == "Paper" && @ai_move == "Scissors" ? lose : draw
+    @player_move == "Scissors" && @ai_move == "Paper" ? win : @player_move == "Scissors" && @ai_move == "Rock" ? lose : draw
+  end
+
+end
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -9,9 +9,14 @@ class Game
     win = "You win at this skillful game of chance"
     lose = "HAL wins. Another time, McCloud"
     draw = "This one's a draw"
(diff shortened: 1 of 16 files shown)


views/fight.erb
```diff
@@ -0,0 +1,66 @@
+<!DOCTYPE html>
+<html>
+<head>
+<style>
+  body {
```
matthewwhitaker: This can be abstracted to a CSS stylesheet to reduce duplication and make the erb files a bit easier to read - happy to talk you through this in more detail if you'd like

spec/features/infrastructure_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    find_button("Scissors")
+  end
+
+  scenario "Page shows the result of the game" do
+    set_up
+    click_button("Rock")
+    expect(page).to have_content 'The result of the round'
+  end
+
+  scenario "Page tells you what everyone chose" do
```
matthewwhitaker: This test doesn't test for the presence of a randomly assigned variable, so if there was a problem returning something or printing to the page it would give a false positive. You can use srand(some_number) to fix the randomness in tests - happy to talk through this if you'd like.

views/result.erb
```diff
@@ -0,0 +1,71 @@
...
+<BODY>
+  <center>
+  <h1> The result of the round </h1>
+  <h2><%=@name%> chose <%=@choice%></h2>
+  <h2>HAL chose <%=@ai_choice%>
+  <h3><%=@result%></h3>
+  <form action="/fight">
+    <button type="submit" name="choice" value="name" class="button button1">Play again?</button>
+  </form>
+  <h3><%=@courage%></h3>
```
matthewwhitaker: This line on the application is formatted in an array with the quote and a nil as elements. No functional problems with it, but potentially not the format you were intending.

lib/game.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize(player_move, ai_move)
+    @player_move = player_move
+    @ai_move = ai_move
+  end
+
+  def winner
+    win = "You win at this skillful game of chance"
+    lose = "HAL wins. Another time, McCloud"
+    draw = "This one's a draw"
+    case @player_move
```
matthewwhitaker: This is much cleaner than my own implementation, thanks for the good example!

PR #1222, opened 2019-02-25
Incomplete RPS

kate102:
The current state of the program is that there is an initial screen that enables you to enter a name. Once this has been submitted the following screen welcomes the player by name and offers a choice of 3 weapons. The user is able to choose a weapon by clicking on the relevant button.

Outstanding Issues:
I am confused by the name conventions and flow of control. 
I have a GET method which I feel should be a POST but doesn't work as a post. 
This appears to be sequential processing and the GET and POST's are methods.
When I attempt to access the param's the process fails.
I would welcome help on the above and where the processing should go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,3 +32,20 @@ Paper beats Rock
 8. Generate the random computer option
 9. Write a test to determine a winner
 10. Write the code to determine a winner
+
+
+#########
+
+I have decided to tackle this another way.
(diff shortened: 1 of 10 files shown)


kate102:
Oh and I forgot to say that there are no tests.
I was unable to get off the ground because I was unable to write tests for a coding framework I fully didn't understand. My strategy changed, after a day of inertia, so that some good could come from the exercise.

PR #1223, opened 2019-02-25
RPS - Single and Multiplayer

SurekaStoeckigt:
100 % test coverage.
Able to play against computer or another player.
All tests passing.


views/draw.erb
```diff
@@ -0,0 +1,2 @@
+<h2>In this round of RPS: <%= @name1 %> vs. <%= @name2 %>, the result is a DRAW!</h2>
+<img class="front-picture" src="http://www.pmslweb.com/the-blog/wp-content/uploads/2017/06/13-playing-rock-paper-scissors-during-stone-age-funny-comic.jpg">
```
dafuloth: imo, I would advise against hotlinking unless a site specifically permits it - when reuse is permitted, it is usually better to have a local copy in the repository/site and give proper credit. A practical reason is that it is an external dependency. If the source breaks or is down, it affects your app too.

dafuloth:
General comment if we care about rubocop at all: it mentions various things, but mainly "inconsistent indentation"

I should've run rubocop on mine...

PR #1224, opened 2019-02-25
rps challenge

dory1318:


app.rb
```diff
@@ -0,0 +1,20 @@
+require 'bundler'
+Bundler.require
+
+class RPS < Sinatra::Base
```
steve-f-kahn: RPS is a class not a constant so should be called 'Rps'

app.rb
```diff
@@ -0,0 +1,20 @@
...
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    @player_1 = params[:player_1]
+    redirect to('/select')
```
steve-f-kahn: The 'to' is not required

spec/features/select_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'After log in' do
+  scenario 'player can select a weapon' do
+    visit('/')
+    fill_in("name", with: "Dora")
+    click_button("Start")
+    expect(page).to have_content "Choose your weapon"
```
steve-f-kahn: This test description is different from what is does. It says a player can select a weapon but it does not select it. A more accurate description might be "player can start the game". To test for selecting weapons you could have three separate tests for each weapon button that you will click.

views/index.erb
```diff
@@ -0,0 +1,21 @@
+<html>
+<head>
+<title>Log in</title>
+</head>
+<body style="background-color:orange; text-align: center;">
+
+<h1 style="font-size:50px; border:2px solid Tomato; ">Welcome Dear Visitor! Let's play a game</h1>
```
steve-f-kahn: I like the styling of the webpage.

views/index.erb
```diff
@@ -0,0 +1,21 @@
...
+
+<form action="/name" method="post">
+
+<p style="font-size:30px; ">Please log in to start the game</p>
+<br>
+<input type="text" name="name" placeholder="nickname" align="left">
+
```
steve-f-kahn: I haven't seen placeholder before does that pre-fill the box they type in?

lib/game.rb
```diff
@@ -0,0 +1,51 @@
+require '/Users/Dori/Projects/homework/week3/rps-challenge/lib/person.rb'
+require '/Users/Dori/Projects/homework/week3/rps-challenge/lib/computer.rb'
+class Rps
+attr_reader :player_input, :computer_input
```
steve-f-kahn: The app class is called RPS and this game logic class is called Rps. This is very confusing.

views/index.erb
dory1318: Placeholder specifies a short hint that describes the expected value of an input field

lib/game.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def initialize(player_input = Person.new(:choice), computer_input= Computer.new)
+    @player_input = player_input
+    @computer_input = computer_input
+  end
+
+def compare_choices(player_input, computer_input)
```
steve-f-kahn: This method is way too long. You could move the puts statements into a new method if not a new class. Puts its not helpful for telling the player what happened as they cannot see your terminal. Perhaps they should just be strings that when returned to the main app are then put into a view and told to the player.

lib/person.rb
```diff
@@ -0,0 +1,8 @@
+class Person
+attr_reader :choice
+
```
steve-f-kahn: Should there name be stored here as well?

PR #1226, opened 2019-02-25
Maria Vanaale  RPS

mmar8:


mmar8:
Half finished

spec/rps_spec.rb
```diff
@@ -0,0 +1,24 @@
+require './lib/rps'
```
clarepins: The rubric says we should test all possible outcomes - rock, paper and scissors. The srand function seemed to work srand(0) = the first element of the array, srand(1) is the second element of the array, but weirdly srand(3) (not 2) was the third element of the array.

lib/rps.rb
```diff
@@ -0,0 +1,15 @@
+class Rock_Paper_Scissors
+  
+  attr_reader :player_name 
+  
+  Rps = ["rock", "paper", "scissors"]
+
+  def initialize(player_name)
+    @player_name = Game.new(player_name)
```
clarepins: I don't think you need this Game.new. I don't think you've defined Game class anywhere and I don't know what it would do rather than work out the computer's move and work out who won, which I would think should be in the Rock_Paper_Scissors class.

PR #1231, opened 2019-03-23
RPS Challenge by Tomé

saypop:


README.md
```diff
@@ -1,22 +1,11 @@
-# RPS Challenge
+# ROCK PAPER SCISSORS! A Basic Web-App
```
riyadattani: Great README. It would be good to see the approach you used when breaking down the problem e.g. did you model it before you started writing the code?

README.md
```diff
@@ -28,59 +17,63 @@ So that I can enjoy myself away from the daily grind
...
 
 In code review we'll be hoping to see:
 
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
+*   All tests passing - CHECK
+*   High test coverage - CHECK
+*   Travis badge [![Build Status](https://travis-ci.org/saypop/rps-challenge.svg?branch=master)](https://travis-ci.org/saypop/rps-challenge)
+*   The code is elegant: every class has a clear responsibility, methods are short etc. - Please provide feedback on this.
```
riyadattani: There is a clear separation of concerns with your classes - impressive. They flow into one another well. The unit tests for each one are testing behaviour, not state which is good too.

lib/bot.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'weapon'
+
+class Bot
+
+  def initialize
+    @weapon = Weapon::WEAPONS.sample
```
riyadattani: I think the repetition in this class can be reduced.

lib/result.rb
```diff
@@ -0,0 +1,31 @@
...
+    @player_1_choice = player_1_choice
+    @player_2_choice = player_2_choice
+  end
+
+  def self.run(player_1_choice, player_2_choice)
+    new(player_1_choice, player_2_choice).run
+  end
+
+  def run
+    return :draw if @player_1_choice == @player_2_choice
```
riyadattani: great use of parameters in this class to avoid ruby in the .erb file.

spec/features/play_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  scenario 'when they choose scissors' do
+    srand(4)
+    enter_name_and_click_go
+    click_button 'PAPER!'
+    expect(page).to have_content('YOUR OPPONENT CHOSE SCISSORS!')
+  end
+end
+
+# a winner will be declared
+
```
riyadattani: This feature could be on a separate page just for ease.

PR #1232, opened 2019-03-24
Sam - finished the basic challenge

samanthaixer:
I think this is a good example of TDD
I used a lot of debugging techniques
I think this is a good example of MVC
I think this is a good example of single responsibility and delegation
I think this is a good example of writing code that is easy to change
I think this is a good example of DRY tests/code
I used doubles and mocks


.rspec
```diff
@@ -1,2 +1,3 @@
 --color --format documentation
```
m-budryte: 2 RSpec fails:
When player selects Rock they see that in the response (FAILED - 1)
  When player selects an option they can see the computer choice (FAILED - 2)

Notes.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+```
+As a marketeer
+So that I can clearly see who has won the game
+I would like to see a clear message with the winner's name
+```
+
+
+## Approach:
```
m-budryte: Detailed outline of approach - makes it easy to understand code, great!

lib/game.rb
```diff
@@ -0,0 +1,26 @@
+require './lib/rps_rules'
+
+class Game
+  attr_reader :player_1, :player_2, :winner
+
+  def initialize(player_1, player_2, rules = RpsRules)
```
m-budryte: I like that Rps rules is a separate class - makes it easier to input any rules in the future + good for testing

lib/rps_rules.rb
```diff
@@ -0,0 +1,30 @@
+class RpsRules
+  attr_reader :list
+  WEAPONS = ["Rock", "Paper", "Scissors"]
+
+  def initialize
+    @list = [{ result: "@player_1", player_1: "Scissors", player_2: "Paper" },
```
m-budryte: Consider refactoring, here is my version of the shorter syntax

RULES = { "Rock" => "Scissors", "Scissors" => "Paper", "Paper" => "Rock" }

    if @player.choice == @computer.choice
      "It's a draw!"
    elsif @player.choice == RULES[@computer.choice]
      "The computer won :("
    else
      "You won!"
    end

lib/rpssl_rules.rb
```diff
@@ -0,0 +1,28 @@
+class RpsslRules
```
m-budryte: Duplicate of rps rules - adding to the previous comment you may consider using this hash:

RULES = { rock: [scissors, lizard],
          paper: [:rock, :spock],
          scissors: [:paper, :lizard],
          lizard: [:paper, :spock],
          spock: [:rock, :scissors] }


*Taken from rubric

public/style.css
```diff
@@ -0,0 +1,36 @@
+body {
```
m-budryte: I like the color scheme!

spec/features/playing_rps_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    fill_in_name_and_submit
+    click_button "Rock"
+    expect(page).to have_content "You selected: Rock"
+  end
+
+  scenario 'When player selects an option they can see the computer choice' do
+    allow_any_instance_of(Array).to receive(:sample).and_return('Scissors')
+    fill_in_name_and_submit
+    click_button "Rock"
+    # Can I mock the behaviour of the computer from here?
```
m-budryte: yes! you can either use srand eg srand(67890) which always returns the same 'random' value.
 The computer needs to be encapsulated into the different class as well as it behaves differently from the 'human' player

spec/unit/game_spec.rb
```diff
@@ -0,0 +1,74 @@
+require 'game'
```
m-budryte: these tests are good - however similiar tests (eg both players choose the same weapon) can be put into one: 
  it "Returns :draw when both players did the same move" do
    allow(player).to receive(:choice).and_return("Rock")
    allow(computer).to receive(:choice).and_return("Rock")
    allow(player).to receive(:choice).and_return("Paper")
    allow(computer).to receive(:choice).and_return("Paper")
    allow(player).to receive(:choice).and_return("Scissors")
    allow(computer).to receive(:choice).and_return("Scissors")
    expect(game.result).to eq("It's a draw!")
  end

views/layout.erb
```diff
@@ -0,0 +1,15 @@
...
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <title>Rock Paper Scissors</title>
+  <link rel="stylesheet" href="style.css">
+</head>
+
+<body>
+  <%= yield %>
```
m-budryte: COOL!

m-budryte:
_I think this is a good example of TDD_
I agree with this statement - I can observe a 100% coverage of tests and the descriptions are clear

_I used a lot of debugging techniques_
Two RSpec tests are failing but with minor formatting issues - have extra \ns

_I think this is a good example of MVC_
It is, the files are clearly arranged in the corresponding folders

_I think this is a good example of single responsibility and delegation_
Yes, however, the computer may as well have its own class :)

_I think this is a good example of writing code that is easy to change_
It is and your own comment and outline of the approach make it easy to read and navigate.

_I think this is a good example of DRY tests/code_
See comments

_I used doubles and mocks_
+++++

lib/rps_rules.rb
samanthaixer: Thanks. I have refactored now to make it much simpler!

PR #1233, opened 2019-03-24
Amy Jordan basic challenge

amyj0rdan:
**I can TDD anything**
I wrote feature tests, which I then broke down into unit tests. Where I wrote too large a feature test, it quickly became apparent because it was harder to write unit tests. I therefore broke the feature tests down further and was more easily able to write appropriate unit tests, make them pass and move on to the next feature test.
My learnings here were to make sure I break the requirements down into the smallest possible features.

**I can debug anything**
I followed the process of tightening the loop and getting visibility to find where any errors occurred in my code and to solve them quickly. I encountered some errors that I hadn't come across and was able to use this approach to quickly resolve these.

**I can use mocks and doubles**
I used mocks and doubles to override the random behaviour of the computer choosing a weapon. I also need to use these in my feature tests for the first time and was able to do so.

**I can build a simple web app**
I created the app from scratch! (Although I did refer back to my Battle repo a fair bit for syntax.) It functions per the basic instructions and has some v basic CSS design. If I had more time this weekend I would have considered the things in my Areas for Development (in README).

**I understand the MVC pattern**
I feel confident that I understand the MVC pattern but didn't feel as confident in deciding what needed to go into which class. I felt more confident on this last week when we didn't have the V and C layers so I feel like I will regain confidence in this with more practice on the V and C elements of building a web app.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,21 @@
 require 'sinatra/base'
 
 class RockPaperScissors < Sinatra::Base
+
+  enable :sessions
+
   get '/' do
-    'Tests work!'
+    erb :index
+  end
+
+  post '/name' do
+    session[:name] = params[:player_name]
+    redirect '/play'
+  end
+
+  get '/play' do
+    @name = session[:name]
+    erb :play
   end
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,9 @@
+require 'sinatra/base'
+
+class RockPaperScissors < Sinatra::Base
+  get '/' do
+    'Hello RockPaperScissors!'
+  end
+
+  run! if app_file == $0
+end
--- a/app.rb
+++ b/app.rb
@@ -2,7 +2,7 @@ require 'sinatra/base'
 
 class RockPaperScissors < Sinatra::Base
   get '/' do
-    'Hello RockPaperScissors!'
+    'Tests work!'
   end
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -1,4 +1,5 @@
 require 'sinatra/base'
+require './lib/player.rb'
 
 class RockPaperScissors < Sinatra::Base
 
@@ -9,13 +10,24 @@ class RockPaperScissors < Sinatra::Base
   end
 
   post '/name' do
-    session[:name] = params[:player_name]
+    session[:player] = Player.new(params[:player_name])
     redirect '/play'
   end
 
   get '/play' do
-    @name = session[:name]
+    @player = session[:player]
     erb :play
+  end
+
+  post '/choice' do
+    @player = session[:player]
+    @player.choice(params[:choice])
+    redirect '/result'
+  end
+
+  get '/result' do
+    @player = session[:player]
+    erb :result
   end
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -1,34 +1,36 @@
 require 'sinatra/base'
 require './lib/player.rb'
 require './lib/computer.rb'
+require './lib/game.rb'
 
 class RockPaperScissors < Sinatra::Base
 
   enable :sessions
+
+  before do
+    @game = Game.get
+  end
 
   get '/' do
     erb :index
   end
 
   post '/name' do
-    session[:player] = Player.new(params[:player_name])
+    player = Player.new(params[:player_name])
+    Game.create_game(player, Computer.new)
     redirect '/play'
   end
 
   get '/play' do
-    @player = session[:player]
     erb :play
   end
 
   post '/choice' do
-    @player = session[:player]
-    @player.choice(params[:choice])
+    @game.player.choice(params[:choice])
     redirect '/result'
   end
 
   get '/result' do
-    @player = session[:player]
-    @computer = Computer.new
     erb :result
   end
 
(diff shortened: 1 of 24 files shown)


spec/computer_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'computer'
+
+describe Computer do
+
+  subject { described_class.new }
+
+  it 'returns rock, paper or scissors' do
+    expect(subject.random_choice).to eq('rock').or eq('paper').or eq('scissors')
```
micosmin: What do you think about stubbing the randomness here instead of eq every option?
allow_any_instance_of(Array).to receive(:sample).and_return('scissors')

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def initialize(player = Player.new, computer = Computer.new)
+    @player = player
+    @computer = computer
+  end
+
+  def winner
+    if @computer.weapon == @player.weapon.downcase
+      "It's a draw!"
```
micosmin: Apparently we should not include presentation screens in this class, but rather have it in another class which this one is calling to get the message - called presentation class maybe

micosmin: Apparently they are using a view for the message:

class Game
  def result
    :win # hard coded for example purposes #<- this would be the view
  end
end

class RPSWeb < Sinatra::Application
  get '/result' do
    @game = Game.new
    erb @game.result #<- calling the view here :win
  end
end

lib/game.rb
```diff
@@ -0,0 +1,31 @@
+class Game
+
+  attr_reader :player, :computer
+
+  WINNER = { 'rock' => 'scissors', 'scissors' => 'paper', 'paper' => 'rock' }
```
micosmin: Recommendation would be to use symbols instead of strings for memory allocation efficiency concerns. Otherwise every time you call this constants the memory will allocate new slots for all the options.

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def self.create_game(player, computer)
+    @game = self.new(player, computer)
+  end
+
+  def self.get
```
micosmin: I think best practice is to have this called self.instance, or self.get_instance. Worth checking

lib/computer.rb
```diff
@@ -0,0 +1,11 @@
+class Computer
+  
+  attr_reader :weapon
+
+  OPTIONS = ['rock', 'paper', 'scissors']
```
micosmin: Apparently you should not redefine your weapons. Weapons should be defined in one place, which here should be in Game and then used in the computer class when needed

app.rb
```diff
@@ -0,0 +1,38 @@
+require 'sinatra/base'
+require './lib/player.rb'
+require './lib/computer.rb'
+require './lib/game.rb'
+
+class RockPaperScissors < Sinatra::Base
+
+  enable :sessions
+
+  before do
```
micosmin: Good use of before in sinatra

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+    @game
+  end
+
+  def initialize(player = Player.new, computer = Computer.new)
```
micosmin: Should hide the initializer from public interface - make it private

lib/game.rb
micosmin: Apparently they are using a view for the message:

class Game
  def result
    :win # hard coded for example purposes #<- this would be the view
  end
end

class RPSWeb < Sinatra::Application
  get '/result' do
    @game = Game.new
    erb @game.result #<- calling the view here :win
  end
end

lib/player.rb
```diff
@@ -0,0 +1,13 @@
+class Player
+
+  attr_reader :name, :weapon
+
+  def initialize(name)
+    @name = name
+  end
+
+  def choice(weapon)
```
micosmin: attr_writter :weapon creates the below
def choice=(weapon)
..@weapon = weapon
end

player.choice = "weapon'
instead of player.choice(weapon)

spec/computer_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'computer'
+
+describe Computer do
+
+  subject { described_class.new }
```
micosmin: subject(:computer) { described_class.new }

spec/features/play_game_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    fill_in_name
+    click_button 'Scissors'
+    expect(page).to have_content "You chose scissors"
+  end
+end
+
+feature 'Computer returns random weapon' do
+  scenario 'computer returns rock' do
+    fill_in_name
+    srand(2)
```
micosmin: What do you think about stubbing the randomness here with this:
allow_any_instance_of(Array).to receive(:sample).and_return('scissors')

spec/features/results_spec.rb
```diff
@@ -0,0 +1,26 @@
+require_relative '../../app.rb'
```
micosmin: require './app'

views/index.erb
```diff
@@ -0,0 +1,20 @@
+<head>
+  <link rel = "stylesheet"
```
micosmin: Could use a layout.erb template and yield to it when you are calling a view. On the plus side, you can add your stylesheet there and it will be available to each view

<head>
...
  <link rel="stylesheet" href="/css/styles.css">
</head>

<body>
  <%= yield %>
</body>

</html>

PR #1235, opened 2019-03-24
James Palmer

jpalmerr:
Basic challenge

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,23 +1,5 @@
 # RPS Challenge
 
-Instructions
--------
-
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Task
-----
-
-Knowing how to build web applications is getting us almost there as web developers!
-
-The Makers Academy Marketing Array ( **MAMA** ) have asked us to provide a game for them. Their daily grind is pretty tough and they need time to steam a little.
-
-Your task is to provide a _Rock, Paper, Scissors_ game for them so they can play on the web with the following user stories:
-
 ```sh
 As a marketeer
 So that I can see my name in lights
@@ -36,11 +18,6 @@ Hints on functionality
 - the game will choose a random option
 - a winner will be declared
 
-
-As usual please start by
-
-* Forking this repo
-* TEST driving development of your app
 
 
 ## Bonus level 1: Multiplayer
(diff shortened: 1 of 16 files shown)


app.rb
```diff
@@ -0,0 +1,67 @@
+require 'sinatra/base'
+
```
hybridbad: require './lib/turn'

hybridbad: whoops wrong button meant to comment

app.rb
```diff
@@ -0,0 +1,67 @@
...
+class RPS < Sinatra::Base
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    @name = params[:name]
+    erb :play
+  end
+
```
hybridbad: inject into instance on your play POST route (well global var but I don't know how do a singleton class yet)

in names
@player_1 = params[:name] 

in play
$turn = Turn.new(player_1, player_2)
@move = params[:move]

in result
@turn = $turn

@turn.game(@move)

README.md
```diff
@@ -1,23 +1,5 @@
 # RPS Challenge
```
hybridbad: This readme doesn't really tell me much about 'your' program only the instructions of the challenge. Maybe add some instructions for people so they know how to run it. Break down the user stories and explain your approach?

hybridbad: Sorry I am blind AF.

Saw your process.md file - maybe export that into the README as thats the first thing that gets loaded

app.rb
```diff
@@ -0,0 +1,67 @@
...
+# how can i move these to a ruby file within lib directory? => i can test them
+
+  def computer_move
+    ['Rock', 'Paper', 'Scissors'].sample
+  end
+
+  def game(user = @move, computer = @computer_move)
+
```
hybridbad: nice use of case switch statement

process.md
```diff
@@ -0,0 +1,37 @@
...
+
+After submitting their move, the user is directed to the /play page.
+At this point, the page simply returns what move the user entered.
+
+###### Ideas
+
+- a 'play again' link returning the user to the /names page
+
+###### Difficulties
+
```
hybridbad: Good reflection!

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,7 @@
+def sign_in_and_play
+  visit('/')
+  fill_in :name, with: 'James'
+  click_button 'Lets play!'
+end
+
```
hybridbad: good use of web helpers, using things learnt from battle challenge

public/stylesheet.css
```diff
@@ -0,0 +1,13 @@
+body {
```
hybridbad: nice to see some css practise! Have you thought about using a css library like bootstrap?

app.rb
jpalmerr: This is what I was trying but think I went wrong somewhere, probably got a detail wrong along the way. May try to refactor if time allows. Thankyou

app.rb
alexanderchen6: yes. agreee need add turn here

PR #1238, opened 2019-03-24
rps-challenge

EvelynAleLeo:


spec/features/features_spec.rb
```diff
@@ -0,0 +1,16 @@
+feature 'Player enter name' do
```
LaurenQurashi: I really like how you've staged your testing, these are in nice small steps which show how the functionality grows.

rps_app.rb
```diff
@@ -0,0 +1,34 @@
...
+    erb(:index)
+  end
+
+  post '/names' do
+    @player1 = params[:player1]
+    session['player_name'] = @player1
+    erb(:play)
+  end
+
+  post '/selection' do
```
LaurenQurashi: Ideally, this should be in another file which you call upon in the route. However, the code in here looks good and is clear to another reader.

README.md
```diff
@@ -1,19 +1,7 @@
...
--------
-
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
 Task
 ----
```
LaurenQurashi: Your Readme needs a few more things in it to be considered for review by the coaches. There are some guidelines on Makers' Github page if you need to amend this in the future :)

views/selection.erb
```diff
@@ -0,0 +1,3 @@
+<p><%= @player_name%> has selected <%= @selection%></p>
+<p>Computer has selected <%= @server_selection%></p>
```
LaurenQurashi: Overall, Really good code! everything works as it should. The only improvements I can think of are: 
Add some CSS so the page looks appealing. 
Move your logic from the app.rb file into another file and call it under your routing.
Well done 🥇

PR #1240, opened 2019-03-24
RPS Challenge. Wardah

WardahTarar:


Gemfile
```diff
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 ruby '2.5.0'
 
 gem 'rake'
+gem 'shotgun'
```
Hives: this isn't needed to make your code run, so no need to include it in the gemfile

README.md
```diff
@@ -1,86 +1,38 @@
...
-
-## Bonus level 1: Multiplayer
-
-Change the game so that two marketeers can play against each other ( _yes there are two of them_ ).
-
-## Bonus level 2: Rock, Paper, Scissors, Spock, Lizard
-
-Use the _special_ rules ( _you can find them here http://en.wikipedia.org/wiki/Rock-paper-scissors-lizard-Spock_ )
+## Alternative 
+You can play the game here https://rps-game1.herokuapp.com/
```
Hives: great touch!

app.rb
```diff
@@ -0,0 +1,29 @@
...
+require './lib/auto.rb'
+
+class RPS < Sinatra::Base
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    $player = Player.new(params[:name])
```
Hives: "global variables are evil" [according to the rubric](https://github.com/makersacademy/rps-challenge/blob/master/docs/review.md#use-of-global-variables)! 

maybe refer to the [last part of the battle project](https://github.com/makersacademy/course/blob/master/intro_to_the_web/killing_the_global_variable.md) to see a way to use a 'class method' to avoid using a global variable. the walkthrough goes through implementing this in detail.

app.rb
```diff
@@ -0,0 +1,29 @@
...
+    erb :names
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post "/game" do
+    @move = params[:move]
+    $auto_move = Auto.new
```
Hives: another global variable

public/style_name.css
```diff
@@ -0,0 +1,36 @@
+body {
+  background-color:#000;
+}
+.glowing-text{
```
Hives: 🤯 LOVE IT 🤯

views/game.erb
```diff
@@ -0,0 +1,24 @@
...
+  <h1> You WON! <p> &#x1F603; &#x1F389; &#x1F929; </p> </h1>
+<% elsif @move=='Scissors' && @auto_move=='Paper' %>
+  <h1> You WON! <p> &#x1F603; &#x1F389; &#x1F929; </p> </h1>
+<% elsif @move=='Paper' && @auto_move=='Rock' %>
+  <h1> You WON! <p> &#x1F603; &#x1F389; &#x1F929; </p> </h1>
+<% elsif @move==@auto_move %>
+  <h1> It's a draw! <p> &#x1F610; </p> </h1>
+<% else%>
+  <h1> You Lost! <p> &#x1F641; </p> </h1>
+<% end %>
```
Hives: you've got a bunch of business logic in the view here. that should probably be in the model, in a Game class or something, and the result should get passed in via the controller. see [this section of the rubric](https://github.com/makersacademy/rps-challenge/blob/master/docs/review.md#calling-business-logic-from-the-view) for more detail

Hives: love the emojis though!

views/game.erb
```diff
@@ -0,0 +1,24 @@
+<center>
+
+<link rel="stylesheet" type="text/css" href="/style.css">
+
+<h2> You chose '<%=@move%>' as your move! </h2>
+<br>
+<h2> Game chose '<%=@auto_move%>' as a random move! </h3>
```
Hives: mismatching start and end tags here

lib/auto.rb
```diff
@@ -0,0 +1,7 @@
+class Auto
```
Hives: this class name is not very descriptive of its responsibility

app.rb
Hives: this one is unnecessary beause you're only referring to it within this scope.
you could do `@auto_move = Auto.new.random_pick` and it would achieve the same thing

app.rb
```diff
@@ -0,0 +1,29 @@
...
+    $player = Player.new(params[:name])
+    @player = $player.name
+    erb :names
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post "/game" do
```
Hives: i think this route could be named better - it's actually the results page, so maybe something like "/result" would be more appropriate?

spec/auto_spec.rb
```diff
@@ -0,0 +1,14 @@
+require './lib/auto.rb'
+
+describe Auto do
+  it "checks the existence of class" do
+  end
```
Hives: nothing in this test!

spec/auto_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  it { is_expected.to respond_to(:random_pick) }
+
+  it "checks if it can return a random pick" do
+    auto = Auto.new
+    allow(auto).to receive(:random_pick).and_return('Rock')
+    expect(auto.random_pick).to eq 'Rock'
```
Hives: this is a vacuous test - since you've stubbed out the behaviour of `auto`, all you're testing is the stubbed out behaviour, not the actual behaviour of the class. if you comment out the content of the Auto#random_pick method then this test still passes.

spec/auto_spec.rb
```diff
@@ -0,0 +1,14 @@
+require './lib/auto.rb'
+
+describe Auto do
+  it "checks the existence of class" do
+  end
+
+  it { is_expected.to respond_to(:random_pick) }
```
Hives: i think it's ok to use something like this as you're developing the class, but you shouldn't leave it in as it doesn't test very much. you should have a test that checks the behaviour of `#random_pick`, not just that it exists

spec/player_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  subject(:jon) { Player.new('Jon Snow') }
+
+  describe '#name' do
+    it "checks for player's name" do
+      expect(jon.name).to eq "Jon Snow"
+    end
+  end
+
+  it 'checks if the class exists' do
+  end
```
Hives: nothing happening here!

spec/player_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  it { is_expected.to respond_to(:pick).with(1).argument }
```
Hives: no need to leave this in - this test is implied by the test below

spec/features/auto_move_spec.rb
```diff
@@ -0,0 +1,12 @@
+require './lib/auto.rb'
+
+feature 'Random move' do
+  scenario 'displayes the randomly chosen move by the game' do
+    sign_in_and_play
+    click_button 'Lets Play'
+    click_button 'Confirm'
+    auto = Auto.new
+    allow(auto).to receive(:random_pick).and_return('Paper')
+    expect(auto.random_pick).to eq 'Paper'
```
Hives: you've stubbed out the behaviour you're testing here, so it's not actually testing anything. if you comment out the content of `Auto#random_pick` then this test still passes!

also, this is more of a unit test for `Auto` than a feature test. the feature test should be checking maybe that the right move is displayed on screen, and then the unit test for `Auto` would check that the method works. Your scenario says 'displays the randomly chosen move...' which is what you should be testing.

spec/features/winner_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+feature 'Game Results' do
+  scenario "checks if it's a draw for having Rock chosen against Rock" do
+    allow_any_instance_of(Auto).to receive(:random_pick).and_return('Rock')
+    sign_in_and_play
+    click_button 'Lets Play'
+    find("option[value='Rock']").select_option
+    click_button 'Confirm'
+    expect(page).to have_content "It's a draw!"
+  end
+end
```
Hives: this is good for your feature tests, but the rubric says you should be testing all possible game outcomes in your feature tests, [see here](https://github.com/makersacademy/rps-challenge/blob/master/docs/review.md#not-testing-all-of-the-game-logic-in-unit-tests). i guess this would be easier if you had a Game class to test

PR #1244, opened 2019-03-24
d

justcallmegoyo:
dd


views/chosen.erb
```diff
@@ -8,7 +8,7 @@
 <center>
 
 <h1>You chose <%=session[:chosen]%></h1>
-
+<h2>The computer chose <%= %></h2>
```
kimdiep: The computer choice can be wrapped in the Computer class. You have the right infrastructure there, you can put the .sample method outside the initialize in the computer class and then you can call this method :) This might cause some funky random behaviour of the computer class, as I experienced as a bug :P

PR #1249, opened 2019-03-25
rps-challenge riya

riyadattani:


lib/computer.rb
```diff
@@ -0,0 +1,10 @@
+class Computer
```
saypop: +1 for extracting this class

lib/game.rb
```diff
@@ -0,0 +1,23 @@
...
+
+class Game
+  attr_reader :player, :computer
+
+  def initialize(player = Player.new, computer = Computer.new)
+    @player = player
+    @computer = computer
+  end
+
+  BEATMAP = { "Rock" => "Scissors", "Scissors" => "Paper", "Paper" => "Rock" }
```
saypop: BEATMAP... great name for an international music show!

lib/game.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'player'
+require_relative 'computer'
+
+class Game
+  attr_reader :player, :computer
+
+  def initialize(player = Player.new, computer = Computer.new)
```
saypop: good use of dependency injection

lib/game.rb
```diff
@@ -0,0 +1,23 @@
...
+    @player = player
+    @computer = computer
+  end
+
+  BEATMAP = { "Rock" => "Scissors", "Scissors" => "Paper", "Paper" => "Rock" }
+
+  def winner
```
saypop: sound logic, it would be easy to change the game to use a different beatmap in the future

views/play.erb
```diff
@@ -0,0 +1,16 @@
...
+  <button name="option" type="submit" value="Paper">Paper</button>
+  <button name="option" type="submit" value="Scissors">Scissors</button>
+</form>
+
+
+<% if @player_move %>
+    Your move: <%=@player_move%>
+    Computer's move: <%=@computer_choice%>
+    <br>
+    <%=@game.winner%>
```
saypop: They advise against this, rather use separate pages for each result than returning strings from your methods.

lib/game.rb
saypop: also you've used hashes and arrays really well throughout

app.rb
```diff
@@ -0,0 +1,26 @@
...
+class RPS < Sinatra::Base
+  enable :sessions
+  configure(:development) { set :session_secret, "something" }
+
+  get '/' do
+    erb :index
+  end
+
+  post '/info' do
+    $player = Player.new(params[:name])
```
saypop: you're still using global variables, happy to walk you through class instance variables if you need

spec/features/winner_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'winner' do
+  scenario 'display the winner after player and computer made a move' do
+    srand(678_910)
+    sign_in_and_play
+    click_button 'Rock'
+    expect(page).to have_content "Your move: Rock"
+    expect(page).to have_content "Computer's move: Scissors"
+    expect(page).to have_content "You win!"
```
saypop: There should be additional tests for the draw and lose outcomes

lib/game.rb
saypop: This could be another class

app.rb
```diff
@@ -0,0 +1,26 @@
+require 'sinatra/base'
```
saypop: General notes:
- good folder layout
- capybara app set
- unit tests cover all of your model
- randomness stubbed out of tests
- you're using global variables
- well named routes, classes and methods
- easy to read code and tests, think I could edit if I needed to
- you clearly have a good handle on MVC and single responsibilty principle

views/play.erb
```diff
@@ -0,0 +1,16 @@
...
+
+<form action='/play'>
+  <p> Choose one option from the following: </p>
+  <button name="option" type="submit" value="Rock">Rock</button>
+  <button name="option" type="submit" value="Paper">Paper</button>
+  <button name="option" type="submit" value="Scissors">Scissors</button>
+</form>
+
+
+<% if @player_move %>
```
saypop: Try to avoid using logic in views, seperate the pages and redirect from controller

PR #1254, opened 2019-04-07
Challenge attempt.

Saicam:


lib/app.rb
```diff
@@ -0,0 +1,56 @@
...
+    @game = Game.instance
+  end
+
+  get "/" do
+    erb :index
+  end
+
+  post "/name" do
+    player_1 = Player.new(params[:player_1_name])
+    if params[:player_2_name].empty?
```
lindseydew: Nice use of optional player 2 logic here

lib/game.rb
```diff
@@ -0,0 +1,30 @@
+class Game
+  attr_reader :current_turn
+  def initialize(player_1, player_2)
+    @players = [player_1, player_2]
+    @current_turn = player_1
+  end
+
+  def player_1
```
lindseydew: Players could perhaps be two independent variables rather than an array of size two, which would mean the helper methods weren't needed

lib/result_checker.rb
```diff
@@ -0,0 +1,67 @@
+class ResultChecker
+  def self.check(player_1, player_2)
+    to_obj(player_1).vs(player_2)
+  end
+
+  def self.to_obj(player_1)
+    case player_1
+    when "rock"
```
lindseydew: Nice use of a case statement! This could also be modelled as a hash map

Saicam: This was actually the question I wanted to make, because it didn't work when I tried to use a hash. I tried again and works so I guess I made a different mistake and I didn't realize.

PR #1255, opened 2019-04-08
RPS Challenge

seanbanford:
RPS  main challenge completed, 1 bonus completed.

### DIFF
--- a/spec/computer_spec.rb
+++ b/spec/computer_spec.rb
@@ -1,0 +1,11 @@
+require 'computer'
+
+describe Computer do
+  describe 'Computer selected weapon' do
+    let(:computer) {double :computer_selection}
+    it 'selects a weapon at random' do
+    allow(computer).to receive(:random_selection).and_return("Scissors")
+    expect(computer.random_selection).to eq "Scissors"
+    end
+  end
+end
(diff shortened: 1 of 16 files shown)


spec/computer_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'computer'
+
+describe Computer do
+  describe 'Computer selected weapon' do
+    let(:computer) {double :computer_selection}
+    it 'selects a weapon at random' do
+    allow(computer).to receive(:random_selection).and_return("Scissors")
```
lindseydew: Nice use of mocking here

spec/feature/show_winner_spec.rb
```diff
@@ -0,0 +1,7 @@
+feature 'Show if you won or lost' do
+  scenario 'Player chooses Rock' do
+    sign_in_and_play
+    click_button 'Rock'
+    page.should satisfy {|page| page.has_content?("Victorious!") or page.has_content?("defeated you!") or page.has_content?("Stalemate...")}
```
lindseydew: Might be nice to break this test down into separate scenarios so that the assertion statement can be more explicit

PR #1257, opened 2019-04-08
RPS challenge

Codeman15:


lindseydew:
Great job on the Friday challenge! I've added a couple of comments. If you would like to talk about any of these face to face please drop me a note on slack and we can schedule a time.

PR #1259, opened 2019-04-20
first commit - added myREADME.md

callum-marshall:


app.rb
```diff
@@ -1,5 +1,5 @@
 require 'sinatra/base'
-require './game.rb'
+require 'game.rb'
```
NapperJLG: Make sure to remember to change file paths when moving files to different folders...we had some trouble running the code.

cpu.rb
```diff
@@ -0,0 +1,7 @@
+class Cpu
+
+  def choice
+    ['ROCK', 'PAPER', 'SCISSORS'].sample
```
NapperJLG: This array can be stored as a constant so that it isn't being created every time the choice method is called (or so the rubric says...I didn't do this either!)

PR #1260, opened 2019-04-20
User story 1 complete. Unable to test fully as rack testing not worki…

stuartpet:
…ng, I will try to solve and resubmit my commit

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,13 +1,20 @@
 source 'https://rubygems.org'
 
+git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
+
+
 ruby '2.5.0'
 
 gem 'rake'
 gem 'sinatra'
+gem 'selenium-webdriver'
 
 group :test do
+  gem 'rack'
   gem 'capybara'
   gem 'rspec'
+  gem 'rspec-sinatra'
+  gem "rack-test", require: "rack/test"
   gem 'rubocop', '0.56.0'
   gem 'simplecov', require: false
   gem 'simplecov-console', require: false
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,22 @@
+require 'sinatra/base'
+require 'rack/test'
+require 'test/unit'
+
+class RPS < Sinatra::Base
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    session[:name] = params[:name]
+    redirect '/play'
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+
+  run! if app_file == $0
+end
--- a/config.ru
+++ b/config.ru
@@ -1,0 +1,3 @@
+require './app'
+
+run MyApp
--- a/spec/register_spec.rb
+++ b/spec/register_spec.rb
@@ -1,0 +1,11 @@
+require 'spec_helper'
+
+
+feature "registering name" do
+  scenario "I can register my name" do
+    visit '/'
+    fill_in "name", with: "Dave"
+    click_button "Submit"
+    expect(page).to have_content "Dave"
+  end
+end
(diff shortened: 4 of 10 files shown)


Gemfile
```diff
@@ -1,13 +1,20 @@
 source 'https://rubygems.org'
 
+git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
```
ocripps24: I think you only need this line if you're planning to use Gems from a custom source i.e. Github rather than the rubygems.org site. I don't think you end up using :github anyway :)

app.rb
```diff
@@ -0,0 +1,22 @@
+require 'sinatra/base'
+require 'rack/test'
```
ocripps24: I don't think you need this here

app.rb
```diff
@@ -0,0 +1,22 @@
...
+require 'test/unit'
```
ocripps24: I don't think you need this here

app.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class RPS < Sinatra::Base
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    session[:name] = params[:name]
```
ocripps24: I think you have to add a line saying 'enable sessions' or similar below the class name if you want to store params in sessions

config.ru
```diff
@@ -0,0 +1,3 @@
+require './app'
+
+run MyApp
```
ocripps24: Should 'MyApp' be the name of your class in the controller app.rb file? i.e. 'RPS' otherwise i'm not sure it knows what to run

spec/register_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'spec_helper'
```
ocripps24: I'm not sure you need to require this here

spec/register_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+
+
+feature "registering name" do
+  scenario "I can register my name" do
+    visit '/'
+    fill_in "name", with: "Dave"
```
ocripps24: I had trouble with this command unless I used parentheses (although perhaps that's an error particular to my code) i.e. fill_in('player_1_name', with: 'Pikachu')

spec/spec_helper.rb
```diff
@@ -1,6 +1,11 @@
+# ENV['RACK_ENV'] = 'test'
```
ocripps24: I'm not sure your rackup environment will run/test properly with this commented out. 

I'd also be expecting to see this line in this file: 
require File.join(File.dirname(__FILE__), '..', 'app.rb')

ocripps24: For some reason the underscores surrounding the 'FILE' have been removed. I understand you are meant to have two underscores either side of 'FILE'

views/index.erb
```diff
@@ -0,0 +1,5 @@
+<<form action="/name" method="post">
```
ocripps24: There's a superfluous '<' at the start of your form

views/index.erb
```diff
@@ -0,0 +1,5 @@
...
+<input type="text" name="name" value=""/>
+<input type="text" value="Submit"/>
```
ocripps24: I think your type 'text' should be 'submit' here

views/register.rb
```diff
@@ -0,0 +1,18 @@
+require "sinatra/base"
+
```
ocripps24: This file appears to be another controller, which is confusing because you have a controller already called 'app.rb'

ocripps24 (commented):
I couldn't test your code so i've gone through and left some comments where I can see things that might affect the functionality of your app

PR #1263, opened 2019-04-21
RPS Challenge

shektor:


SarpongAbasimi (commented):
Overall great codebase. Very easy to read and understand.
The app works as expected, but I think you could have enhanced the UI a bit better to improve the user experience.

PR #1265, opened 2019-04-21
RPS challenge - user can play against computer 

jordantroberts:
User can play against the computer. Tried to implement some basic CSS but I have no experience with it so far so would like to practice this to make the game look better :)


lib/game.rb
```diff
@@ -0,0 +1,27 @@
+class Game
+
+  attr_reader :player, :computer
+
+  def initialize(player, computer)
+    @player = player
+    @computer = computer
+  end
+
+  def winner?
```
ayelishaamoah: I think this method could be split out to make it shorter - although it is still very human readable. One possible solution would be to store the rules in a hash that you can iterate over.

spec/features/player_choice_spec.rb
```diff
@@ -0,0 +1,26 @@
+feature 'Given options for RPS' do
```
ayelishaamoah: very concise tests - good use of the helper function

views/choice.erb
```diff
@@ -0,0 +1,16 @@
+<head>
+ <link rel="stylesheet" href="style2.css">
+</head>
+
+<h1 style="text-align:center;"><font size="5" color="orange" font face="arial"RESULTS!</font></h1><br><br>
```
ayelishaamoah: I can see that you've included some styling inline - have you considered adding this to your css file?

README.md
```diff
@@ -28,59 +20,26 @@ So that I can enjoy myself away from the daily grind
...
 - Rock beats Scissors
 - Scissors beats Paper
 - Paper beats Rock
 
-In code review we'll be hoping to see:
-
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
+## Steps to run this program
```
ayelishaamoah: useful instructions and I like that you linked to the url that you need to use to open the app.

app.rb
```diff
@@ -0,0 +1,41 @@
...
+    session[:player_name] = params[:player_name]
+    redirect '/play'
+  end
+
+  get '/play' do
+    @player_name = session[:player_name]
+    erb :play
+  end
+
+  get '/rock' do
```
ayelishaamoah: I've noticed that three routes for getting the users choice are the same except for the player choice that's being passed in. 
Have you thought about using a post request in the form that you have in your play view storing this as an instance variable?

app.rb
```diff
@@ -0,0 +1,41 @@
...
+
+class RPS < Sinatra::Base
+  enable :sessions
+  
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    session[:player_name] = params[:player_name]
```
ayelishaamoah: You are currently able to play the game without entering a name - have you thought about throwing an exception if this happens?

PR #1268, opened 2019-04-21
Rock, Paper, Scissors Attempt

Hannah-Frost:


views/outcome.erb
```diff
@@ -0,0 +1,9 @@
+<% if @game.draw? %>
```
sejinkay: I didn't know that you can do if/elsif like this, looks neat and easy to understand

views/welcome.erb
```diff
@@ -0,0 +1,8 @@
+<center><h1>Welcome to RPS, <span style="color: red"><%= @player_name %>!</span></h1>
+<h2>Here are the rules:</h2>
```
sejinkay: I like how you explain the rule to the user(user friendly), it's easy thing to forget as we all assume that everyone knows the rule.

app.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  get '/game' do
+    erb :game
+  end
+
+  post '/outcome' do
+    @player_name = session[:player_name]
+    @player_decision = params[:decision]
+    @computer_decision = ["Rock", "Paper", "Scissors"].sample
```
sejinkay: Is there a reason why you put this code in the app.rb rather than the game.rb? :)

README.md
```diff
@@ -1,86 +1,31 @@
...
-Task
-----
+## What criteria should this program satisfy?
 
-Knowing how to build web applications is getting us almost there as web developers!
+- Enter your name before the game.
+- Presented with choices (rock, paper and scissors).
+- User can choose one option.
+- The game will choose a random option.
+- A winner will be declared.
```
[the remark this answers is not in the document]
Hannah-Frost: I will do this now, thank you!

app.rb
Hannah-Frost: It was originally temporary, however I am looking into adding it into the Game class now. Thank you for your feedback :)

app.rb
```diff
@@ -0,0 +1,35 @@
...
+    @game = Game.new(@player_decision, @computer_decision)
+    erb :outcome
+  end
```
CalumDarroch: Good clean logic. However, you might consider extracting some of it to one or two more classes! This way you can obey the SRP, extract the logic from the controller layer to the model layer, and it will make your code easier to change in future! =)

lib/game.rb
```diff
@@ -0,0 +1,25 @@
...
+    return true if @player == "Rock" && @computer == "Scissors"
+    return true if @player == "Paper" && @computer == "Rock"
+    return true if @player == "Scissors" && @computer == "Paper"
+  end
+
+  def lose?
+    return true if @player == "Paper" && @computer == "Scissors"
+    return true if @player == "Scissors" && @computer == "Rock"
+    return true if @player == "Rock" && @computer == "Paper"
+  end
```
CalumDarroch: One line conditionals are good. However, we could do better! One thing to look at would be to extract this logic to a different model completely... perhaps we could store the winning combinations as a hash?

lib/game.rb
```diff
@@ -0,0 +1,25 @@
...
+  attr_reader :player, :computer
+
+  def initialize(player, computer)
+    @player = player
+    @computer = computer
+  end
+
+  def draw?
+    @player == @computer
+  end
```
CalumDarroch: Good draw method! Keeps things nice and simple.

views/index.erb
```diff
@@ -0,0 +1,7 @@
+<center><h1>Welcome to Rock, Paper, Scissors</h1>
+<h2>Please enter your name.</h2>
+<form action="/name" method="post">
+  <label for="name">Name:</label>
+  <input type="text" name="player_name">
+  <input type="submit" value="Submit">
+</form></center>
```
CalumDarroch: Nice formatting on erb files generally - everything displays well when running the app.

lib/game.rb
```diff
@@ -0,0 +1,25 @@
+class Game
+  attr_reader :player, :computer
+
+  def initialize(player, computer)
+    @player = player
+    @computer = computer
+  end
```
CalumDarroch: These variables "player" and "computer" could perhaps be more clearly named, so that you know they are the chosen moves rather than the player or computer's names, for example.

Hannah-Frost: I've updated my Game class to have 'player_decision' and 'computer_decision' now. Thank you :)

PR #1270, opened 2019-04-22
RPS challenge - First attempt

sejinkay:
The user can play basic game, need to fix some tests to work/kill global variables

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,5 +1,6 @@
 require 'sinatra/base'
 require_relative 'lib/player'
+require_relative 'lib/game'
 
 class RPS < Sinatra::Base
   enable :sessions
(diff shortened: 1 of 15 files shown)


app.rb
```diff
@@ -0,0 +1,35 @@
...
+  enable :sessions
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/names' do
+    # $player = Player.new(params[:player])
+    redirect to('/play')
+  end
```
Hannah-Frost: Remember to delete any comments that you are no longer using in your code :)

public/style.css
```diff
@@ -0,0 +1,59 @@
...
+  html { font-family: 'Inter var', sans-serif; }
+}
+
+:root {
+  --background-main: #edf4f7;
+  --background-side: #f4f8fa;
+  --primary-blue: #32444d;
+  --secondary-blue: #b3c3cc;
+  --primary-orange: #ff6719;
+  --secondary-orange: #feada6;
```
Hannah-Frost: This looks cool! I didn't have time to look into css so it'd be awesome if you wouldn't mind explaining some of this to me!

sejinkay: Any time! :)

PR #1271, opened 2019-04-22
RPS Challenge Compare

kiahjade:


app.rb
```diff
@@ -0,0 +1,32 @@
...
+  post '/move' do
+    @random_generator = RandomMove.new
+    @opponent_move = @random_generator.random_move
+    @player_move = params[:player_move]
+    @game = Game.new(@player_move, @opponent_move)
+    erb :move
+  end
+
+  get '/secret' do
+    'This is a secret page'
```
sejinkay: Is this part of the game? :)

views/play.erb
```diff
@@ -0,0 +1,7 @@
+<h1>Hi <span style="color: red"><%= @player %></span>!</h1>
+<h2>Rock, Paper or Scissors?</h2>
+
+<form action='/move' method='post'>
+  <input type='text' name='player_move'>
```
sejinkay: I find it interesting that you can type the answer rather than just clicking the button, maybe add some action to it when it happens?

app.rb
```diff
@@ -0,0 +1,32 @@
...
+    erb :index
+  end
+
+  post "/play" do
+    @player = params[:player]
+    erb :play
+  end
+
+  post '/move' do
+    @random_generator = RandomMove.new
```
sejinkay: At first sight, I found the name @random_generator bit confusing as I wasn't sure what random thing it was generating but it got clearer at the second line. Apart from isolating it for the unit test, is there a reason why you separated the file?

PR #1273, opened 2019-04-22
Partial solution

ayelishaamoah:
Initially managed to get the app working when playing against the computer.

Attempted bonus feature - multiplayer option where I am unable to get the winner to display - this needs to be looked at. 

I would also like to refactor and add exception handling.


ayelishaamoah:
I have fixed the issue with the winner page not displaying the winner.  Having taken another look at my code, there was an issue with setting the choice on the second player - this was due to human error when refactoring and not updating where the parameter was being stored when adding multiplayer functionality.

I was able to find this error by using the p command when running the tests on the winner router which highlighted that the second players choice was always blank. This meant that feature tests weren't passing as the set_winner method was never meeting any of the conditions and not returning a valid winner.

This has highlighted the importance if gaining visibility on error messages as well as the importance of having multiple types of testing (e.g browser testing) as RSpec unit tests were passing.

lib/game.rb
```diff
@@ -0,0 +1,58 @@
+class Game
+  attr_reader :player, :player_one, :player_two, :turns
+
+  def initialize(player_one, player_two = "The Computer")
+    @player_one = player_one
+    @player_two = player_two
+    @player = player_one
+    @turns = 0
+    @rules = {
```
jordantroberts: Really clever idea to use a hash to map the rules - I used if/else statements on mine which got so messy and this is much neater and easier to read 👍

public/style.css
```diff
@@ -0,0 +1,85 @@
+* {
```
jordantroberts: CSS was brilliant !  Really impressive :)

lib/game.rb
```diff
@@ -0,0 +1,58 @@
...
+      "Paper" => "Rock",
+      "Scissors" => "Paper"
+    }
+  end
+
+  def self.create(player_one, player_two)
+    @game = Game.new(player_one, player_two)
+  end
+
+  def self.instance
```
jordantroberts: Why have you called the methods on the self object here? This is just out of interest as I personally don't really understand when and why you would need to do this? Would be grateful if you could explain? 👼

PR #1275, opened 2019-04-22
RPS Challenge

ghost:


spec/features/play_rps_spec.rb
```diff
@@ -0,0 +1,29 @@
+feature 'So that I can enjoy myself' do
+  scenario 'be able to play rps' do
+    visit '/'
+    fill_in :player_name, with: 'player'
```
domw30: You could possibly use Web Helpers on your feature test to refactor code and use less lines.

views/result.erb
```diff
@@ -0,0 +1,3 @@
+<h1>...and the computer chose <%= @computer_selection %>!</h1>
+
+You might have won or you might have lost. I'm not sure how to code this final bit yet.
```
domw30: Implement winner method so the view highlights the winner.

spec/features/play_rps_spec.rb
domw30: Possible changes

domw30 (commented):
Possible changes.

PR #1276, opened 2019-04-22
RPS Player vs Computer.

SarpongAbasimi:
I only worked on player vs computer in this challenge😅. The app is still not perfect👎🏿 so I promise to continue working on it. I would be grateful if you could give me feedback on the readability, simplicity of my codebase and test. Also,  feel free to add any other comment you want. Grazie!


lib/game.rb
```diff
@@ -0,0 +1,58 @@
...
+    [:Rock, :Paper, :Scissors].sample
+  end
+
+  def computerchose
+    @computer_choice = random_choice
+  end
+
+  def choice
+    if @player_choice
+      "#{@player.name.capitalize} chose #{@player_choice} #{@name} chose #{@computer_choice}"
```
shektor: This looks to be related to the view, could it be moved into the erb file?

lib/game.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+  end
+
+  def message
```
shektor: As with the choice method and separating view concerns, could this logic be moved to the view?

lib/game.rb
```diff
@@ -0,0 +1,58 @@
...
+      @score += 10
+      "Sorry #{@name} won! Better luck next time"
+    when 'draw'
+      'It was a Draw'
+    else
+      'Make a move'
+    end
+  end
+
+  def result
```
shektor: This method looks to be doing quite a lot, could it be refactored to follow single responsibility principles

app.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  get '/name' do
+    erb :name
+  end
+
+  get '/start' do
+    player_one = Player.new(session[:player_name])
+    @game = Game.new(player_one, session[:user_choice])
+    @game.computerchose
```
shektor: Could running this game logic be moved into the class?

PR #1282, opened 2019-04-23
RPS Challenge

Nandini0206:


spec/features/web_helpers.rb
```diff
@@ -0,0 +1,4 @@
+def sign_in_and_play
```
tsankhalpara: good use of web helpers to refactor the code

PR #1293, opened 2019-05-20
RPS 

dafplant:


app.rb
```diff
@@ -0,0 +1,27 @@
+require 'sinatra/base'
```
lbarber729: Do these need to be in app.rb, could they be in spec_helper?

app.rb
```diff
@@ -0,0 +1,27 @@
...
+require 'rspec'
+
+class Rps < Sinatra::Base
+
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    @player_name = (params[:player_name])
```
lbarber729: Could you use sessions and then redirect here?

spec/features/play_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature "Choose R-P-S" do
+  scenario 'can choose a play option' do
+    visit('/')
```
lbarber729: Could you add these 3 lines to a web_helpers so you don't have to repeat in it in your spec's?

PR #1296, opened 2019-05-20
RPS-Challenge - Single Player

VladC24:


app/views/play.erb
```diff
@@ -0,0 +1,44 @@
...
+
+<body>
+<h1><%= @game.player.name %> vs. A.I. </h1>
+
+<p><%= @game.player.name %> chose: <b><%= @game.player.weapon.upcase %></b></p>
+
+<p>A.I. chose: <b><%= @game.ai.weapon.upcase %></b></p>
+
+<div class='weapon'>
+  <% if @game.result(@game.player.weapon, @game.ai.weapon) == 'draw' %>
```
LaurenceTaylor: Could add views for each 'result' and redirect in rps.rb rather than having logic (if statements) in play.erb

PR #1297, opened 2019-05-20
Finished vs computer

n3ttl3t:


lib/game.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+    elsif self.player_choice == "scissors"
+      if self.game_choice == "paper"
+        "You win!"
+      else
+        "I win!"
+      end
+    end
+  end
+
```
olliebrownlow: Consider, for separation of concerns purposes, returning presentation strings in the relevant view.

lib/game.rb
```diff
@@ -0,0 +1,48 @@
...
+    @player_name = player_name
+    @player_choice = player_choice
+    @options = OPTIONS
+    @game_choice = random_move
+  end
+
+  def random_move
+    @options.sample
+  end
+
```
olliebrownlow: This could have been done more elegantly with different game logic - for example, using a hash to map the rules of the game and then utilising that to return the round result. Also, see below on returning presentation strings.

PR #1304, opened 2019-06-02
Ed's RPS Challenge submission

edwardreeve:


edwardreeve:
just updating to Spock/Lizard version

lib/game.rb
```diff
@@ -0,0 +1,38 @@
...
+    @winner = nil
+  end
+
+  def play
+    w1 = @player1.weapon
+    w2 = @player2.weapon
+    if @player2.weapon.nil?
+      w2 = @player2.choose_weapon
+    end
+    if w1 == w2
```
lindseydew: This method has a lot going on. The conditional logic at the bottom could be extracted into a private method which takes in w1,w2 as a dependency.
Eg:
```
private def calc_winner(w1,w2)
       if w1 == w2
```
etc

lib/player.rb
```diff
@@ -0,0 +1,14 @@
+# creates named players who can select random weapons
+class Player
+  attr_reader :name, :weapon
+  DEFAULT_NAME = 'RPS-Bot'
```
lindseydew: Nice - I like how you are handling the computer bot here. Could even check in the initialize method whether the name is `RPS-Bot` and call `choose_weapen` immediately if so. That ensures a method will be selected in the case of a computer player.

spec/features/results_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    expect(page).to have_content('Aaaand the winner is')
+  end
+
+  scenario 'your choice of weapon is displayed (single player)' do
+    one_player_game_played
+    expect(page).to have_content('Ed played rock')
+  end
+
+  scenario 'RPS-Bot has their choice displayed (single player)' do
+    srand(1)
```
lindseydew: Great use of a seed value here.

spec/features/two_player_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  scenario 'there is a button to begin the game' do
+    two_player_game
+    expect(page).to have_button('Start the game')
+  end
+
+  scenario 'clicking button takes you to two player game page' do
+    start_two_player_game
+    expect(page).to have_content("Choose your weapon")
+  end
+end
```
lindseydew: Very thorough feature testing - well done

PR #1305, opened 2019-06-02
Completed RPS challenge

kks110:


kks110:
Needs some refactoring to cut down on repetition.

lib/game_logic.rb
```diff
@@ -0,0 +1,48 @@
...
+      player1_choice = choices[0]
+      player2_choice = choices[1]
+    end
+
+    return @win = 'draw' if player1_choice == player2_choice
+
+    case player1_choice
+    when 'rock'
+      return @win = true if player2_choice == 'scissors'
+      return @win = true if player2_choice == 'lizard'
```
lindseydew: Nice use of a case statement.

spec/game_one_player_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+  it 'can store a player and return it' do
+    expect(game.player1).to eq(player)
+  end
+
+  it 'can tell you if its a 1 player game or 2 player' do
+    expect(game.players).to eq(1)
+  end
+
+  it 'can randomly pick a weapon' do
+    expect([:rock, :paper, :scissors, :lizard, :spock]).to include(game.pick_weapon)
```
lindseydew: Can use `srand` to test values that are generated by the computer

lib/game_two_player.rb
```diff
@@ -0,0 +1,28 @@
...
+  include GameLogic
+
+  PLAYERS = 2
+
+  attr_reader :player1, :player2, :win
+
+  def initialize(name1, name2, player_class = Player)
+    @player1 = player_class.new(name1)
+    @player2 = player_class.new(name2)
+  end
```
lindseydew: Nice use of dependency inversion.

lib/game_logic.rb
```diff
@@ -0,0 +1,48 @@
+module GameLogic
+
+  def pick_weapon
+    [:rock, :paper, :scissors, :lizard, :spock].sample
+  end
+
+  def determine_result(choices = [])
```
lindseydew: This method could take in two arguments rather than an array, since the implementation assumes an array of size 2 is being passed to the method.

PR #1306, opened 2019-06-02
Most beautiful RPS app you have seen

anthony-hull:
Guaranteed to remind you of the benefits of 1990 web design


lib/game.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def who_wins?
+    return DRAW if a_draw?
+    calculate_choices
+    winner_integer
+  end
+
+  def a_draw?
+   true if @move == @ai_move
```
lindseydew: This can be simplified to 
` @move == @ai_move ` (it will resolve to true if this equality is fulfilled)

spec/features/PlayRPS_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  end
+  scenario 'select a move option and page displays your choice' do
+    play_game('scissors')
+    expect(page).to have_content('Anthony chose scissors') 
+  end
+
+  scenario 'AI move is displayed' do
+    srand(4)
+    play_game('scissors')
```
lindseydew: Nice use of a seed here.

spec/system/game_spec.rb
```diff
@@ -0,0 +1,42 @@
+require './lib/game.rb'
+describe Game do
+  
+  let(:player1) {spy(:player1, name: 'Anthony')}
+  it 'returns player username' do
+    srand (7)
```
lindseydew: Is this seed setting the state of Game for later in the tests? If so, it should probably go in the overall set up. Or ideally, tests which rely on a certain state should have that state set in that particular test.

lib/game.rb
anthony-hull: True! This was picked up on in our code review as well.

PR #1307, opened 2019-06-02
RPS challenge submission

LaurenOM:


lib/game.rb
```diff
@@ -0,0 +1,40 @@
...
+  def add_player(name, player = Player)
+    @players << player.new(name)
+  end
+
+  def player_name(player_no)
+    @players[player_no - 1].name
+  end 
+
+  def computer_move
+    ["Rock", "Paper", "Scissors"].sample
```
lindseydew: Computer move could be an instance variable of your game class

spec/game_spec.rb
```diff
@@ -0,0 +1,53 @@
+
+require 'game'
+
+describe Game do 
+  let(:game) { Game.instance }
+  let(:player_class) { spy(:player_class, :new => player) }
+  let(:player) { spy(:player, :name => "Lauren") } 
+  
+  it 'chooses rock paper scissors randomly' do
+    expect(['Rock', 'Paper', 'Scissors']).to include(game.computer_move)
```
lindseydew: Could use `srand` here to determine the next call of sample and test each of the three options are the only outputs of that array.

app.rb
```diff
@@ -0,0 +1,51 @@
...
+  get '/result' do 
+    @move = session[:move]
+    @move2 = session[:move2]
+    erb(:result)
+  end
+
+  post '/new-game' do 
+    session.clear
+    redirect '/'
+  end
```
lindseydew: Nice idea to be able to refresh the game and start again

lib/game.rb
```diff
@@ -0,0 +1,40 @@
...
+  def self.instance
+    @game ||= Game.new
+  end
+
+  def add_player(name, player = Player)
+    @players << player.new(name)
+  end
+
+  def player_name(player_no)
+    @players[player_no - 1].name
```
lindseydew: Having external clients accessing the contents of an array using the index could lead to trouble if the array length is less than the index digit supplied. If you want to be able to retrieve an item from a data structure, my suggestion would be to use a hash map so that the client method can call this with a key and not have to know about the length of the array.

views/result.erb
```diff
@@ -0,0 +1,25 @@
...
+<div>
+<%= @game.check_winner(@move, @move2)%>
+<br>
+</div>
+<div class="button-container">
+<br>
+<form action='/new-game' method='post'> 
+<div><button type="submit">New game</button></div>
+</form>
+<form action='/turn' method='get'>
```
lindseydew: This can be achieved with a hyper link also:
`<a href="/turn">Rematch </a>`

PR #1311, opened 2019-06-03
RPS challenge

Ma9sha:
I spent most of my time fixing the logic and not enough on Tests, I have few more test to add with regards to:
when Computer wins
When Player wins
and when they Draw.


views/welcome.erb
```diff
@@ -0,0 +1,20 @@
...
+<head>
+<meta charset="utf-8">
+<link rel="stylesheet" type="text/css" href="style.css">
+</head>
+
+  <body>
+  <div>
+  <h2>Hello <%= @playername %> </h2>
+  <p><h3>Welcome to rock paper scissor game</h3></p>
+
```
lindseydew: Great implementation of the form well done

spec/unit/game_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'game'
+
+describe ComputerGame do
+
+  it 'gives a random choice of rock, paper, scissor' do
+    choices = ['rock', 'paper', 'scissor']
```
lindseydew: I can see why you've implemented the test in this way. 

We could make an even stronger assertion by using `srand` (which is a seed) this will determine the next value of calling:
``` ['rock', 'paper', 'scissor'].sample ```
So by testing each seed which returns a distinct value we can assert that our method only returns one of these values too.

File: spec/features/web_helpers.rb [hunk 3]
```diff
@@ -0,0 +1,7 @@
+def sign_in_and_play
+
+  visit '/enter_name'
+  fill_in :player_name, with: 'Manisha'
+  click_button "Sign in"
+
```
lindseydew: Nice abstraction here.

public/style.css
```diff
@@ -0,0 +1,10 @@
+
+div {
+  border: 1px solid black;
+  margin: 75px 300px 75px;
+  padding: 50px 75px 75px 50px;    
+  background-color: lightblue;
+  font-family:    Arial, Helvetica, sans-serif
+  font-size:      40px;
+  font-weight:    bold;
```
lindseydew: Nice use of css

spec/unit/game_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+    game = ComputerGame.new
+    game.choice
+    expect(choices).to include(game.choice)
+  end
+
```
lindseydew: It would be nice to have a test on the `outcome` method

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+    elsif (@choice == 'paper' && @user_choice == 'rock')
+      com_result
+    elsif (@choice == 'scissor' && @user_choice == 'rock')
+      play_result
+    elsif (@choice == 'rock' && @user_choice == 'paper')
+      play_result
+    elsif (@choice == 'paper' && @user_choice == 'scissor')
+      play_result
+    else
+      draw
```
lindseydew: This draw method could be a conditional expressed like this:
``` 
if(@choice == @user_choice) 
  "its a draw"
```

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+  end
+
+  def com_result
```
lindseydew: I think these methods should be variables because there are just returning the same string

lib/game.rb
```diff
@@ -0,0 +1,42 @@
+class ComputerGame
+  attr_reader :choice, :user_choice
+
+  def initialize
```
lindseydew: The user choice could be set at the initialise method eg:

```  
    def initialize(user_choice)
        @user_choice = user_choice
```

This ensures that the variable is set on creation of the object. If the `player` method is never called then the `@user_choice` variable would be Nil and the behaviour of the object does not make sense

app.rb
```diff
@@ -0,0 +1,26 @@
+require 'sinatra/base'
+require './lib/game'
+
+
+class Rpschallenge < Sinatra::Application
+
+get '/enter_name' do
+  erb(:index)
+end
+
```
lindseydew: Great use of `get` and `post`

PR #1315, opened 2019-06-03
Adrian's RPS Friday Challenge

adrem7:


app.rb
```diff
@@ -0,0 +1,51 @@
...
+    erb(:player_1_move)
+  end
+
+  post "/player_1_moved" do
+    @game = $game
+    @game.assign(@game.player_1, params['player_1_move'])
+    redirect '/player_2_move'
+  end
+
+  get "/player_2_move" do
```
lindseydew: This can probably be the same request as `player_1_move` by making the player name a variable for the template to read

lindseydew: The browser can send which player it is as a parameter to the post request too.

app.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  get "/" do
+    erb(:index)
+  end
+
+  post "/names" do
+    player_1 = Player.new(params['player_1'])
+    player_2 = Player.new(params['player_2'])
+    $game = Game.new(player_1, player_2)
+    redirect '/play'
```
lindseydew: Great use of a post with redirect to a get request

lib/game.rb
```diff
@@ -0,0 +1,53 @@
+class Game
+
+  def initialize(player_1, player_2)
+    @players = [player_1, player_2]
```
lindseydew: Not sure about assigning this to an array of size two. Feel that these could just be two separate variables

spec/system/game_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'game'
+
+describe Game do
+  context 'players draw' do
+    let(:player1)   { double(:player_1, move: "Rock") }
```
lindseydew: Nice use of doubles here - well done!

PR #1316, opened 2019-06-14
James Hylins RPS Challenge Solution

zepherine2006DEV:


README.md
```diff
@@ -27,60 +21,28 @@ As a marketeer
...
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
+* Write a draft of this approach, so that I don't rush into development!
+* Create a basic working environment based on a simple feature test.
+  * Bundle; Pull request; first commit
+  * Configure rspec (no examples), rack config, coverage tool.
+  * Create feature test for "Hello Player!" and get to pass.
+* Read the use cases carefully, so that I don't miss anything.
+* Create a sketch of the screens and classes (see below)
```
LucyMHall: These are really helpful and show a clear focus on the user's experience.

app.rb
```diff
@@ -0,0 +1,40 @@
+require 'sinatra/base'
```
LucyMHall: General Note: your file structure is very neat, and file naming is consistent.

spec/spec_helper.rb
```diff
@@ -1,6 +1,14 @@
+ENV['RACK_ENV'] = 'test'
```
LucyMHall: All the initial setup seems good

spec/features/play_game_spec.rb
```diff
@@ -0,0 +1,14 @@
+# As a marketeer
+# So that I can enjoy myself away from the daily grind
+# I would like to be able to play rock/paper/scissors
```
LucyMHall: Although I understand what this test means, I think that the expectation for the page doesn't link up with what you are testing e.g. having your name returned to you once you have clicked on rock, paper or scissors does not indicate you have played the game. Maybe include a confirmation line where the choice is repeated back to the player.

spec/features/play_game_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+feature 'Play round of paper, sissors, stone' do
+  scenario 'Player chooses paper and plays game' do
+    visit('/')
+    fill_in('player_name', with: 'James')
+    click_button('submit')
+    choose('Paper')
+    click_button('submit')
+    expect(page).to have_content("James")
+  end
+end
```
LucyMHall: I think you might be missing a results feature test where the computer's choice are displayed along with the overall results. According to the code rubric you need to test for a win, loss and draw outcome.

spec/game_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'game'
+
+describe Game do
+  describe '#get_result' do
+    before(:each) do
+      @weapon = double("weapon")
+      @player = double("player")
+    end
+    
+    it 'should return player win if player has paper and game has rock' do
```
LucyMHall: You may not need to test all of these different scenarios. I think you just need to test that the computer can go into your hash and pull out the relevant information in general, although I do like the thoroughness.

spec/random_weapon_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'random_weapon'
+
+describe RandomWeapon do
+  describe '#weapon' do
+    it 'returns rock, paper or scissors' do
+      expect(subject.weapon).to eq("rock").or eq("paper").or eq("scissors")
```
LucyMHall: I haven't seen this syntax before but i like it.

views/index.erb
```diff
@@ -0,0 +1,12 @@
+<h1>
+Welcome to Rock, Paper, Sissors!
```
LucyMHall: Typo :)

lib/random_weapon.rb
```diff
@@ -0,0 +1,7 @@
+class RandomWeapon
```
LucyMHall: I don't know if this needs to be its own class, but I think that due to the focus on the single responsibility principle it makes sense

lib/player.rb
```diff
@@ -0,0 +1,11 @@
+class Player
```
LucyMHall: I understand how the choice method works here but it might be better not to have it as an attribute_accessor because you don't want someone to be able to manually change it after the choice has been made.

lib/game.rb
```diff
@@ -0,0 +1,43 @@
+class Game
+  
+  PAPER = 'paper'
+  SCISSORS = 'scissors'
+  ROCK = 'rock'
+
+  RULES = [
+    { :player => PAPER, :game => PAPER, :result => "draw" },
```
LucyMHall: It says in the code rubric that it is better if the model doesn't return a string as this is more display logic which would go in the view

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+    { :player => ROCK, :game => PAPER, :result => "lose" },
+    { :player => SCISSORS, :game => PAPER, :result => "win" },
+    { :player => SCISSORS, :game => ROCK, :result => "lose" }
+  ]
+
+  attr_reader :game_choice
+  
+  def initialize(player, random_weapon)
+    @player = player
+    @random_weapon = random_weapon
```
LucyMHall: So do you pass a new instance of the random_weapon class to the game and then run the method weapon on it?

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+    { :player => PAPER, :game => SCISSORS, :result => "lose" },
+    { :player => ROCK, :game => SCISSORS, :result => "win" },
+    { :player => ROCK, :game => PAPER, :result => "lose" },
+    { :player => SCISSORS, :game => PAPER, :result => "win" },
+    { :player => SCISSORS, :game => ROCK, :result => "lose" }
+  ]
+
+  attr_reader :game_choice
+  
+  def initialize(player, random_weapon)
```
LucyMHall: Good dependency injections

lib/game.rb
```diff
@@ -0,0 +1,43 @@
+class Game
```
LucyMHall: It is suggested in the code rubric that the computer is written as another instance of the player class, I think this might make this class a bit easier to read as at the moment it is having to handle all the rules and simulating the computer's go

lib/random_weapon.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  def weapon
```
LucyMHall: It is suggested in the code rubric that you save weapons as symbols rather than strings as they are immutable objects which mean the same thing whenever you use them, WEAPONS = [:rock, :paper, :scissors]

app.rb
```diff
@@ -0,0 +1,40 @@
...
+require_relative './lib/player'
+require_relative './lib/game'
+require_relative './lib/random_weapon'
+
+class RPSGame < Sinatra::Base
+
+  enable :sessions
```
LucyMHall: Great that you didn't use global variables!

app.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  get '/' do
+    erb :index
+  end
+
+  post '/register' do
```
LucyMHall: Nice post - redirect - get pattern being used, where your post routes are only handling the information you have sent it, not dealing with anything new

app.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  get '/turn' do
+    @player_name = session[:player].name
+    erb :turn
+  end
+
+  post '/play' do
+    @choice = params[:choice]
+    @player = session[:player]
+    @player.choice = @choice
```
LucyMHall: I'd recommend shortening this to @player.choice = params[:choice] rather than separating this into two different lines

app.rb
```diff
@@ -0,0 +1,40 @@
...
+    redirect "/result"
+  end
+
+  get '/result' do
+    @player_name = session[:player].name
```
LucyMHall: Because you have passed the instance of your player to your game instance when you make a new game, and save it as an attribute, it might be a good idea to refer to your player via your game class e.g. @game.player_1. This way at the beginning of each route you only need to tell your view about the instance of your game class

app.rb
```diff
@@ -0,0 +1,40 @@
...
+    @player = session[:player]
+    @game = Game.new(@player, RandomWeapon.new)
```
LucyMHall: It seems strange that a game instance is only being made now when you were technically playing the game since the turn route

views/turn.erb
```diff
@@ -0,0 +1,15 @@
+<h1>
```
LucyMHall: General note about your views: all your views are really nice and they don't contain any unnecessary logic!

PR #1317, opened 2019-06-15
Lucy Hall Pull Request

LucyMHall:


lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  private
+
+  def draw?
+    @player_1.choice == @player_2.choice
+  end
+
+  def winning_scenarios
```
zepherine2006DEV: perhaps the method name could be more verby. e.g. find_winning_scenario?

PR #1318, opened 2019-06-16
AJ RPS One Player 

ajosephides:


lib/app.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  post '/name' do
+    session[:game].create_player(params[:player_name])
+    redirect '/choose-move'
+  end
+
+  get '/choose-move' do
+    @game = session[:game]
+    erb(:choose_move)
```
ruthmoog: You could consider making the app file more consistent by using either
 - `erb(:page)` and `redirect to('/page')`
or
 - `erb :page` and `redirect '/page'`

lib/game.rb
```diff
@@ -0,0 +1,22 @@
...
+require_relative './ai.rb'
+require_relative './winner.rb'
+
+class Game
+
+  attr_reader :player, :player2, :winner
+
+  def initialize(player_class = Player, player2 = Ai)
+    @player_class = player_class
+    @player2 = player2.new
```
ruthmoog: We talking about the pros/cons of creating `Ai.new` in the argument params vs when assigning the player2 instance variable.

lib/player.rb
```diff
@@ -0,0 +1,15 @@
+class Player
+
+  attr_reader :name, :move
+
+  def initialize(name)
+    @name = name
+  end
+
+  def save_move(move)
+    raise 'Not a valid choice' unless ['Rock', 'Paper', 'Scissors'].include?(move)
```
ruthmoog: rubocop flagged this line as a bit long 👮 Consider refactoring; I reckon if you named the array with a nice descriptive name you can make this line more readable 👌

lib/app.rb
ajosephides: Yes! Thank you

lib/player.rb
ajosephides: Top suggestion - thank you

lib/app.rb
ajosephides: All done.
👍

lib/game.rb
ajosephides: 👍

lib/player.rb
ajosephides: I created a constant of valid moves and replaced it here.

PR #1324, opened 2019-06-16
Ruth Baker Rock, Paper, Scissors

ruthmoog:
I made a game that works!  but the results page is meh so you'll have to trust me ;)
need to update the README but it's well past my bed time. 💤 

🚀 🔖  ✂️


app.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class RockPaperScissors < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/names' do
+    session[:player_1] = params[:player_1]
```
ajosephides: Player1 either as a session variable or as an instance variable appear on a couple of the pages. Game is not initialised until the end. This means the controller has two object its is handling and in a larger application can begin to 'bloat'. 
https://blog.makersacademy.com/forget-fat-models-its-time-for-skinny-controllers-and-skinny-models-a9b84ec481b7

app.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  post '/referee' do
+    @game = Game.new(params[:player_pick].downcase.to_sym)
+    p params[:player_pick]
+    session[:winner] = @game.calculate_winner
+    redirect to('/result')
+  end
+
+  get '/result' do
+    @winner = session[:winner]
```
ajosephides: Having the players and computer as instance variables on the Game object would allow for their name and move to be passed into the page and thus presented to the user.

app.rb
```diff
@@ -0,0 +1,36 @@
...
+    erb(:result)
+  end
+
+  # this will start the server automatically if the app.rb file is run in the
+  # command line with `ruby app.rb` :)
+  run! if app_file == $0
+end
```
ajosephides: Nice, easy to follow user journey through the game.

lib/game.rb
```diff
@@ -0,0 +1,31 @@
+class Game
```
ajosephides: Game appears to be handling the logic for more than one thing.
It is generating the computer's choice, it is storing the player's choice and its holding the game logic and deciding the winner. This could be broken up into different classes e.g. computer is its own class like a player

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+      winner = "it's a draw!"
+    elsif player_1_win?
+      winner = "player wins!"
+    else
+      winner = "computer wins!"
+    end
+    winner
+  end
+
+  def player_1_win?
```
ajosephides: This is a little tricky to read as it is one function but spans across 3 lines. 
Can it be broken up? A private helper function might be of use here.

spec/features/feature_spec.rb
```diff
@@ -0,0 +1,33 @@
+# As a marketeer
+# So that I can see my name in lights
+# I would like to register my name before playing an online game
+
+# As a marketeer
+# So that I can enjoy myself away from the daily grind
+# I would like to be able to play rock/paper/scissors
+
+describe "Play an online game of Rock Paper Scissors", type: :feature do
```
ajosephides: Cool. All seems to be tested. Couple of points..
In Capybara you can use `feature` instead of `describe` and `scenario` instead of `context` all seems to work regardless but I think it shows more explicitly to the user what framework you are using.
I like how the features test individual segments of the game and then culminate in the full game.

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+    @player_pick == :rock && @computer_pick == :scissors ||
+    @player_pick == :paper && @computer_pick == :rock ||
+    @player_pick == :scissors && @computer_pick == :paper
+  end
+
+end
+
+class Computer
```
ajosephides: I didn't see this earlier - I see computer is a separate class. This should be in a separate file and tested in isolation

spec/game_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        allow(computer).to receive(:random) { :rock }
+
+        expect(game.calculate_winner).to eq("it's a draw!")
+      end
+    end
+  end
+
+end
+
+describe Computer do
```
ajosephides: I think anything to do with Computer should be in its own unit test file.

spec/game_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'game'
+
+RSpec.describe "Game" do
+
+  context "when player selects rock" do
```
ajosephides: Because of the nature of the logic in your `winner` method I think, strictly speaking, all options should be tested. Only the user choice of `rock` is tested here. How do you know that no typos, logic error might exist if the user chose paper?

views/result.erb
```diff
@@ -0,0 +1 @@
+<h1><%= @winner %></h1>
```
ajosephides: As a user it would have been good to see the results of both competitors here so that I can truly believe the result.

app.rb
ruthmoog: 🙏

PR #1327, opened 2019-06-17
Rhys RPS challenge

rhysco8:


app.rb
```diff
@@ -0,0 +1,35 @@
...
+class RpsApp < Sinatra::Base
+
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    $player = Player.new(params[:player_name])
```
aniasobo: obligatory "tut tut, global variables"

app.rb
```diff
@@ -0,0 +1,35 @@
...
+    @player_name = $player.name
+    @player_move = $player.move
+    @computer_move = $computer.move
+    @outcome = Game.new.outcome(@player_move, @computer_move)
+    erb :play
+  end
+
+  post '/move' do
+    $player.choose(params[:move])
+    computer_move = ["Rock", "Paper", "Scissors"].sample
```
aniasobo: not sure if your logic works, but keeping the moves/weapons in a constant is the preferred solution

aniasobo: ah ok, it's actually bad practice to store the moves in the controller. refactor to store them in a constant in the appropriate class?

PR #1328, opened 2019-06-17
RPS challenge - User vs Computer

Danielandro:
Not attractive, but it does work and test coverage is good


lib/game.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  private
+
+  def draw?
+    @player.weapon == @computer.weapon
+  end
+
+  def player_wins?
+    @player.weapon == "Rock" && @computer.weapon == "Scissors" ||
```
monikakaczan: Really nice and clean solution for conditionals!

spec/features/name_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature "Enter player's name" do
+  scenario "Player registers name before playing game" do
+    register_name_and_play
```
monikakaczan: good usage of a method in order to not repeat yourself

spec/units/game_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'game'
+
+describe Game do
+  subject(:game) { Game.new(player, computer) }
+  let(:player) { double :player }
+  let(:computer) { double :computer }
+
```
monikakaczan: A big plus for using doubles!

PR #1329, opened 2019-06-17
Work very much in progress

aniasobo:


lib/rps-controller.rb
```diff
@@ -0,0 +1,12 @@
+require 'sinatra/base'
```
rhysco8: Why did you choose to have the controller and views in the lib folder?

config.ru
```diff
@@ -0,0 +1,2 @@
+require_relative 'lib/rps-controller.rb'
+run Rps
```
rhysco8: Could run rubocop to pick up small linting issue

lib/rps-controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    redirect '/game'
+  end
+
+  get '/game' do
+    @playername = session[:playername]
+    erb :game
+  end
+
+  run! if app_file == $0
+end
```
rhysco8: Could run rubocop to pick up small linting issue

PR #1343, opened 2019-07-14
Will's RPS game. It don't look pretty but it works (hopefully!)

willhowes:


lib/game.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def select_weapon(weapon)
+    @player_weapon = weapon
+  end
+
+  def result
+    OUTCOMES[@player_weapon.to_sym][@computer_weapon]
+  end
+
+  def confirm_winner
```
loicdewit: Could use a case statement to make it more concise, although it's a detail.

lib/random_weapon.rb
```diff
@@ -0,0 +1,12 @@
+class RandomWeapon
+  def weapon
+    weapons = [:rock, :paper, :scissors]
+    weapons[random_num]
+  end
+
```
loicdewit: Could use the sample method. This would allow you to reduce this file to two lines.

views/index.erb
```diff
@@ -0,0 +1,8 @@
+<h1>ROCK PAPER SCISSORS!</h1>
+<h2>Sign in to play</h2>
+<form action="/names" method="post">
+  <label for="player_name">
```
loicdewit: The comments I am making are quite trivial/about details but that's because overall the code is very readable. 

Having said that, I think it is good practice to close the HTML elements you create with closing tags. For instance, in your label: </label>

views/result.erb
```diff
@@ -0,0 +1,7 @@
+<h1><%=@game.confirm_winner%></h1>
```
loicdewit: Could have set to a variable and then call the variable using ERB.

PR #1344, opened 2019-07-14
Rianne's RPS game 

riannemcc:


lib/game.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def self.instance
+    @game
+  end
+
+  def result
+    computer_turn = @computer.computer_turn
+    player_turn = @player.chosen_item
+    return "It's a draw!" if player_turn == computer_turn
```
ted-scanlan: could be made even tidier if you make the conditionals their own method e.g. return "its a draw" if draw?   (with draw being another method)

app.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  post '/name' do
+    player = Player.new(params[:player_name])
+    @game = Game.create(player, Computer.new)
+    redirect '/play'
+  end
+
+  before do
+    @game = Game.instance
```
ted-scanlan: I dont quite understand what this bit is doing. why does game need to create an instance of itself when youve already created a game in the 'post' stage?

app.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post '/game' do
+    item = params[:item].to_sym
```
ted-scanlan: This could be wrong, but should'nt something like converting a paramter to a symbol be done in the model, as it requires ruby logic?

PR #1346, opened 2019-07-14
RPS World Championship 

ckettell:


lib/rps_calculator.rb
```diff
@@ -0,0 +1,52 @@
...
+
+class RpsCalculator
+
+  attr_reader :choice, :name, :rival
+
+  def initialize(choice, name = "Gladiator")
+    @choice = choice
+    @name = name
+    @rival
+    @rules = {
```
hemser1: This seems heavy on the initialize method, could this get broken into another method?

lib/rps_calculator.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+    end
+
+    private
+
+    def result
+      @rules["#{@choice.downcase}/#{opponent}"]
+    end
+
+    def opponent
```
hemser1: could this be broken into a new class inline with SRP?

spec/features/greeting_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'capybara/rspec'
+
+feature 'Opening page' do
+  scenario 'The opening page asks the user to enter their name and weapon choice' do
+    visit ('/')
+    expect(page).to have_content "Welcome, Gladiator. Enter your name and choose your weapon"
+  end
+end
+
+feature 'The battle page' do
```
hemser1: With rubocop, should we remove all hashed out code before pushing? (I don't know the answer I am just saying! haha)

PR #1356, opened 2019-07-15
my version of the rps challenge

richieganney:


app.rb
```diff
@@ -0,0 +1,37 @@
...
+    session[:player_1] = params[:player_1]
+    redirect '/game'
+  end
+  
+  get '/game' do
+    @player_1 = session[:player_1]
+    erb :game
+  end
+
+  get '/rock' do
```
TBrockster: I think this should be done with only 1 route, and the logic pushed down to the model.

lib/game.rb
```diff
@@ -0,0 +1,41 @@
+class Game
+  
+  attr_reader :computer
+
+  def initialize
+    @computer = ["rock", "paper", "scissors"].sample
+  end
+
+  def play(choice)
```
TBrockster: This can be done in a slightly 'cleverer' way - https://stackoverflow.com/questions/26393870/ruby-rock-paper-scissor-game-using-while-if-else-statements-player-vs-comput

views/game.erb
```diff
@@ -0,0 +1,15 @@
+<html>
+    <body style="background-color:rgb(66, 190, 194)">
```
TBrockster: I like the CSS.

PR #1358, opened 2019-07-15
bit late...sorry.

hjdr:


app/controllers/application_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    set :views, "app/views"
+    set :public_dir, "public"
+  end
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/name' do
+    session[:name] = Player.new(params[:name])
```
gennyallcroft: is it ok to use session for your variables?

app/controllers/application_controller.rb
```diff
@@ -0,0 +1,48 @@
+require 'sinatra'
+
+class ApplicationController < Sinatra::Base
+
+  configure do
```
gennyallcroft: what does this bit do?

app/controllers/application_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  get '/game' do
+    @player = session[:name].name
+    erb(:game)
+  end
+
+  post '/compare-result' do
+    @result = RpsGame.new.result(params[:player_hand])
+    if @result == 'win'
+      redirect '/result-win'
```
gennyallcroft: could use a message instead?

spec/features/rps_gameplay_spec.rb
```diff
@@ -0,0 +1,23 @@
+
+feature "RPS gameplay" do
+  scenario "when the player selects rock and opponent selects scissors" do 
+    enter_name_and_play
```
gennyallcroft: love this!

app/models/rps_game.rb
```diff
@@ -0,0 +1,21 @@
+
+class RpsGame
+
+  def result(player_hand)
+    compare_hands(player_hand, random_hand)
```
gennyallcroft: could store the player hand as a variable in the Player class

PR #1378, opened 2019-08-27
website

claredowson:


app.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  get '/play' do
+    @player_1_name = session[:player_1_name]
+    @player_2_name = session[:player_2_name]
+    erb :play
+  end
+
+  post '/names' do
+    session[:player_1_name] = params[:player_1_name]
```
lindseydew: Awesome! I can see that you are reading the params for the form and setting them as a session variable - great job! 

Where does the page go next? It is being redirected to `\action`, but looking at the code it looks like your intent was to perhaps to go `\play` where you read the player variables names out?

app.rb
```diff
@@ -0,0 +1,52 @@
...
+  get '/action' do
+    @action = session[:rock]
+    @action = session[:paper]
+    @action = session[:scissors]
+    erb :action
+  end
+
+  post '/action' do
+    session[:rock] = params[:rock]
+    session[:paper] = params[:paper]
```
lindseydew: Are you able to read these params ok? I think there could be something missing in the form definition perhaps?

PR #1380, opened 2019-08-27
Eloise rps challenge

EloisePrasad:


app.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  post '/' do
+    @player1 = params[:player1]
+    redirect '/play'
+  end
+
+  post '/play' do
+    @player1 = params[:player1]
+    erb(:play)
```
lindseydew: I can see what you've done here will work - well done. \

It's common practise for the response of `posts` requests to `redirect` to a get request because then users are able to refresh of revisit that page.

If you need to share data across controllers you can make use of the `sessions` object in ruby

views/play.erb
```diff
@@ -0,0 +1,13 @@
+<%= "Take your action #{@player1}" %>
+
```
lindseydew: Nice! This can be refactored a little so that you don't have to define a new endpoint for each option. You could make use of an option dropdown: https://www.w3schools.com/html/html_form_elements.asp

spec/features/play_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    visit '/'
+    fill_in :player1, with: 'Eloise'
+    click_button 'Submit'
+    click_button 'Scissors'
+    expect(page).to have_content 'You picked Scissors'
+  end
+
+# Wasn't sure how to determine value for randomisation of the computer pick
+  # scenario 'it can give the computer choice' do
+  #   visit '/'
```
lindseydew: Great question! Can make use of a `seed` here to make the output deterministic. Take a look at this project: https://github.com/makersacademy/skills-workshops/tree/master/week-3/debugging_2

views/option.erb
```diff
@@ -0,0 +1,8 @@
+<%= "You picked #{@choice}" %>
+<br>
+<%= "The computer picked #{@computer_choice}" %>
```
lindseydew: Nice! Could have a method in the game class which computes the result?

PR #1416, opened 2019-10-06
Kevin's .... time sink

kevindhanna:


kevindhanna:
This test works on my laptop when I run `rspec`, but not when running `bundle exec rspec`

PR #1438, opened 2019-10-18
Will Bath RPS Challenge

Will-Bath:


lib/opponent.rb
```diff
@@ -0,0 +1,56 @@
...
+      return @who_wins
+    elsif @player_weapon == 'scissors' && @opponent_move == 'paper'
+      @who_wins = 'You win!'
+      return @who_wins
+    elsif @player_weapon == 'scissors' && @opponent_move == 'scissors'
+      @who_wins = "It's a tie!"
+      return @who_wins
+    end
+
+  end
```
Will-Bath: I'm sure that this method could be done much more efficiently, but it was like this for now as I was low on time to test the functionality. When I have the program fully working, I could refactor this method to work better in a smaller amount of space.

lib/result.rb
```diff
@@ -0,0 +1,7 @@
+# class Result
```
Will-Bath: I didn't end up using this method so it would be better practice to remove it from the pull request.

spec/features/enter_name_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'Enter name' do
```
Will-Bath: I was happy with the feature tests I wrote for this project as they explain what function they're testing and are easy to follow.

spec/features/select_weapon_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    visit('/')
+    fill_in :player_name, with: 'Will'
+    click_button('Prepare to battle!')
+    expect(page).to have_content 'Rock, Paper, or Scissors'
+  end
+
+  scenario 'Player can select either Rock, Paper, or Scissors' do
+    visit('/')
+    fill_in :player_name, with: 'Will'
+    click_button('Prepare to battle!')
```
Will-Bath: This series of Capybara actions here is repeated a lot, so maybe I could implement a web helper or module that runs this block every time in multiple places.

joshblackmoor: Yes I agree, you could DRY out your tests by creating another file with a method that when called it would perform the sign in to the site which is performed on every test. When you do this make sure you add the ruby file with the sign in method to the spec so every test file is able to access this.

spec/spec_helper.rb
```diff
@@ -2,6 +2,18 @@
 require 'simplecov'
 require 'simplecov-console'
 
+ENV['RACK_ENV'] = 'test'
+
+# require our Sinatra app file
+require File.join(File.dirname(__FILE__), '..', 'app.rb')
```
jamesjoshuahill: Unfortunately, if you require you app code before `SimpleCov.start` the test coverage report will not be accurate. As it is the report indicates 100%, but it's actually 90%:
```ruby
SimpleCov.start

# require app after starting SimpleCov
require File.join(File.dirname(__FILE__), '..', './lib/app.rb')
Capybara.app = Battle
```

PR #1439, opened 2019-10-18
My Pull Request

Adam1824:


spec/features/display_result_spec.rb
```diff
@@ -0,0 +1,46 @@
+feature 'Outcome of the game' do
+  scenario 'player 1 winning' do
+    visit '/'
+    loop do
+      fill_in :player_1_name, with: 'Adam F'
+      click_button 'Submit'
+      fill_in :answer, with: 'Rock'
+      click_button 'Submit'
```
josiepollard: This was repeated in a few of the loops, try and put them all together in another file and then call it. this is to make it more efficient

app.rb
```diff
@@ -8,11 +9,23 @@ class Rps < Sinatra::Base
   end
 
   post '/names' do
-    $player_1 = Player.new(params[:player_1_name])
+    $game = Game.new(Player.new(params[:player_1_name]))
```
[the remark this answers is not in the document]
josiepollard: instead of making a global variable, make a class method where you pass in the players as a object. in that method, create an instance variable where in that you create a new game object while passing in the player. then make another class method where you return the instance variable just made in the perilous class method. in app.rb, create an instance variable which calls the first class method and then on another line call the second class method.

spec/features/display_result_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      if page.has_text? "Adam F won!"
+        break
+      end
+      click_button 'Home'
+    end
+    expect(page).to have_content ("Adam F won!")
```
jamesjoshuahill: At the end of the game it would be great to see a summary with what the computer played. For example:
```
Josh played Rock
Computer played Paper

Computer won!
```

spec/game_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      game.generate_computer_answer
+      expect(['Rock', 'Paper', 'Scissors']).to include(game.computer_answer)
+    end
+  end
+
+  describe '#outcome' do
+    context 'can work out who wins' do
+      it 'player can win' do
+        game.store_answer('Paper')
+        game.computer_answer = 'Rock'
```
jamesjoshuahill: It looks like you had to expose `computer_answer` as a `attr_accessor`. How else could you control the randomness in these tests?

PR #1440, opened 2019-10-18
pull request

josiepollard:


views/playername.erb
```diff
@@ -0,0 +1,14 @@
+
+<h1>Hello <%= @name %> </h1>
```
josiepollard: maybe add a bit to tell the player to select rock paper scissors

views/playername.erb
```diff
@@ -0,0 +1,14 @@
...
+
+<div>
+    <form method="post">
+      <input formaction="/rock" type="submit" value="Rock"/>
```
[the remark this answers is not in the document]
Adam1824: 'formaction' is two words - form, meaning you can submit an input to the website and action meaning which route to go to next

views/playername.erb
Adam1824: yeh you can either had a second heading doing <h2> or add regular text <p>.

Adam1824: I agree, you can add some text either in a seconding heading by doing '<h2></h2>' or you can just add some text by doing '<p></p>'

views/playername.erb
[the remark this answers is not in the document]
jamesjoshuahill: I've never used the `formaction` attribute in a submit tag before 👀 Thank you for bringing it to my attention 😄 

I found this in the MDN documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit#formaction.

Looks like it's a way for a submit button to override the `action` attribute of a form tag. For example:
```html
<form method="post" action="/rock" ...
```

PR #1441, opened 2019-10-18
rps

zero201926:


zero201926:
I ran out of time to do all Rspec

config.ru
```diff
@@ -0,0 +1,4 @@
+require 'rubygems'
+require File.join(File.dirname(__FILE__), './lib/app.rb')
```
zero201926: Not quite sure what this line does

spec/rps_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+    it 'when "paper" entered goes to paper method' do
+      allow(paper).to receive(:choice).and_return "paper"
+      paper.paper
+    end
+    it 'when "scissors" entered goes to scissors method' do
+      allow(scissors).to receive(:choice).and_return "scissors"
+      scissors.scissors
+    end
+  end
```
zero201926: This is incomplete, not sure how to tests the corresponding code

lib/views/style_helper.erb
```diff
@@ -0,0 +1,51 @@
...
+  text-align: center;
+}
+
+.block:hover {
+  background-color: #ddd;
+  color: black;
+}
+</style>
+
+<% end %>
```
zero201926: All of this file was me trying to DRY my css, but I didn't get it to work.

lib/rps.rb
```diff
@@ -0,0 +1,53 @@
...
+    case choice
+    when "rock"
+      rock
+    when "paper"
+      paper
+    when "scissors"
+      scissors
+    end
+  end
+end
```
zero201926: I would like to use [https://github.com/makersacademy/rps-exemplar/blob/master/lib/turn.rb](url), the hash makes more sense the the newer exemplars hash, and would shorten my code a lot.

config.ru
jamesjoshuahill: How could you unpack what this does? Maybe start by asking, what is `__FILE__`?

spec/spec_helper.rb
```diff
@@ -1,7 +1,12 @@
+ENV['RACK_ENV'] = 'test'
+require File.join(File.dirname(__FILE__), '..', './lib/app.rb')
```
jamesjoshuahill: Unfortunately, if you require you app code before `SimpleCov.start` the test coverage report will not be accurate. As it is the report indicates 100%, but it's actually 90%:
```ruby
SimpleCov.start

# require app after starting SimpleCov
require File.join(File.dirname(__FILE__), '..', './lib/app.rb')
Capybara.app = RPS
```

config.ru
zero201926: __File__ gives file path to current file.
 When dirname is given a pathname, it will delete any suffix beginning with the last slash ( '/' ) character and return the result.
File.join: joins together elements to form a file path.

PR #1442, opened 2019-10-18
Update README.md

lukegittins01:


Gemfile
```diff
@@ -2,11 +2,13 @@ source 'https://rubygems.org'
 
 gem 'rake'
 gem 'sinatra'
+gem 'rspec'
+
 
 group :test do
```
lukegittins01: What does the 'group :test do' and why would I need it instead of just getting the gems normally at the top?

lib/app.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  get '/' do
+    'Test!'
+  end
+
+  get '/enter-name' do
+    erb :name_form
+  end
+
+  post '/display-name' do
```
lukegittins01: When I wanna go to another page, is it not always just erb :htmlfile? I found that this did not always load the page I wanted and instead stayed on the current one?

spec/features/testing_name_input_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    visit('/enter-name')
+    expect(page).to have_content 'Enter your name below:'
+  end
+end
+
+feature 'Testing name input' do
+  scenario 'check page content has Hello Luke' do
+    visit('/enter-name')
+    fill_in :name, with: 'Luke'
+    click_button 'Submit'
```
lukegittins01: I could have put this in a method and just called this method to save repeating myself for future tests

Gemfile
jamesbingham: We discussed that in the gemfile the gems are grouped together in a variable called :test so those gems could be initiated when it is run. This could possibly be for testing.

lukegittins01: Yeah that could be the case, it would make sense if you could group the gems together and then install just those gems if you wanted to

lib/app.rb
lukegittins01: We discussed the POST-Redirect-GET area and will look online for more detail online

lukegittins01: https://github.com/dearshrewdwit/rps-challenge-apprs-exemplar/blob/bfd22538dfdf8175ae57d7314dd9f425347cdfac/app/app.rb#L8

I could have used sessions to store the information I was receiving from the user

lib/views/name_form.erb
```diff
@@ -0,0 +1,9 @@
+<html>
+  <body>
+    <h1>Enter your name below:</h1>
+    <form action="/display-name" method="post">
+      <input type="text" name="name">
+      <input type="submit" value="Submit">
+    </form>
+  </body>
+</html>
```
lukegittins01: When I was working onto the second part of the task I could have used this form structure to do it:
https://github.com/dearshrewdwit/rps-challenge-apprs-exemplar/blob/bfd22538dfdf8175ae57d7314dd9f425347cdfac/app/views/index.erb#L10

Gemfile
jamesjoshuahill: Good guesses. Take a look at https://bundler.io/ and see if you can find out more.

spec/features/testing_name_input_spec.rb
lukegittins01: Ok thank you, I guess that makes more sense to leave it until you need to extract it 👍

jamesjoshuahill:
Well done for getting the game working @lukegittins01! 🎲

lukegittins01:
> Well done for getting the game working @lukegittins01! 🎲

Thank you @jamesjoshuahill 😃 👍

PR #1443, opened 2019-10-18
all non extention user stories completed

adamr27:


adamr27:
I feel like there is a way to refactor the 9 feature tests I have to do with each different outcome that the page can produce but an immediate way to do that does not come to mind

adamr27:
some of my methods are quite long and could also be refactored, not only this but I would have liked to have tried one of the extension tasks

EarthWormGym:
> I feel like there is a way to refactor the 9 feature tests I have to do with each different outcome that the page can produce but an immediate way to do that does not come to mind

Not sure if possible but could maybe pass arguments/ parameters into a module containing the loop you have used multiple times to shorten/ refactor each feature test.

views/paper.erb
```diff
@@ -0,0 +1,17 @@
...
+<br><% if cp_option == "Rock" %>
+<%="You WIN!"%>
+<% elsif cp_option == "Paper" %>
+<%="It's a DRAW!"%>
+<% elsif cp_option == "Scissors" %>
+<%="You LOSE!"%>
+<% end %>
+
+<form action="/name" method="post">
+  <input type="submit" value="Click to play again">
```
jamesjoshuahill: Great idea to have a button to play again 😄 When I clicked this button then I saw `Hello` with no player name. How could you keep the player's name for the next game?

app.rb
```diff
@@ -0,0 +1,42 @@
...
+    erb :rock
+  end
+
+  post '/paper' do
+    erb :paper
+  end
+
+  post '/scissors' do
+    erb :scissors
+  end
```
jamesjoshuahill: I really like that you provided three buttons for the user to choose what to play. However, there is lots of repetition in the rock, paper and scissors views. How could we use one endpoint and one view to handle this post request?

views/paper.erb
adamr27: I have added a post /play in my app.rb as before it was posting to /name which would then redirect to /play in which it wouldnt know what the name is anymore but now it works

PR #1444, opened 2019-10-18
works for multiplayer but have incomplete single player

albie-b:


lib/checker.rb
```diff
@@ -0,0 +1,17 @@
...
+  elsif @player_2_card == "rock" && @player_1_card == "scissors"
+    "#{@player_2_name} wins"
+  elsif @player_2_card == "paper" && @player_1_card == "rock"
+    "#{@player_2_name} wins"
+  elsif @player_2_card == "scissors" && @player_1_card == "paper"
+    "#{@player_2_name} wins"
+  else
+    "Tie"
+  end
+end
```
TobiasBenjaminYoungCGI: Inefficient code, is there potentially a way to reduce the size of this code? DRY it out?

spec/feature/rps_spec.rb
```diff
@@ -0,0 +1,23 @@
+feature 'can play rps' do
```
TobiasBenjaminYoungCGI: Nice testing, 100% coverage!

lib/checker.rb
albie-b: fixed this now, also I didn't TDD this because I copied it from my solution to mastery quizzes

albie-b: * I didn't TDD the function because the original part of it was from my answer to a mastery quizz

lib/main.rb
```diff
@@ -0,0 +1,13 @@
+require_relative 'checker'
+class Check
+
+  def initialize(player1_card, player2_card, name1, name2, options)
+    @player_1_card, @player_2_card = player1_card, player2_card
+    @player_1_name, @player_2_name, @options = name1, name2, options
+  end
+
+  def confirm
+    checker
```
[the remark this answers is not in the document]
albie-b: done

app.rb
```diff
@@ -0,0 +1,38 @@
...
+    $player_2_name = params[:player_2_name]
+    redirect '/play'
+  end
+
+  get '/play' do
+    @player_1_name, @player_2_name = $player_1_name, $player_2_name
+    erb :play
+  end
+
+  post '/game' do
```
jamesjoshuahill: This method is quite long now. Sometimes this is referred to as a fat controller. See if you move logic out of the controller https://github.com/makersacademy/rps-challenge/blob/master/docs/review.md#fat-controllers.

PR #1446, opened 2019-10-18
rps

eagle310:


spec/features/game_result_spec.rb
```diff
@@ -1,14 +1,11 @@
...
-      expect(page).to have_content "It's a draw"
-    elsif page.has_content? 'Computer played : scissors'
-      expect(page).to have_content "Kirk wins"
-    else
-      expect(page).to have_content "Computer wins"
-    end
+    expect(page).to have_content "You played : "
+    expect(page).to have_content "Computer played : "
+    expect(page).to have_content "win"
   end
```
eagle310: test could test output rather than just checking for the word 'wins'

2Blagden: I guess you could force an output by adding an argument somewhere in your rock paper scissors method (this would involve a lot of refactoring)

app.rb
```diff
@@ -0,0 +1,36 @@
+require 'sinatra'
+require 'shotgun'
```
jamesjoshuahill: When I tried to run the tests with `bundle exec rspec` I got an error:
```
LoadError: cannot load such file -- shotgun
```
I can't see shotgun being used in this file, is it required? If it is, then you should add it to your Gemfile, otherwise please could you remove it so the tests can be run?

lib/game.rb
```diff
@@ -0,0 +1,33 @@
...
+  def game_result
+    @game_options.each do |move|
+      if player_move == move[0]
+        case @computer_move
+        when move[1]
+          return "Computer wins"
+        when move[2]
+          return "#{player_name} wins"
+        when move[3]
+          return "It's a draw, everyone wins"
```
jamesjoshuahill: This message is amazing 💯

PR #1447, opened 2019-10-18
Josh Blackmoor - RPS

joshblackmoor:


lib/game.rb
```diff
@@ -0,0 +1,24 @@
...
+      return "It was a draw"
+    elsif player_input == "rock" && computer_input == "scissors"
+      return "#{@player_name} Wins"
+    elsif player_input == "paper" && computer_input == "rock"
+      return "#{@player_name} Wins"
+    elsif player_input == "scissors" && computer_input == "paper"
+      return "#{@player_name} Wins"
+    else 
+      return "Computer Wins"
+    end
```
joshblackmoor: This could probably be done in a more efficient way than just using IF statements

joshblackmoor:
- May need to add in some more feature tests

- Need to remove the last rubocop errors that I am getting

- Work on the UI and adding in an option to play either against the computer or multiplayer next

lib/game.rb
Will-Bath: Mine was even longer! I would be interested to see how you make this more efficient later on.

lib/game.rb
```diff
@@ -0,0 +1,24 @@
+class Game 
+  attr_reader :player_name
+  def initialize(player_name)
+    @player_name = player_name
+  end 
+
+  def computer_selection 
+    return ['rock', 'paper', 'scissors'].sample
+  end
```
Will-Bath: I like this use of sampling an array to get a random computer choice.

lib/game.rb
jamesjoshuahill: Could you find an alternative approach in the exemplar or another apprentice's solution?

PR #1448, opened 2019-10-18
main features working, outcome spec needs rewriting

2Blagden:
I know my coverage is low, this is because refactoring changed how my methods worked completely so I'll do this ASAP


spec/outcome_spec.rb
```diff
@@ -0,0 +1 @@
+#I refactored my code last minute and the tests were no longer valid, I'll add these ASAP
```
2Blagden: This is where I need to improve!

spec/enter_name_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+end
+
+feature 'entering your name works' do
+  scenario 'enter name jim' do
+    visit('/')
+    fill_in :playername, with: 'Jim'
+    click_button 'Submit'
+    expect(page).to have_content 'Jim'
+  end
+end
```
2Blagden: This doesn't seen to follow DRY. Could I avoid this?

eagle310: make a method to sign in and play and put it in web helpers

spec/outcome_spec.rb
2Blagden: https://github.com/dearshrewdwit/rps-challenge-apprs-exemplar/blob/bfd22538dfdf8175ae57d7314dd9f425347cdfac/app/app.rb#L17

spec/enter_name_spec.rb
jamesjoshuahill: I think you could put both expectations into one scenario:
```ruby
expect(page).to have_content 'Play Rock Paper Scissors'
expect(page).to have_content 'Jim'
```

spec/make_a_move_spec.rb
```diff
@@ -0,0 +1,7 @@
+feature 'player can enter their choice and play against computer' do
+  scenario 'enter r' do
+    visit('/play')
+    fill_in :rockpaperscissors, with: 'r'
+    expect(page).to have_content ''
```
jamesjoshuahill: This doesn't test for what is on the page. What should we see?

views/index.erb
```diff
@@ -0,0 +1,13 @@
+<h1>My name Jeff</h1>
```
jamesjoshuahill: 🤔 but my name is Josh...

app.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  get '/play' do
+    @playername = params[:playername]
+    erb :play
+  end
+
+  post '/move' do
+    @rps = params[:rockpaperscissors]
+    @result_text = bot(@rps, bot_move())
```
jamesjoshuahill: How could you use objects to represent the game?

PR #1449, opened 2019-10-18
completed challenge

EarthWormGym:
Testing and code in general for rps function is quite manual atm but I am going to add a random feature to it next week.

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -1,0 +1,2 @@
+require_relative "./rps_app"
+run Rock_paper_scissors
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -1,0 +1,18 @@
+
+class Game
+
+  attr_reader :winner, :loser
+
+  def workout_winner(player, comp)
+    if (player == "rock" && comp == "scissors") ||
+      (player == "scissors" && comp == "paper") ||
+      (player == "paper" && comp == "rock")
+      @winner = 'computer'
+      @loser = 'adam'
+    else
+      @winner = 'adam'
+      @loser = 'computer'
+    end
+  end
+
+end
--- a/lib/player.rb
+++ b/lib/player.rb
@@ -1,0 +1,13 @@
+class Player
+
+  attr_reader :player_name, :move
+
+  def initialize(name)
+    @player_name = name
+  end
+
+  def players_move(move)
+    @move = move
+  end
+
+end
--- a/rps_app.rb
+++ b/rps_app.rb
@@ -1,0 +1,36 @@
+require 'sinatra'
+require 'game'
+require 'player'
+
+class Rock_paper_scissors < Sinatra::Base
+
+  enable :sessions
+
+  get '/' do
+    erb :main_page
+  end
+
+  get '/infra' do
+    'Testing infrastructure working!'
+  end
+
+  post '/register-names' do
+    $player = Player.new(params[:player])
+    redirect '/registered-players'
+  end
+
+  get '/registered-players' do
+    @player = $player.player_name
+    erb :registered_page
+  end
+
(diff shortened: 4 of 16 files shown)


lib/game.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def workout_winner(player, comp)
+    if (player == "rock" && comp == "scissors") ||
+      (player == "scissors" && comp == "paper") ||
+      (player == "paper" && comp == "rock")
+      @winner = 'computer'
+      @loser = 'adam'
+    else
+      @winner = 'adam'
+      @loser = 'computer'
```
EarthWormGym: just realised I did this backwards meaning the wrong person wins each time, I shall change this round then see if my test still works if not that needs change also.

lib/player.rb
```diff
@@ -0,0 +1,13 @@
+class Player
+
+  attr_reader :player_name, :move
+
+  def initialize(name)
+    @player_name = name
+  end
+
+  def players_move(move)
+    @move = move
```
EarthWormGym: move was something I was planning on using for my rock paper scissors function but I never used it so this anything containing move can be removed.

spec/game_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'game'
+require 'player'
+
+describe Game do
+
+  it 'tests for winner' do
+    adam = Player.new('Adam')
+
+    player_move = adam.players_move("rock")
```
EarthWormGym: could use doubles here to improve test perhaps but how would I use one here?

lib/game.rb
adamr27: When you go to correct this, you could refactor the winner/loser to use the player name params so that the code won't break if the player doesnt enter Adam

config.ru
```diff
@@ -0,0 +1,2 @@
+require_relative "./rps_app"
+run Rock_paper_scissors
```
jamesjoshuahill: Please use CamelCase for class names. You can run `bundle exec rubocop` to get more feedback on the code style.

lib/game.rb
jamesjoshuahill: I can't see unit tests for all these combinations, maybe writing some more would help flush out this issue?

rps_app.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class Rock_paper_scissors < Sinatra::Base
+
+  enable :sessions
+
+  get '/' do
+    erb :main_page
+  end
+
+  get '/infra' do
```
jamesjoshuahill: Do you still need this method?

rps_app.rb
```diff
@@ -0,0 +1,36 @@
+require 'sinatra'
+require 'game'
+require 'player'
```
jamesjoshuahill: When I tried to run your code with `bundle exec rackup` I got an error:
```
LoadError: cannot load such file -- game
```

rps_app.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  post '/register-names' do
+    $player = Player.new(params[:player])
+    redirect '/registered-players'
+  end
+
+  get '/registered-players' do
+    @player = $player.player_name
+    erb :registered_page
+  end
```
jamesjoshuahill: After I registered my player name, I didn't see a link to start the game. I had to go to `/game` manually.

rps_app.rb
EarthWormGym: Ive looked at this task a few times since we pushed it and I believe I have fulfilled all the other comments but I am not sure at all as to why my rack-up doesn't work I seem to be requiring everything correctly and have got the correct gems?

PR #1452, opened 2019-11-01
Rhys' RPS Challenge 

rhyse1:


app.rb
```diff
@@ -0,0 +1,40 @@
...
+    post '/result' do
+        @result = nil
+        @player = session[:player]
+        @player.MakeMove(params[:move])
+        @computer = Computer.new
+        case @player.CheckWon(@computer.Move)
+            when true
+                @result = "You have won!"
+            when false 
+                "You have lost"
```
lindseydew: I don't think the controller is doing what you want here. It only sets the state of @result in one of the cases. In the other cases it is left blank. 

There is also too much logic happening here for a contoller

src/Player.rb
```diff
@@ -0,0 +1,41 @@
...
+        @move = nil
+        @name = name
+    end
+    def MakeMove(move)
+        @move = move
+    end
+
+    def CheckWon(otherMove)
+        case @move
+            when "Rock"
```
lindseydew: As @DMartinBGL mentioned there are more concise ways of expressing this.

src/Player.rb
```diff
@@ -0,0 +1,41 @@
+
+class Player
+    attr_accessor :name
+    attr_reader :move
+    def initialize(name)
+        @move = nil
+        @name = name
+    end
+    def MakeMove(move)
```
lindseydew: This casing is not the convention in ruby. In ruby methods are lower case, separated by underscores.

views/home.erb
```diff
@@ -0,0 +1,30 @@
...
+<div class="form-container">
+    <form action="/game_setup" method="POST">
+        <label>Player Name</label>
+        <input type="text" name="name">
+        <input type="submit" value="Start Game">
+    </form>
+</div>
+
+<style>
+body {
```
lindseydew: CSS could be separated in its own file

spec/player_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+        expect(subject.CheckWon("Rock")).to eq true
+    end
+
+    it "test more win conditions" do
+        subject.MakeMove("Scissors")
+        expect(subject.CheckWon("Rock")).to eq false
+    end
+
+    it "test more win conditions" do
+        subject.MakeMove("Paper")
```
lindseydew: Good coverage of unit testing

spec/player_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+        expect(subject.CheckWon("Paper")).to eq "Draw"
+    end
+end
```
lindseydew: Don't seem to be any feature tests?

PR #1456, opened 2019-11-01
Rock Paper Scissors Challenge

marcb1988:
Implemented first user story and began implementing code to satisfy the second, but did not complete it due to several false starts.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,16 @@
+require "sinatra"
+require "sinatra/base"
+
+class RPS < Sinatra::Base
+  enable :sessions
+  get "/" do
+    erb :index
+  end
+
+  post "/name" do
+    @player_name = params[:player_1]
+    erb :play
+  end
+
+  run! if app_file == $0
+end
--- a/app.rb
+++ b/app.rb
@@ -16,5 +16,9 @@ class RPS < Sinatra::Base
     @player_name = session[:player_1_name]
     erb :play
   end
+
+  get "/player1_move" do
+    "Player1 made a choice"
+  end
   run! if app_file == $0
 end
--- a/app.rb
+++ b/app.rb
@@ -1,24 +1,41 @@
 require "sinatra"
 require "sinatra/base"
+require "./lib/player"
 
 class RPS < Sinatra::Base
   enable :sessions
+
   get "/" do
     erb :index
   end
 
   post "/name" do
-    session[:player_1_name] = params[:player_1]
+    # session[:player_1_name] = params[:player_1]
+    @player = Player.create(params[:player_1])
     redirect "/play"
   end
 
+  # before do
+  #   @player = Player.instance
+  # end
+
   get "/play" do
-    @player_name = session[:player_1_name]
+    @player = Player.instance
     erb :play
   end
 
-  get "/player1_move" do
-    "Player1 made a choice"
+  post "/player1_move" do
+    @player = Player.instance
+    session[:player1_move] = params[:radio]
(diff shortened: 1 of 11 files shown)


lib/player.rb
```diff
@@ -1,11 +1,30 @@
...
+    @player = Player.new(player_1)
+  end
+
+  def self.instance
+    @player
+  end
+
   def weapon(choice)
     @choice = choice
+    @comp_choice = ["Rock", "Paper", "Scissors"].sample
```
bradleyplater: Comp could be misleading, computer would work better for readability

lib/player.rb
```diff
@@ -1,11 +1,30 @@
...
+    @winner = "You" if @choice == "Rock" && @comp_choice == "Scissors"
+    @winner = "You" if @choice == "Paper" && @comp_choice == "Rock"
+    @winner = "You" if @choice == "Scissors" && @comp_choice == "Paper"
+    @winner = "The Computer" if @choice == "Rock" && @comp_choice == "Paper"
+    @winner = "The Computer" if @choice == "Paper" && @comp_choice == "Scissors"
+    @winner = "The Computer" if @choice == "Scissors" && @comp_choice == "Rock"
+    @winner = "Nobody" if @choice == "Rock" && @comp_choice == "Rock"
+    @winner = "Nobody" if @choice == "Paper" && @comp_choice == "Paper"
+    @winner = "Nobody" if @choice == "Scissors" && @comp_choice == "Scissors"
```
bradleyplater: Could be put into Case statement, and private functions could be made to DRY the code

bradleyplater: Also no need for @winner, as you could just return the string its self rather than storing it within a variable

bradleyplater:
It works which is good, few things could be changed this may be personal preference, within your player class, you define 'self.instance' and 'self.create' when you call these you are essentially just calling, '@player = Player.new(...)' and 'return @player ' which you could also write within the app.rb file rather than calling, '@player = Player.create(...)' within it. Also within you app.rb file you are creating a session with the parameter of 'radio', this makes little sense to someone who is reading it and i would change this to be called 'move'. You are also calling a method 'weapon' which is passed through a move so it would make sense to change the method to move, because it otherwise weapon sounds like an attribute and not an action. Any questions come speak to me :)

lib/player.rb
```diff
@@ -1,11 +1,30 @@
 class Player
-  attr_reader :name, :choice
+  attr_reader :name, :choice, :winner
 
   def initialize(name)
     @name = name
   end
 
+  def self.create(player_1)
```
bradleyplater: Possibly remove this

lib/player.rb
```diff
@@ -1,11 +1,30 @@
...
+    @player = Player.new(player_1)
+  end
+
+  def self.instance
```
bradleyplater: Possibly remove this

app.rb
```diff
@@ -1,24 +1,41 @@
...
-    @player_name = session[:player_1_name]
+    @player = Player.instance
     erb :play
   end
 
-  get "/player1_move" do
-    "Player1 made a choice"
+  post "/player1_move" do
+    @player = Player.instance
+    session[:player1_move] = params[:radio]
```
bradleyplater: Maybe change :radio to :move

lindseydew:
Great job @marcb1988 and nice review @bradleyplater 💯 🥇 

It would be great to have a feature test on the result of the game, and there are a couple of refactoring opportunities that Bradley mentioned but on the whole really great work 👍

PR #1458, opened 2019-11-01
RPS challenge 

swade19:


lindseydew:
Excellent start 💯 

Endpoints are well designed and well tested 👍 

If you would like more practise of objects in web apps I can recommend this exercise https://github.com/makersacademy/database-models-challenge for writing ruby objects that interact with the database

PR #1461, opened 2019-11-01
Meh

jtsalva:


app.rb
```diff
@@ -0,0 +1,47 @@
...
+    session[:name] = name
+    player = Player.new(name)
+    bot = Player.new('Bot')
+    session[:game] = Game.new(player, bot)
+
+    redirect '/play'
+  end
+
+  get '/play' do
+    @game = session[:game]
```
thorpelawrence: Does this need to be an instance variable? Is `@game` used in the erb file?

app.rb
```diff
@@ -0,0 +1,47 @@
+require 'sinatra/base'
+require_relative 'lib/game'
+require_relative 'lib/player'
+
+class RPS < Sinatra::Base
```
thorpelawrence: I think this file would be clearer if named `rps.rb` instead of `app.rb`

lib/game.rb
```diff
@@ -0,0 +1,42 @@
+class Game
+  attr_reader :player_one, :player_two, :player_one_choice, :player_two_choice, :history
+
+  CHOICES = {
```
thorpelawrence: :+1: for enum instead of hardcoded magic strings

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+  def winner
+    return nil if @player_one_choice == nil or @player_two_choice == nil
+    result = CHOICES[@player_one_choice] - CHOICES[@player_two_choice]
+
+    if (@player_one_choice == :rock and @player_two_choice == :paper) or (@player_two_choice == :rock and @player_one_choice == :paper)
+      result *= -1
+    end
+
+    if result == 0
+      "Draw"
```
thorpelawrence: Not necessary for now, but in future this might want to be a constant value so that it can be reused elsewhere, also allows for example translation to other languages

lib/player.rb
```diff
@@ -0,0 +1,32 @@
...
+  def loss
+    @losses += 1
+  end
+
+  def games_played
+    @wins + @losses
+  end
+
+  def make_random_choice
+    case Kernel.rand(3)
```
thorpelawrence: I'm not sure specifically for Ruby, but is the `Kernel` needed here, is it different from `rand(3)` on its own?

views/play.erb
```diff
@@ -0,0 +1,71 @@
...
+      <button class="choice scissors-button" disabled>Scissors</button>
+    <% else %>
+      <p id="waiting-for-other-player-message">Waiting for <%= @other_player.name %> to choose...</p>
+    <% end %>
+  </div>
+</div>
+<form id="choose-paper" action="/play/choose_paper" method="post"></form>
+<form id="choose-scissors" action="/play/choose_scissors" method="post"></form>
+<form id="choose-rock" action="/play/choose_rock" method="post"></form>
+<script src="https://code.jquery.com/jquery-3.4.1.min.js" integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo=" crossorigin="anonymous"></script>
```
thorpelawrence: Given the time it's very understandable, but maybe jQuery wouldn't be needed here?

spec/player_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    }.to change {subject.losses}.from(0).to(losses)
+  end
+
+  it 'has number of games played' do
+    expect(subject.games_played).to eq(subject.wins + subject.losses)
+  end
+
+  describe 'random choices' do
+    it 'can make a random choice to paper' do
+      allow(Kernel).to receive(:rand).and_return 0
```
thorpelawrence: Is the reason kernel is referred to earlier with `Kernel.rand(...)` because of this? That makes sense, as I wasn't aware it was possible to do that!

spec/player_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'player'
+
+describe Player do
+  subject { Player.new(name) }
+  let(:name) { 'Bob Ross' }
```
thorpelawrence: Happy little :cloud:s

public/main.css
```diff
@@ -0,0 +1,7 @@
+* {
+    font-family: 'Roboto';
```
thorpelawrence: Would probably be a good idea to have a fallback font in case the browser doesn't have access to 'Roboto' or it isn't loaded properly, e.g.
```
font-family: "Roboto", sans-serif
```

views/index.erb
```diff
@@ -0,0 +1,17 @@
+<html lang="en">
```
thorpelawrence: Good HTML formatting!

PR #1467, opened 2019-11-03
Amy's pull request for RPS challenge

amybalmforth:


spec/features/result_spec.rb
```diff
@@ -0,0 +1,51 @@
+feature 'Result of game, player wins' do
+
+  scenario 'See computer choice' do
+    allow_any_instance_of(Computer).to receive(:choice).and_return('scissors')
```
Liatmoss: Use a 'before do' so you don't repeat the same step in every test

views/choose.erb
```diff
@@ -0,0 +1,9 @@
+<body bgcolor="seashell">
```
Liatmoss: I like the styling of your website. Makes it unique to you

PR #1472, opened 2019-11-03
Liat rps challenge

Liatmoss:


views/choose.erb
```diff
@@ -0,0 +1,11 @@
+<%= @player_name %> choose Rock, Paper or Scissors
+<br>
+<form><center>
+  <select name="dropdown">
```
amybalmforth: I like the drop down menu, neater than buttons!

spec/computer_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'computer'
+
+  describe 'computer' do
+    subject(:computer) { Computer.new }
+
+      it 'should return a computer choice' do
+        expect(computer.choice).to eq("scissors").or eq("paper").or eq("rock")
```
amybalmforth: This seems an interesting way to test randomness which I never came across before ("or eq") - it still tests the expected output so seems like a good solution

PR #1478, opened 2019-11-03
Donald Newholm pull request

CodeKrakken:


lib/game.rb
```diff
@@ -0,0 +1,18 @@
...
+    @computer_weapon = ["Rock", "Paper", "Scissors"].sample
+  end
+
+  def compare_weapons(player_weapon)
+    return "It's a tie." if player_weapon == @computer_weapon
+    return "You win." if player_weapon == "Scissors" && @computer_weapon == "Paper"
+    return "I win." if player_weapon == "Paper" && @computer_weapon == "Scissors"
+    return "You win." if player_weapon < @computer_weapon
+    
+    "I win."
```
eliseaston: Bar the questions above, in general this seems a v neat solution, to have a default "I win" result and just specify if otherwise. Well done!

lib/game.rb
```diff
@@ -0,0 +1,18 @@
...
+  attr_reader :computer_weapon
+  
+  def initialize
+    @computer_weapon = ["Rock", "Paper", "Scissors"].sample
+  end
+
+  def compare_weapons(player_weapon)
+    return "It's a tie." if player_weapon == @computer_weapon
+    return "You win." if player_weapon == "Scissors" && @computer_weapon == "Paper"
+    return "I win." if player_weapon == "Paper" && @computer_weapon == "Scissors"
```
eliseaston: Why do you need to specify an extra "I win" here?

lib/game.rb
```diff
@@ -0,0 +1,18 @@
...
+    return "You win." if player_weapon < @computer_weapon
```
eliseaston: What does it mean for the weapons to be < or > to each other, if they're not numerical value?

app.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  get '/choose' do
+    @@game = Game.new
+    @player_name = session[:player_name]
+    erb :choose
+  end
+
+  get '/rock' do
+    @player_weapon = "Rock"
+    @@game.computer_weapon
```
eliseaston: Do you need to declare @@game.computer_weapon here? Would @@game.compare_weapons work without declaring it?

spec/features/choose_weapon_spec.rb
```diff
@@ -0,0 +1,7 @@
+feature 'Weapon chooser' do
+  scenario 'displays chosen weapon' do
+    sign_in_and_play
+    click_button 'Rock'
+    expect(page).to have_content 'You chose Rock'
+  end
+end
```
eliseaston: Comment on tests in general: seem to have good coverage of the program, good use of capybara for feature testing etc. Well done!
You could consider having a test to check the randomness of the Computer's selection. And also some unit tests for your game.rb file.

CodeKrakken: Thanks! I did consider testing for randomness, but I ran out of time unfortunately. Couldn't figure out how to write the test in Capybara.
As for unit tests, I did use TDD and I was surprised how few tests I ended up with! I probably went off piste without realising it at some stage. Thanks for your review!

lib/game.rb
CodeKrakken: Thanks! I'll answer your previous questions here too - the < and > seemed a good way to compare most situations, as Ruby considers B > A, and therefore Scissors > Rock > Paper (each thing beats the weapon that is alphabetically greater than itself). The exception was Scissors beating Paper, which I had to compare before the main comparison on line 13. This is why there is an extra "I win". It was a neat method til I realised very late that I had to compare Scissors and Paper separately.

spec/features/choose_weapon_spec.rb
CodeKrakken: Thanks! I did try to test for randomness, but I ran out of time unfortunately. Couldn't figure out how to write the test in Capybara.
As for unit tests, I did use TDD and I was surprised how few tests I ended up with! I probably went off piste without realising it at some stage. Thanks for your review!

PR #1484, opened 2019-11-04
RPS Challenge

jade-genevieve:


KrisN64209:
Code is simple and easy to understand, tests all passing, and game works up to this point when ran in browser

lindseydew:
Nice job @jade-genevieve and agree with @KrisN64209 review :)

PR #1485, opened 2019-11-04
the page show the names of the players

miglekuba:


lindseydew:
Looks great well done @miglekuba and lovely review @swade19. Good idea about the progress update. 

I can see that you are still working on the game logic - what I would suggest working on this week is the interaction between the ruby model and the database.  This might be a helpful exercise to practise that https://github.com/makersacademy/database-models-challenge

PR #1486, opened 2019-11-04
Lena's RPS challenge

olkras03:


lindseydew:
Great start on this 💯 
If you would like more practise of objects in web apps I can recommend this exercise https://github.com/makersacademy/database-models-challenge for writing ruby objects that interact with the database

PR #1494, opened 2019-11-22
Rob Grant (Oct 19) - RPS Challenge Pull Request

djlonboy:
- Setup the project folder includng folders for lib, spec (feature & unit)
 - Added model file (rps.rb), controller (app.rb), first view file (index.rb)
 - Added config.ru for rackup
 - Updated spec_helper file with correct gems and link to app file for capybara
 - Updated README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,86 +1,34 @@
 # RPS Challenge
+This web app allows users to play an online game of Rock, Paper, Scissors.
 
-Instructions
--------
-
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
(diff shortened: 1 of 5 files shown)


lib/computer.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def initialize(name)
+    @name = name
+  end
+
+  def choose
+    # Randomly chooses Rock, Paper or Scissors
+    choice = rand(0..2)
+    choices = ["Rock", "Paper", "Scissors"]
+    choices[choice]
```
sibamunsanje: This is outstanding work, Rob. My only suggestion is to refactor the code so that you can add .sample to your choices array. This can help your code read better with two lines instead of three.

PR #1497, opened 2019-11-22
stories completed

pavidshetty:


config.ru
```diff
@@ -1,2 +1,2 @@
-require_relative "./app" #you can also require './app./rb'
+require './app./rb'
```
david-mears: This line prevented rackup from working - it said `cannot load such file -- ./app./rb`. This is just a misspelling of the file name.

app.rb
```diff
@@ -1,14 +1,35 @@
 require 'sinatra'
 require 'sinatra/base'
+require '.lib/game'
+require '.lib/player'
```
david-mears: These `require`s were breaking the app with `cannot load such file` error messages. It's just a misspelling. Because `.` means 'the folder which contains this file', it needs to have a `/` just after it.

lib/game.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :p2_move
+
+  def initialize(p2_move)
+   @p2_move = ["Rock", "Paper", "Scissors"].sample
+  end
+
+
+  def result
+    if @p1_move == "rock" && @p2_move == "scissors"
+      puts "Player 1 wins"
```
david-mears: Where will `puts` output this string? How will the result be accessible to your program's logic?

lib/game.rb
```diff
@@ -0,0 +1,26 @@
...
+    elsif @p1_move == "scissors" && @p2_move == "paper"
+      puts "Player 1 wins"
+    elsif @p1_move == @p2_move
+      puts "It's a draw"
+    else
+     puts "Player 2 wins"
+    end
+  end
+
+  def self_instance
```
david-mears: Use `self` with `.` to make `self.instance`.

views/results.erb
```diff
@@ -0,0 +1,3 @@
+<p> <%=@p1_name%> selected <%=p1_move%>!</p>
```
david-mears: Rspec told me: `NameError:
       undefined local variable or method `p1_move' for #<RockPaperScissors:0x00007fac57591ba8>
       Did you mean?  @p1_move`

app.rb
```diff
@@ -1,14 +1,35 @@
...
+  get '/choice' do
+    erb :choice
+  end
+
+  # post '/move' do
+  #   redirect '/results'
+  # end
   #
+  # get '/results' do
+  #   @p1_move = params[:p1_move]
```
david-mears: You probably have an empty `params`, because this is a `get` route. In choice.erb, I see
```html
<form action="/move" method="post">
```
which tells me that the data is sent to the route `post '/move' do`. That's where `params` will be useful.

app.rb
```diff
@@ -1,14 +1,35 @@
...
+  #   @game.result(params[:p1_move], player2_move)
```
david-mears: If you want to display @p1_name in results, it will need to be available in this `get '/results'/` method too. But @p1_name stopped existing on line 19 in this file. How could you store this name so that it would continue to be available for methods outside `post '/names'`?

david-mears: There are various possible options!

david-mears:
It seems as though you have had a working program, but some of it broke before your final commit, while you were trying to extend the program? Is that right? That suggests to me that you would benefit from a more consistent process. Particularly, making commits more often, and test-driving.

Committing each time you add a new working piece of functionality would let you roll back to a recent working version, AND let you see what you have changed since then (using `git diff`).

Your process (and the program's stability) would benefit from including unit tests (tests of game.rb and player.rb). The tests you have look good, so use them more, and try to test-drive more: you can use tests to make sure that your program is still working after you make a change. It's a good idea to make a commit at least each time you pass a new test, which will line up with each time you add a new piece of functionality. So try to make a habit of running `rspec` more often, especially right before committing: linking it to `git commit` in your muscle memory.

pavidshetty:
> It seems as though you have had a working program, but some of it broke before your final commit, while you were trying to extend the program? Is that right? That suggests to me that you would benefit from a more consistent process. Particularly, making commits more often, and test-driving.
> 
> Committing each time you add a new working piece of functionality would let you roll back to a recent working version, AND let you see what you have changed since then (using `git diff`).
> 
> Your process (and the program's stability) would benefit from including unit tests (tests of game.rb and player.rb). The tests you have look good, so use them more, and try to test-drive more: you can use tests to make sure that your program is still working after you make a change. It's a good idea to make a commit at least each time you pass a new test, which will line up with each time you add a new piece of functionality. So try to make a habit of running `rspec` more often, especially right before committing: linking it to `git commit` in your muscle memory.

Thanks David ....Yes you are absolutley right ...I was more eager to write the codes and get the web app working, rather than writing each unit and test the functionality....Ill start again and test drive more .....

PR #1500, opened 2019-11-22
Rps Challenge

SimoneW2:


spec/player_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'player'
+
+describe Player do
+  subject(:simone) {Player.new('Simone')} #Creates an instance of player, and passes the argument :dave with the string "Dave"
```
savagevince: I'm not sure I understand the "subject[:simone]" in the context of the test.  Can you explain what it is doing?

views/play.erb
```diff
@@ -1 +1,9 @@
-<h2> Hi <%= @player_name %> are you ready to play?</h2>
+<h2> Hi <%= @player_name %> are you ready to play Rock, Paper, Scissors?</h2>
```
savagevince: I really like the layout of the radio buttons down the page using <p></p> and <br>

spec/features/player_options_spec.rb
```diff
@@ -0,0 +1,12 @@
+
+  feature 'Player selects an option ' do
+    scenario 'submitting rbs option in the second form' do
+      visit('/')
+      fill_in :player_name, with: 'Simone'
+      click_button 'Submit'
+      visit('/game_redirect')
```
david-mears: I found this test wasn't passing, so I went through it step by step. It fails at this step: at http://localhost:9292/game_redirect I see the error 'Sinatra doesn't know this ditty'.

spec/player_spec.rb
david-mears: Simone is defining a variable `simone`, which will be recreated for each test.

PR #1503, opened 2019-11-25
Sam King Pull Request (Oct 2019)

sk52:


README.md
```diff
@@ -1,5 +1,49 @@
 # RPS Challenge
 
+Samuel King Progress Report - Friday 22nd November 2019 @ 15:52
+-------
+
+Overview:
+-------
+
+* Two user stories completed
```
marianneoco: Excellent README

marianneoco: I admire your commitment to doing this properly. This would be very useful for someone reviewing.

README.md
```diff
@@ -1,5 +1,49 @@
...
+* Yet to be implemented is the implementation of 2 player mode
+  * This should be straightforward due to the structure of the Game class
+  * It will require a number of additional routes in the controller to make this function, but should not require much, if any, change to the model - [plan](https://github.com/sk52/rps-challenge/blob/master/IMG_6375.jpg)
+* I would like to refactor to remove my use of global variables as instructed [here](https://github.com/makersacademy/course/blob/master/intro_to_the_web/killing_the_global_variable.md)
+* I would also like to attempt to refactor to remove the need for logic in the results view - this would need a rethink on the model as to what the #play method returns (player object if winner, string "draw" if no winner)
+
+Lessons learned from this challenge:
+-------
+
+* Learned about false positives - in my feature tests on the results page I encountered these. This was because the feature test itself is present in the error page for page_status 500. Solutions are:
```
marianneoco: Interesting.

sk52:
I would be particularly keen to get feedback on: 
- My general approach and TDD style
- My feature test coverage - it does not cover all outcomes of RPS combinations, what are your thoughts on this?
- My plan (see image file linked in README) for implementing two player, do you have any feedback or thoughts on this approach?

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+    }
+    combinations[[players.first.move, players.last.move]]
+  end
+
+  def draw?
+    play == "draw"
+  end
+
+  def player_1
+    players.first
```
marianneoco: What is the advantage of using this as opposed to attribute readers and assigning @player1/ @player2 at initiation of the game? Just a Q as I'm not sure :)

lib/player.rb
```diff
@@ -0,0 +1,15 @@
...
+  def initialize(name)
+    @name = name
+  end
+
+  def make_move(move)
+    @move = move
+  end
+
+  def random_move
+    @move = ["Rock", "Paper", "Scissors"].sample
```
marianneoco: Good use of the .sample method

spec/features/result_spec.rb
```diff
@@ -0,0 +1,24 @@
+feature 'Result' do
+  scenario 'result screen shows computers move' do
+    enter_name_and_start_game
+    srand(4)
+    click_button("Rock")
+    expect(page).to have_content "Computer selected Scissors!"
```
marianneoco: I -suspect- you should test all possible outcomes. Eg this wouldn't throw up an error if your click_button("Paper") wasn't working.

marianneoco: Especially as your unit test also uses rock.

spec/units/player_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  describe '#name' do
+    it 'should return the name' do
+      expect(sam.name).to eq("Sam")
+    end
+  end
+
+  describe '#make_move' do
+    it 'should set the player move' do
+      expect { sam.make_move("Rock") }.to change { sam.move }.to("Rock")
```
marianneoco: This is a nice short rspec test! I haven't used change... to but it's useful

app.rb
```diff
@@ -0,0 +1,48 @@
+require 'sinatra/base'
+require './lib/game'
+require './lib/player'
+
+class RockPaperScissors < Sinatra::Base
```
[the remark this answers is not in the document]
marianneoco: I'm not 100% sure how this would work from the plan - would be helpful to add detail on app.rb/ additional views required. I ended up with many extra views :')

lib/game.rb
sk52: Great question - I felt that both approaches would have the same outcome, but decided that by defining the method it would give me more control over this, with less being assigned in my initialize method. Ultimately though no perceivable benefit in the games current state.

spec/features/result_spec.rb
sk52: That's a great point - I'll consider implementing this.

README.md
```diff
@@ -1,5 +1,49 @@
...
+  * Either be more specific in where I am expecting something on the page so that it doesn't just check the whole page
```
david-mears: Well spotted with this unexpected behaviour from using Sinatra with Capybara. That seems like a good solution: specifying an html element to contain the expected text.

app.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    player_1 = Player.new(params[:name])
+    player_2 = Player.new("Computer")
+    @game = Game.create(player_1, player_2)
```
david-mears: Is there a reason to use `Game.create` rather than `Game.new`?

As far as I can tell, the difference is that `#initialize` directly creates an instance, but `#create` attaches one to an instance variable.

david-mears: I've now seen that you can access this via your `#instance` method, which saves space in app.rb.

I think your instance variable `@game` which you are accessing with `#instance` is really a class variable, as it belongs to the class (and the class methods), rather than to a Game object. Class variables should have `@@` at the start and canbe initialized in a method, or in a similar way to constants (within class, outside methods).

david-mears: Hmm, in fact you could make all your Game instance methods operate on the level of the class, if you wanted to make sure there could only be one Game at a time. That would eliminate the redundancy of having a `create` method whose only responsibility is to delegate to `new`.

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def self.create(player_1, player_2)
+    @game = Game.new(player_1, player_2)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def play
```
david-mears: In general, method names are verbs if they mutate a state (change something's value), and nouns if they just calculate and return.

You asked about whether a method should stick to returning just one type (string or object, in this case). That's an excellent point, which you'd be forced to resolve if you had to choose to name this method after what it returned: if you called it `result` then it would seem a bit odd to return Player objects; if you called it `winner` it would seem a bit odd to return a string "draw".

Rather than returning a string, you could return a symbol: these aren't mutable, so there's a little less risk of it being altered by something later. There's a convention of using symbols for this kind of thing, where it's an immutable token recording a state, and which won't be directly displayed.

app.rb
sk52: This is a good point - would this be something along the lines of `def self.new(player_1, player_2)`? Which would allow me to remove `create`. I suppose in my implementation's current state it's probably preferable to only allow one instance.

I'll do some reading on this, and look into class variables too!

david-mears: `def self.new(player_1, player_2)` Something along those lines, but how would you know whether `Game.new` would try to call `#initialize` or `#self.new`?

PR #1504, opened 2019-11-25
My first game

amgc3:
I have the player and the computer. It works for these two.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,5 @@
+require 'sinatra/base'
+
+class RockPaperScissors < Sinatra::Base
+
+end
--- a/app.rb
+++ b/app.rb
@@ -1,5 +1,21 @@
 require 'sinatra/base'
 
 class RockPaperScissors < Sinatra::Base
+enable :sessions
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/player-name' do
+    session[:name] = params[:name]
+    @player = session[:name]
+    erb(:play)
+  end
+
+  # get '/play' do
+  #   @player = session[:name]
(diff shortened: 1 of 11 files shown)


app.rb
```diff
@@ -0,0 +1,25 @@
...
+  get '/' do
+    erb(:index)
+  end
+
+  post '/player-name' do
+    session[:name] = params[:name]
+    @player = session[:name]
+    erb(:play)
+  end
+
```
beccyv: I like the way you've done the sessions, I didn't do this, but I wish I had because my game would have been better for it.

lib/game.rb
```diff
@@ -0,0 +1,25 @@
...
+    if @player1_choice == "rock" && @player2_choice == "scissors"
+      return "Player 1 wins"
+    elsif @player1_choice == "paper" && @player2_choice == "rock"
+      return "Player 1 wins"
+    elsif @player1_choice == "scissors" && @player2_choice == "paper"
+      return "Player 1 wins"
+    elsif @player1_choice == @player2_choice
+      return "It's a draw"
+    else
+      return "Player 2 wins"
```
beccyv: I like the way you did this. I should have done this, but didn't.

app.rb
```diff
@@ -0,0 +1,25 @@
...
+  post '/play' do
+    session[:move] = params[:move]
+    @player1_choice = session[:move]
+    $game = Game.new(session[:name], @player1_choice)
+    @result = $game.play
+    erb(:result)
```
beccyv: I like the fact that this gives a more flexible output in the views page for result.

PR #1506, opened 2019-11-25
Rebecca Varley pull request

beccyv:


lib/game.rb
```diff
@@ -0,0 +1,28 @@
...
+    @cchoice = computers_choice
+  end
+
+  def compare
+    choice_set
+    if @rpschoice == @cchoice
+      @winner = "Nobody"
+    elsif (@rpschoice == "ROCK" && @cchoice == "SCISSORS") || (@rpschoice == "SCISSORS" && @cchoice == "PAPER") || (@rpschoice == "PAPER" && @cchoice == "ROCK")
+      @winner = @player
+    elsif (@rpschoice == "ROCK" && @cchoice == "PAPER") || (@rpschoice == "SCISSORS" && @cchoice == "ROCK") || (@rpschoice == "PAPER" && @cchoice == "SCISSORS")
```
amgc3: I would have just used else @winner = computer since it would take all the cases that have not been included in the previous two.

lib/game_multiple.rb
```diff
@@ -0,0 +1,21 @@
+class Gamemultiple
+
+  attr_reader :choice, :choice_2, :player, :player_2, :winner
+
+  def initialize(choice, choice_2, player, player_2)
+    @choice = choice.upcase
```
amgc3: I should use .upcase as well.

Views/winners_multiple.erb
```diff
@@ -0,0 +1,19 @@
...
+  a {
+    color: #609096;
+  }
+</style>
+
+<h2>Welcome to the game <%= @player_1_name %> and <%= @player_2_name %> </h2></br>
+
+<h2><%= @player_1_name %> chose <%= @choice %></h2></br>
+
+<h2> <%= @player_2_name %> chose <%= @choice2 %></h2></br>
```
rahulshah7: Use consistent variable names e.g. @choice_2 rather than @choice2. And @choice_1 instead of @choice

rahulshah7 (changes requested):

PR #1510, opened 2019-11-25
Rps Challenge

sujesh-roy:
Tried to relate this challenge to Battle and approached it in the same way.

### DIFF
--- a/spec/features/enter_name_spec.rb
+++ b/spec/features/enter_name_spec.rb
@@ -1,0 +1,6 @@
+feature 'Enter names' do
+  scenario 'submit player name' do
+    enter_and_play
+    expect(page).to have_content 'Dave vs. CPU' # Expect to see First_name vs Second_name
+  end
+end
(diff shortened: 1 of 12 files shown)


spec/features/enter_name_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature 'Enter names' do
+  scenario 'submit player name' do
+    enter_and_play
```
david-mears: To use your web_helpers.rb, you will need to `require` it, either here or in `spec_helper.rb` (which Rspec loads first).

david-mears:
In future challenges, could you include more commits, so that we can have insight into and give feedback on your process? For example, this submission doesn't load with `rackup` (there's a bug), but I can't tell if that's a recently introduced error, or you were blocked by it for a long time.

spec/features/enter_name_spec.rb
sujesh-roy: Thanks for your feedback David. I will look into this and try to resolve the issue.

PR #1516, opened 2019-12-01
RPS Challenge

dbacall:


lib/player.rb
```diff
@@ -0,0 +1,9 @@
+class Player
+
+  attr_reader :name
+
+  def initialize(name = name)
```
even13: probably don't need the = name unless you'd like it to have a default value :)

lib/game.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def self.create(name)
+    @game = Game.new(name)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def outcome(player_choice, bot_choice)
```
even13: maybe outcome could have its own class? SRP

PR #1518, opened 2019-12-01
RPS Challange

jj49411:


lib/game.rb
```diff
@@ -0,0 +1,23 @@
...
+attr_reader :computer_move
+MOVE = [:rock, :paper, :scissors]
+
+
+  def random_move 
+    MOVE.sample
+  end
+
+  def result(player_move)
+    puts "----------- #{player_move} ----------"
```
Katerina-codes: This display code belongs in the view not in this class :) @jj49411

PR #1519, opened 2019-12-01
RPSLS Challenge - pull request

IngramCapa:


spec/computer_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'computer'
+require 'rules'
+
+describe Computer do
+
+  let(:name) { double :name }
+  let(:choices) { double :choices }
```
shafali03: Good use of doubles

PR #1522, opened 2019-12-01
RPS-Challenge Attempt

Jamie95187:


README2.MD
```diff
@@ -0,0 +1,51 @@
+## Introduction
+
+A rock paper scissors game that allows the user to input their name into the game. The player is greeted and has three options to choose from. Then they are prompted by a message displaying what their opponent (the computer) has chosen at random. Finally the results of the game are shown on the page.
+
```
AndreaDiotallevi: Nice description.

app.rb
```diff
@@ -0,0 +1,43 @@
...
+class Rps < Sinatra::Base
+
+  enable :sessions
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/name' do
+    session[:player1_name] = params["player"]
```
AndreaDiotallevi: This works fine. You also could have used class methods, instead of sessions.

app.rb
```diff
@@ -0,0 +1,43 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    session[:player1] = Player.new(session[:player1_name])
+    @player1 = session[:player1]
```
AndreaDiotallevi: I think line numbers 19 & 20 could be just one line: ```@player1 = Player.new(session[:player_name])```

app.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  post '/move' do
+    session[:move] = params["move"]
+    redirect '/move_page'
+  end
+
+  get '/move_page' do
+    @move = session[:move]
+    # This sets the move for the player object
+    @player1 = session[:player1]
```
AndreaDiotallevi: This is a repetition of line 20. So probably using class methods would have helped minimising repetitions.

app.rb
```diff
@@ -0,0 +1,43 @@
...
+    @player1.send @move.downcase.to_sym
```
AndreaDiotallevi: You should probably implement the methods ```downcase.to_sym``` within the "model" part of MVC (i.e. inside the Player class).

app.rb
```diff
@@ -0,0 +1,43 @@
...
+    session[:game] = Game.new(@player1)
+    @game = session[:game]
```
AndreaDiotallevi: You don't need to use a session here as it is within the same route.

lib/game.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'player'
```
AndreaDiotallevi: I don't think you need to require 'player'.

lib/game.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class Game
+  attr_reader :player_1, :comp_move
+
+  MOVES = ["Rock", "Paper", "Scissors"]
```
AndreaDiotallevi: Since these strings won't change, you could have used symbols.

lib/game.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def initialize(player1)
+    @player_1 = player1
+  end
+
+  def random_move
```
AndreaDiotallevi: Probably the responsibility of generating a random computer move should be implemented by a Computer class. The Game class should be responsible only for calculating the result and keeping the score.

lib/game.rb
```diff
@@ -0,0 +1,29 @@
...
+    @comp_move = MOVES[rand(3)]
+  end
+
+  def result
+    return "Draw" if @player_1.move == @comp_move
+
+    lose? ? "You lose" : "You win"
```
AndreaDiotallevi: The text message that will be show in the screen should be a "View" concern. This method should return either :win, :lose or :tie. The actual message should be implemented by ```*.erb``` files.

spec/unit/game_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  before do
+    allow(player1).to receive(:move).and_return "Rock"
+  end
+
+  context "Player has selected Rock" do
+    it 'should return draw' do
+      subject_rock(subject)
+      expect(subject.result).to eq "Draw"
```
AndreaDiotallevi: Here you should check the the result shown is as expected (i.e. :win, :lose or :tie). The test for the actual text message should be inside the feature tests.

views/move_page.erb
```diff
@@ -0,0 +1,5 @@
+<h2>You have selected <%=@move%></h2>
+
+<h2>Your opponent has selected <%=@game.comp_move%></h2>
+
+<h2><%=@game.result%></h2>
```
AndreaDiotallevi: Here should be implemented: "You won" for example. You might need a separate ```erb``` file for a win, lose and tie.

spec/features/result_page_spec.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'sign_in_and_play_helper'
```
AndreaDiotallevi: If ```require_relative 'features/web_helpers'``` in the spec_helper.rb, you don't need to require it here.

README2.MD
```diff
@@ -0,0 +1,51 @@
...
+## Folder layouts
+
+The unit tests can be found in rps-challenge/spec/unit and the feature tests can be found in rps-challenge/spec/features. The ruby class files can be found in rps-challenge/lib and the embedded ruby files (ERB) can be found in rps-challenge/views. The main app.rb is found in rps-challenge.
+
+## What was used
+
+* Sinatra - web application software used to structure the framework of the project
+* Capybara - web based testing automation software
+* Rspec - a testing tool used for testing behaviour driven code
+* SimpleCov - a tool used to show test coverage on a web application
```
AndreaDiotallevi: Very good README.

spec/features/result_page_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  context "player has selected Rock" do
+    before do
+      click_on "Rock"
+    end
+    scenario "player has selected to Rock" do
+      expect(page).to have_content "You have selected Rock"
+    end
+
+    scenario "show msg for what has been selected by opponent" do
+      expect(page).to have_content "Your opponent has selected"
```
AndreaDiotallevi: Probably you should test the whole phrase. "...has selected ?"

PR #1530, opened 2019-12-02
Partial solution to rps challenge

CpaterCodes:
A lot more I wish I could have accomplished, but have laid out some basic foundations. Backend structures present to take player actions and provide randomised cpu action, with links set up to eventually take player input.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,4 +1,5 @@
 require 'sinatra/base'
+require './lib/gamesession'
 
 class Rps < Sinatra::Base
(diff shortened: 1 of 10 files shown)


app.rb
```diff
@@ -0,0 +1,20 @@
+require 'sinatra/base'
+require './lib/gamesession'
+
+class Rps < Sinatra::Base
```
HarryMumford: Nice skinny controller

PR #1546, opened 2020-01-13
rps challenge - weekend challenge

IliasMariosG:
Worked up to POST-GET-REDIRECT pattern


app.rb
```diff
@@ -0,0 +1,19 @@
...
+class RPS < Sinatra::Base
+  enable :sessions
+  get '/' do
+    erb :index
+  end
+  run! if app_file == $0
+
+  post '/enter_name' do
+    session[:marketeer_name] = params[:marketeer_name]
+    erb :name_homepage
```
ratcliffetj: Not good practice to render view within a post request route

PR #1557, opened 2020-01-13
Fergus Pull Request

narocroc123:


spec/units/player_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+  end
+
+  describe '#result' do
+    it 'returns player 1 has won' do
+      srand(45670)
+      expect(fergus.result).to eq("Fergus Wins")
+    end
+  end
+end
```
sarita-sumbria: I tried to submit my name but couldn't  see my name on the web app. There is need to debug the code.

app.rb
```diff
@@ -0,0 +1,30 @@
...
+  post '/game' do
+    @computer_move = $player.computer_move
+    @player_move = $player.move
+    @player_name = $player.name
+    @result = $player.result
+    erb :game
+  end
+
+  run! if app_file == $0
+end
```
sarita-sumbria: I tried to run the app but got an error page  after registering my name.  The error is about  wrong argument at /name.

lindseydew: I think the cause of the error is coming from:
`params[:player_name, :player_move]`

If you want to look up two variables in a hashmap it would need to be separate:
`params[:player_name], params[:player_move]`

lindseydew: Not sure if you would know what the player_move is in this case though?

spec/units/player_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  describe '#name' do
+    it 'returns the players name' do
+      expect(fergus.name).to eq('Fergus')
+    end
+  end
+
+  describe '#result' do
+    it 'returns player 1 has won' do
+      srand(45670)
```
lindseydew: I'm not sure what weapon you are selecting for this test?

PR #1560, opened 2020-01-13
RPS CHallenge

samidm13:


lib/game.rb
```diff
@@ -0,0 +1,23 @@
+class Game
+
+require_relative '../app.rb'
+
+  def game(move, compmove)
+
+    if @move == "Rock" && @compmove == "Paper"
```
lindseydew: I don't think you need the @ symbols for the variables in this context. The parameters move and compmove from the game method should mean you can refer to them as these names (we only need @ symbols to make variables accessible outside the class)

PR #1565, opened 2020-01-14
Iryna's RPS Challenge 

iryna-deans:


lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+  def self.create(player_1,computer)
+    @game = Game.new(player_1,computer)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def play(player_move, computer_move)
+    if player_move == computer_move
```
Rhodine: I like that you have written your if statement in a clear and succinct manner

app.rb
```diff
@@ -0,0 +1,41 @@
+require 'sinatra/base'
```
Rhodine: It could be a good idea to consider making an app folder for your app.rb and views folder to structure your structure your project more and make it even more readable

app.rb
```diff
@@ -0,0 +1,41 @@
...
+require './lib/player'
+require './lib/game'
+require './lib/computer'
+
+  class RPS < Sinatra::Base
```
Rhodine: Ruby class files should be named with the snake_case version of the class name (so with a class RPS, your app.rb file would be better named rps.rb)

lib/game.rb
```diff
@@ -0,0 +1,31 @@
+class Game
+
+  attr_reader :player_1, :computer
+
+  def initialize(player_1,computer)
+    @player_1 = player_1
+    @computer = computer
+  end
+
+  def self.create(player_1,computer)
```
Rhodine: Good use class methods and getting rid of the global variable

views/index.erb
```diff
@@ -0,0 +1,20 @@
+<!doctype html>
```
Rhodine: You could also consider using some CSS elements as a further step.

spec/computer_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'Computer'
+
+describe Computer do
+  subject(:computer) { Computer.new }
+
+  it 'returns scissors' do
```
Rhodine: I like that you tested checking the results from the array

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+    @game = Game.new(player_1,computer)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def play(player_move, computer_move)
```
Rhodine: I like that you created the Game classwhich calls the players and then then runs your play method.

spec/computer_spec.rb
iryna-jurkiv: Another way to test the array would be use to srand(from workshop)

PR #1566, opened 2020-01-14
David's RPS Challenge

Davidbaez87:
I could finish, because I got stock in the conditionals for when the PC wins. I was confused if to use Ruby code or use HTML. the info I found on research was to overwhelmed.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,37 @@
+require 'sinatra/base'
+class RPS < Sinatra::Base
+  enable :sessions
+  # def initialize(comp, player_1)
+  #   @pc = comp
+  #   @player_1_name = player_1
+  # end
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    session[:player_1_name] = params[:player_1_name]
+    redirect '/play'
+  end
+
+  get '/play' do
+    @player_1_name = session[:player_1_name]
+    erb :play
+  end
+
+  get '/rock' do
+    @player_1_name = session[:player_1_name]
+
+    erb :rock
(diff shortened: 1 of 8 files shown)


JCWBiles:
There's only one test.
app.rb & game.rb were not clear to see in the file tree.

app.rb
```diff
@@ -0,0 +1,37 @@
...
+  # def initialize(comp, player_1)
+  #   @pc = comp
+  #   @player_1_name = player_1
+  # end
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    session[:player_1_name] = params[:player_1_name]
```
lindseydew: Great use of sessions here!

game.rb
```diff
@@ -0,0 +1,23 @@
+class Game
+def start_game
+  choices = ["rock", "paper", "scissors"]
+  result = (choices.random() * 3)
+  if (p1 == pc)
+    puts "It's a draw"
```
lindseydew: This is a great start on implementing the game logic. How would a user be able to put their choice in here? Does the result need to be printed out with a puts statement?

app.rb
```diff
@@ -0,0 +1,37 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    @player_1_name = session[:player_1_name]
+    erb :play
+  end
+
+  get '/rock' do
+    @player_1_name = session[:player_1_name]
```
lindseydew: Could this be refactored to make the weapon choice a variable to make this code more DRY?

PR #1568, opened 2020-01-14
adding progress of challenge

CalMcD:
Callum's Friday Challenge

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,20 @@
+require 'sinatra/base'
+require './lib/player'
+
+class Rps < Sinatra::Base
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    player = Player.new(params[:player_name])
+    redirect '/play'
+  end
+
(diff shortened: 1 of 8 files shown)


views/play.erb
```diff
@@ -0,0 +1 @@
+<h2> Welcome <%= @player_name %>! </h2>
```
laek: Does this file have access to this variable? Is it linked with the player object created on /name?

app.rb
```diff
@@ -0,0 +1,20 @@
...
+    player = Player.new(params[:player_name])
+    redirect '/play'
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  run! if app_file == $0
+end
```
laek: Code is easy to follow, file structure is clear and things are easy to find in project. It seems like the test might not be passing because the name parameter is not passed from /name path to the /play path. Adding something under /play path that links it with the new player object created in /name might help with that.

PR #1570, opened 2020-01-14
RPS challenge

Rhodine:


app.rb
```diff
@@ -0,0 +1,42 @@
...
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    player_1 = Player.new(params[:player_1_name])
+    computer = Player.new(params[:computer])
+    $game = Game.new(player_1, computer)
```
iryna-jurkiv: I think here it would be good to use a class method so you don't need to use a global variable in your application controller 
https://github.com/makersacademy/course/blob/master/intro_to_the_web/killing_the_global_variable.md

app.rb
```diff
@@ -0,0 +1,42 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    @game = $game
+    erb :play
+  end
+
+  post '/rps' do
+    session[:rps] = params[:rps]
```
iryna-jurkiv: I like the way you use sessions to store user move from post & then use it in get (for your play erb) (instead of using global variables)

app.rb
```diff
@@ -0,0 +1,42 @@
+require 'sinatra/base'
```
iryna-jurkiv: I think the diagram above is great at depicting client & server requests and responses. Thank you for including it!

views/choose_rps.erb
```diff
@@ -0,0 +1,37 @@
+<div style = "background-color: pink; position: relative; center: 50px; padding: 50px; font-family: 'Comic Sans MS';">
```
iryna-jurkiv: Great use of inline CSS 👍

PR #1577, opened 2020-02-09
Challenge complete

jlblumberg:


lib/game.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def self.create(player_1, player_2)
+    @game = Game.new(player_1, player_2)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def compute_result
```
Dlibmanw: The compute_result method is really great! and well done for making it so far through this challenge!

PR #1579, opened 2020-02-09
RPS CHALLENGE

BarriF13:


lib/turn.rb
```diff
@@ -0,0 +1,32 @@
+class Turn
+  GAME_RULES = {
+    rock: {rock: :draw, paper: :lose, scissors: :win},
```
richardpattinson: I like this way of determining the winner, very scalable.

spec/features/play_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+# I would like to be able to play rock/paper/scissors
+feature 'Play the game' do
+  PLAY_SEED = 221563 # This is a option for scissors
+  before do
+    visit('/')
+    fill_in 'name', with: 'Barri'
+    click_button 'Submit'
+  end
+  # the marketeer will be presented the choices (rock, paper and scissors
+  scenario 'Can see options : Rock paper scissors' do
```
richardpattinson: Very thorough testing. V impressive.

app.rb
```diff
@@ -0,0 +1,30 @@
+require 'sinatra/base'
```
richardpattinson: Skinny controller,well done this is  best practice.

PR #1580, opened 2020-02-09
Weekend challenge

lookupdaily:
- Completed basic challenge
- Attempted styling, still in progress! Some issues with display of radio inputs.
- App currently fails if user doesn't select a move


Challenge-instructions.md
```diff
@@ -0,0 +1,86 @@
...
+
+SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
+  SimpleCov::Formatter::Console,
+  # Want a nice code coverage website? Uncomment this next line!
+  # SimpleCov::Formatter::HTMLFormatter
+])
+SimpleCov.start
+```
+
+You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
```
saidbyced: 👍 Great idea to maintain instructions while creating your own README!

app.rb
```diff
@@ -0,0 +1,44 @@
...
+require_relative 'lib/player'
+require_relative 'lib/game'
+require_relative 'lib/computer'
+
+class RockPaperScissors < Sinatra::Base
+  enable :sessions
+
+  before do
+    @game = Game.instance
+  end
```
saidbyced: 😍 love this - keepin' it DRY and removing any global variables

lib/game.rb
```diff
@@ -0,0 +1,41 @@
...
+      @winner = @player
+    elsif player_move == 'scissors' && opponent_move == 'rock'
+      @winner = @opponent
+    elsif player_move == 'scissors' && opponent_move == 'paper'
+      @winner = @player
+    elsif player_move == 'paper' && opponent_move == 'scissors'
+      @winner = @opponent
+    elsif player_move == 'paper' && opponent_move == 'rock'
+      @winner = @player
+    end
```
saidbyced: ☝️Maybe add an 'else' to catch edge cases?

app.rb
```diff
@@ -0,0 +1,44 @@
...
+    redirect '/move'
+  end
+        
+  get '/move' do
+    erb :move
+  end
+
+  get '/winner' do
+    @game.select_winner
+    @game.winner
```
saidbyced: ☝️could calling .winner cause it to select the winner and pass the result (keeping it DRY)? It might allow you to use a private method for selecting the winner too.

public/css/style.css
```diff
@@ -0,0 +1,85 @@
...
+.radio {
+  display: block;
+  width: 100%;
+}
+
+label {
+  display: block;
+  position: relative;
+  z-index: 9;
+}
```
saidbyced: 😍 adding a CSS doc made this look AWESOME

lookupdaily: the struggle of my weekend...

app.rb
lookupdaily: Turns out that second line wasn't needed! It's an attribute to use in display - thanks for spotting!

lib/game.rb
lookupdaily: good point - done!

PR #1581, opened 2020-02-09
rps, user vs computer only

Mrtly:
I followed the walkthrough to do this as I really really wanted to complete something for once 😳, and to have time to work on the front end. Simple thing, user vs computer only, but it looks pretty and it kinda works, so I'm happy.


public/style.css
```diff
@@ -0,0 +1,75 @@
+#page-container {
```
zsoobratty: Incredible - LOVE the CSS

views/play.erb
```diff
@@ -0,0 +1,34 @@
...
+        <br><br>
+        <div id="opponent"><h2>
+            Computer chose <%= @game.opp_choice.to_s.capitalize %>! </h2>
+        </div>
+    </div>
+
+    <div class="box3" id="win">
+    <div id="win">
+        <h1>
+        <% if @game.win? %>
```
zsoobratty: I like how you managed to put emoji's in! Very cool

public/phone.css
```diff
@@ -0,0 +1,40 @@
+
+#page-container {
```
zsoobratty: Again brilliant! Love how you made it work on phones too!

lib/opponent.rb
```diff
@@ -0,0 +1,7 @@
+class Opponent
```
zsoobratty: Good idea separating the classes, rather than just having it all in one.

PR #1586, opened 2020-02-10
RPS Challenge partial solution

Dlibmanw:


spec/features/make_a_choice_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'spec_helper'
+
+feature 'Making a choice' do
+    before do
```
jlblumberg: Good use of 'before' to avoid repetition in tests!

spec/round_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+require 'round'
+
+describe 'Round' do
```
jlblumberg: This whole thing is great. Very succinct, easy to read, and you're testing all aspects of the game's parts (name, player_shape, and computer_shape)

spec/spec_helper.rb
```diff
@@ -1,6 +1,9 @@
 require 'capybara/rspec'
 require 'simplecov'
 require 'simplecov-console'
+require File.join(File.dirname(__FILE__), '..', 'app.rb')
+ENV['RACK_ENV'] = 'test'
```
jlblumberg: I'm not sure if you need this line - was this taken from battle? If we do need it, what does it do?

app.rb
```diff
@@ -0,0 +1,29 @@
...
+        erb :play
+    end
+
+    post '/play' do
+        session[:player_1_shape] = params[:shape]
+        session[:computer_shape] = :Rock
+        redirect '/play'
+    end
+
+    run! if app_file ==$0
```
jlblumberg: Do we need this line? I put it too, but not really sure what it does.

PR #1589, opened 2020-02-10
RPS Challenge - Incomplete

Kefuri:


app.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  get '/' do
+    erb(:name_input)
+  end
+
+  post '/names' do
+    $player_1 = Player.new(params[:player_1])
+    $player_2 = Player.new(params[:player_2])
+
+    # session[:player_1] = params[:player_1]
```
EllieRichardsonJones: These could be deleted?

views/gameplay.erb
```diff
@@ -0,0 +1,22 @@
+<style>
```
EllieRichardsonJones: Love that there's a style bit!

app.rb
```diff
@@ -0,0 +1,33 @@
...
+require './lib/game'
+
+class App < Sinatra::Base
+
+  get '/' do
+    erb(:name_input)
+  end
+
+  post '/names' do
+    $player_1 = Player.new(params[:player_1])
```
EllieRichardsonJones: Don't use global variables!! $$$

PR #1592, opened 2020-02-10
RPS Challenge, 

Pazoia:
Up to player name input and shape selection.


spec/features/players_choice_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'spec_helper'
+
+feature 'players choice' do
+  before do
+    sign_in_and_play
```
tobydawson1: nice use of web helper

PR #1597, opened 2020-02-10
Weekend challenge #3

davmcgregor:


app/app.rb
```diff
@@ -0,0 +1,54 @@
+require 'sinatra/base'
+require './lib/player.rb'
+require './lib/game.rb'
+
+class RockPaperScissors < Sinatra::Base
+  enable :sessions
+
+  before do
+    @game = Game.instance
```
blu3skies: Good work not using global variables

app/app.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    p params
```
blu3skies: Once you program is complete you no longer need visibility of the params

PR #1604, opened 2020-03-08
RPS challenge, an approach to the solution.

rafahg:


Thatguy560:
It works fine! My only suggestion is to use the walkthrough less in the future.

PR #1608, opened 2020-03-08
RPS challenge - Partially complete

artemis-p:


lilawalker:
Appreciation: love the design of your site. Very creative! Your code is very clean and easy to read.
Suggestion: Perhaps take out the testing infrastructure page and test since you already know your site is up and working based on the other tests. You could also use a web helper for spec files for the login page instead of using 'before do' so that you can use it across all your feature tests.

PR #1609, opened 2020-03-08
RPS Challenge

lilawalker:


artemis-p:
Appreciation: Well done for the effort! I really like the structure of your code. It's clean and easy to read. I like that you've separated the CSS file from the HTML and added colors in it.
Suggestion: Perhaps add "require 'sinatra/base' on top of your file "app.rb". I found this article which gives a useful explanation: https://github.com/zhengjia/sinatra-explained/blob/master/app/tutorial_1/tutorial_1.md

PR #1610, opened 2020-03-09
RPS Challenge

plkujaw:


dariathompson:
First of all, I like how pretty you made it look on the page! Very elegant and minimalistic

And the way you made your code shorter by storing 3 variables in one
```
def initialize(options)
    @player_name = options["player_name"]
    @player_choice = options["player_choice"]
    @computer_choice = options["computer_choice"]
  end
```

Suggestions:
Work with randomness more and make your tests pass

PR #1611, opened 2020-03-09
RPS Challenge

ghost:


KG700:
I like how you've made your win conditions a constant at the top, and I like that the whole game is played on one page.

An improvement you could make is to add some emojis and css.

PR #1615, opened 2020-03-09
Karla's RPSLS Challenge

KG700:


Swarbzz:
Love the CSS and emoji implemented within your code and adding on the new gecko and Spock rules. Small suggestions you are small spelling errors and a little bit of refactoring. Maybe adding a separate class and calling on that.

PR #1617, opened 2020-03-09
RPS Challenge - Patrick

poliver24:


views/move.erb
```diff
@@ -0,0 +1,47 @@
...
+      <%=@game.player%> Wins! <%@game.player_win%>
+    <%end%>
+
+  <%elsif @move == 'Scissors'%>
+    <%if @computer == 'Rock'%>
+      Computer Wins! <%@game.computer_win%>
+    <%else%>
+      <%=@game.player%> Wins! <%@game.player_win%>
+    <% end %>
+  <%end%>
```
Kittaru87: This doesn't need to be in here - it's already in the model

Kittaru87:
I really like the option to choose a one off game,  a 3-turn or a 5-turn game. Plus the use of self in order to get rid of the global variable is a good thing (I need to do this). 

You definitely need more than 1 test though.... project should be test driven. You can also slim down your views as the functionality is already in the model

PR #1620, opened 2020-03-09
My ups challenge - Peter

Thatguy560:


rafahg:
Nice code Peter, the css style is simply inspiring.
As an improvement just to finish the app and reach full functionality.

PR #1621, opened 2020-03-09
rps challenge

WilliamJGrace:


sgneha:
**Appreciation:**
Properly working game and clean code.
**Suggestion:**
It would be good if you avoid using global variables and according to Single Responsibility Principle try using one more class apart from Player.

PR #1622, opened 2020-03-09
weekend challenge incomplete

srikaanthtb:


lib/game.rb
```diff
@@ -0,0 +1,31 @@
+class Game
+    attr_reader :choice, :computer_choice, :c_choice
+
+    options = ["Rock", "Paper", "Scissors"]
+
+    def computer_choice
+        @computer_choice = options.sample
+        @c_choice = options.sample
```
OrionReed: @c_choice appears to be redundant, I suggest removing it

PR #1623, opened 2020-04-04
ROSHAMBO

Edward-Phillips:
commit 48ac032 meets the two user stories, will commit further for work on bonus challenges/refactoring.
Had issues using Shotgun - I know there's this session_secret setting and I've chucked a line in for that, but I wasn't able to have things maintained across pages so have to launch using rackup at the moment.


app.rb
```diff
@@ -0,0 +1,36 @@
...
+    Game.current_game.player_two.pick(ARSENAL.sample)
+    Game.current_game.player_one.pick(params[:choice])
+    redirect '/results'
+  end
+
+  get '/results' do
+    @result = Game.current_game.janken
+    erb(:results)
+  end
+end
```
Gina-Frankel, line 36: good - no global variables!

spec/features/playing_roshambo_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+  scenario "after choosing to play should be able to choose paper" do
+    named_player
+    expect { choose('p') }.to_not raise_error
+  end
+  scenario "after choosing to play should be able to choose scissors" do
+    named_player
+    expect { choose('s') }.to_not raise_error
+  end
+  scenario "You should be able to tell the paper button is for paper." do
```
Gina-Frankel, line 14: small detail - in other scenario's you have used word 'option' here you used 'button'

spec/features/playing_roshambo_spec.rb
```diff
@@ -0,0 +1,54 @@
+feature "User Story 2 - Playing Roshambo - Part 1: choose your weapon: " do
+  scenario "after choosing to play should be able to choose rock" do
+    named_player
```
Gina-Frankel, line 3: named_player  method is a good way to make it more readable - perhaps could make it sound more like a verb eg - 'name_player' - as it is good practice to make method names sounds like verbs

views/challenger.erb
```diff
@@ -0,0 +1,9 @@
+<div>
+<h1>LIG<%=@ronin%>HTS</h1>
+<form action='/RO-SHAM-BO' method='post' >
+  <input type="radio" id='r' name="choice" value="rock"> <label for="rock"> Rock, an ancient but powerful weapon.</label><br>
+  <input type="radio" id='p' name="choice" value="paper"> <label for="paper">Paper, the stuff history books are made of.</label><br>
+  <input type="radio" id='s' name="choice" value="scissors"> <label for="scissors">Scissors, sharp yet fragile.</label><br>
+  <input type='submit' value ='submit' >
+</form>
+</div>
```
Gina-Frankel, line 9: Like the use of the radio button, and the description you addd :)

app.rb
```diff
@@ -0,0 +1,36 @@
+require 'sinatra/base'
+require './lib/game'
+require './lib/player'
+
+class Roshambo < Sinatra::Base
+
+  enable :sessions
+  ARSENAL = ["rock", "paper", "scissors"]
+
```
Gina-Frankel, line 9: Perhaps the ARSENAL and anything related to it (Eg line 27) should be part of the model rather than the controller?

views/index.erb
```diff
@@ -0,0 +1,6 @@
+<div>
+<h1>Greetings Challenger, Please enter your name:</h1>
+<form action = '/challenger_approaches' method = 'post'>
+<input type="type", name="player_name">
+<input type="submit", value="Play!">
+</form>
```
Gina-Frankel, line 6: There is a symbol after the form - I am not sure what it means?

spec/game_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  it " should be able to determine which of the two players in the game is the winner - Part 3" do
+    rps.player_one.stub(:weapon) { 'scissors' }
+    rps.player_two.stub(:weapon) { 'scissors' }
+    rps.player_one.stub(:name) { 'Ed' }
+    rps.player_two.stub(:name) { 'notEd' }
+    expect(rps.janken).to eq("No one, it's a draw")
+  end
+
```
Gina-Frankel, line 39: Good use of stubs to control random output

lib/game.rb
```diff
@@ -0,0 +1,27 @@
...
+    p2 = @player_two.weapon
+    if p1 == p2
+      "No one, it's a draw"
+    elsif p1 == 'rock' && p2 == 'scissors' || p1 == 'scissors' && p2 == 'paper' || p1 == 'paper' && p2 == 'rock'
+      @player_one.name
+    else
+      @player_two.name
+    end
+  end
+
```
Gina-Frankel, line 19: I did have to look up what janken means but after looking it up understand why you called it that 

janken method = very clear and readable - only thing happening in method which is good 

possibly if you wanted to be more specific in the message you could say - if player 1 did rock and computer did paper - you could say 'paper covers rock' ....

app.rb
Edward-Phillips: Yeah you're right, I probably should've extracted it to a computer class

spec/features/playing_roshambo_spec.rb
Edward-Phillips: ooh you got me - I partially refactored to not use radio buttons due to difficulties in telling capybara to click a radio button, but once I figured that out I added it back in - will hace to edit the test name!

spec/game_spec.rb
Edward-Phillips: thanks, I'm going to avoid this method of stubbing in favour of the one that you used in your work as this one pops up with a warning about using deprecated methods after running rspec

views/index.erb
Edward-Phillips: ah I think that is saying that there should be a trailing newline at the end of the file, if you run rubocop on your code you should be able to see that to follow best practice there should be a trailing newline at the end of every file - I've no clue why but it's clearly popular as github seems to like the idea too.

lib/game.rb
Edward-Phillips: Yeah I guess I'd probably then add a method which uses the results of janken to inform the phrase it uses, I was trying to add anything too fancy on to this project to stick to the basic requirements of the user story, but if I were adding more for myself I would definitely include some fluff about how you won/lost.

spec/features/playing_roshambo_spec.rb
Edward-Phillips: yeah you're right, in my head I was kind of thinking of the method as a way of saying "get to the point that you have a" almost like that method retrieves a saved state of the website, not quite the right way to think about methods.

PR #1626, opened 2020-04-05
Beam me up, Spocky

hturnbull93:
I have implemented both single and multiplayer, and lizard spock too.

Also looks alright with some reasonable css


alexakearns:
Clear TDD process throughout build. The way the game splits into one player v multiplayer is well refactored. Use of images and styling makes it simple and easy for the user.

spec/features/computer_choice_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect(page).to have_content 'Computer chose Paper'
+  end
+  
+  scenario 'Computer chooses Scissors' do
+    srand(3)
+    dave_sign_in
+    click_button('Rock')
+    expect(page).to have_content 'Computer chose Scissors'
+  end
+end
```
alexakearns, line 22: Nice coverage of all possibilities.

spec/features/enter_name_spec.rb
```diff
@@ -0,0 +1,11 @@
+feature 'Enter Name' do
+  scenario 'Enter Dave and submit, see Dave' do
+    dave_sign_in
+    expect(page).to have_content 'Dave'
+  end
+
+  scenario 'Enter Jim and submit, see Jim' do
+    jim_sign_in
```
alexakearns, line 8: Bet Jim is a legend

spec/game_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+
+  # Player_instance double
+  let(:player_inst) { double(:player_inst) }
+  
+  # Comp_class double
+  let(:comp_class) { double(:comp_class, new: comp_inst) }
+
+  # Comp_instance double
+  let(:comp_inst) { double(:comp_inst) }
+
```
alexakearns, line 16: Good use of doubles

lib/game.rb
```diff
@@ -0,0 +1,56 @@
...
+    @game = Game.new(player_1_name: player_1_name, player_2_name: player_2_name)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def initialize(player_1_name:, player_2_name: nil, player_class: Player, comp_class: Comp)
+    @player_1 = player_class.new(player_1_name)
+    @player_2 = player_2_name.nil? ? comp_class.new : player_class.new(player_2_name)
```
alexakearns, line 15: Great use of ternary operator

lib/game.rb
```diff
@@ -0,0 +1,56 @@
+require_relative 'player'
+require_relative 'comp'
+
+class Game
```
alexakearns, line 4: 🎮

spec/game_spec.rb
hturnbull93: Thanks

spec/features/computer_choice_spec.rb
hturnbull93: Thanks :)

spec/game_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+  describe '#multiplayer?' do
+    it 'when player 2 supplied, #multiplayer? true' do
+      expect(two_player.multiplayer?).to eq true
+    end
+    it 'when player 2 omitted, #multiplayer? false' do
+      expect(subject.multiplayer?).to eq false
+    end
+  end
+
+  describe '#winner' do
```
hturnbull93, line 30: @jamesjoshuahill so I wrote all the combos as tests, is this too much?

views/play.erb
```diff
@@ -0,0 +1,52 @@
...
+  <link href="/main.css" rel="stylesheet" type="text/css" />
+  <script src="https://kit.fontawesome.com/88c2a27665.js" crossorigin="anonymous"></script>
+</head>
+<body>
+  <h1><%= @game.player_1.name %> vs <%= @game.player_2.name %></h1>
+    <% if @game.multiplayer_turn_1? %>
+      <h2><%= @game.player_1.name %>'s choice. No peeking <%= @game.player_2.name %>!</h2>
+    <% elsif @game.multiplayer_turn_2? %>
+      <h2><%= @game.player_2.name %>'s choice. No peeking <%= @game.player_1.name %>!</h2>
+    <% end %>
```
hturnbull93, line 17: @jamesjoshuahill so I used a lot of chained messages here in order to keep the controller as skinny as possible (most of the time it only sets the game as an inst var. This feels like there is too much logic going on in the view? Should I have made separate views for this? That would feel like a duplication of effort if I had a separate view for singleplayer, multiplayer turn 1 and turn 2? Not sure!

lib/game.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def player_1_win_cons
+    [OPTIONS[player_1_position - 2], OPTIONS[player_1_position - 4]]
+  end
+
+  def player_1_wins?
+    player_1_win_cons.include?(@player_2.choice)
+  end
+end
```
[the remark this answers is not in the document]
jamesjoshuahill: I guessed what `same_choice?`, `player_1_wins?` meant and it looks like these methods matched my expectations. I couldn't guess what `player_1_win_cos` meant from the method name and looking at the method it's not clear to me what it does.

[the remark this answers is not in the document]
jamesjoshuahill: I notice that the `player_1_position` method is only used twice in one method. Perhaps extracting a variable would be enough?

views/play.erb
jamesjoshuahill: Good point. When I see chained messages I consider the principle of least knowledge (Law of Demeter). In this case you've reduced the number of instance variables that the view knows about to one. However, the trade off is that the view knows a lot of about `@game`. Therefore, changing the game object is more likely to require changing the view.

I wonder if introducing more instance variables could reduce how much the view knows about these objects and reduce coupling?

spec/game_spec.rb
jamesjoshuahill: I think it's great that you've covered all the combinations 💯 

There is now a lot of repetition in this test code. Could you use helper methods in this `describe` block to extract the duplication and make these tests more readable?

lib/game.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def multiplayer_turn_2?
+    @multiplayer && !@player_1.choice.nil?
+  end
+
+  OPTIONS = %w[Rock Lizard Spock Scissors Paper]
+  
+  def winner
+    same_choice? ? nil : player_1_wins? ? @player_1 : @player_2
```
jamesjoshuahill, line 36: I find this line hard to read because it has two ternary operators. How about splitting out the first condition as a guard clause?

jamesjoshuahill (commented):
Great work @hturnbull93! 🎉 I hope these comments help you find ideas for some further improvements to try.

lib/game.rb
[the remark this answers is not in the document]
hturnbull93: That's a good idea about the variable, I guess I got trigger-happy on the extractions.  To summarise `player_1_win_cons` sets an array with the choices that an opponent could make that would mean that player 1 wins. I got the idea by looking at this [diagram](https://i.stack.imgur.com/ltfS9m.png), and noting that any choice will beat the next choice clockwise or two choices behind (anti clockwise). But I didn't know how to give that an elegant name as a method. :man_shrugging:

views/play.erb
hturnbull93: Right right. I guess I got too caught up in making the controller skinnier that I made the view fatter. So the paradigm is to push down from view to controller to model, now away from controller to model and view.

spec/game_spec.rb
hturnbull93: Just so I'm thinking correctly, I could define a helper method that takes the options for the choices and sets up an it block with the description string interpolated perhaps and the parameters passed to the expect?

lib/game.rb
hturnbull93: Good idea, thanks :+1:

lib/game.rb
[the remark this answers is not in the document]
jamesjoshuahill: Great diagram! Perhaps where it gets tricky to follow is the use of indexes in `OPTIONS`. The 2 and 4 are magic numbers that don't make sense without the diagram. One way we could test the design is by thinking about how easy it would be to add or remove a choice?

I wonder if there is a clearer way to represent the relationship between a choice and its winning combinations?

spec/game_spec.rb
jamesjoshuahill: You can define a ruby method within a `describe` block and use it inside your `it` blocks.

views/play.erb
jamesjoshuahill: Heh! It's a balancing act 😄

PR #1627, opened 2020-04-05
RPS-challenge

AlexLewis10:


lib/game.rb
```diff
@@ -0,0 +1,17 @@
+class Game
+
+attr_reader :player, :computer_turn
+
+  def initialize(player, computer=['rock', 'paper', 'scissors'].sample)
+    @player = player
+    @computer_turn = computer
+  end
+
+  def decide_winner
```
FayeCarter, line 10: It's incredible you managed to get the game work with one method

lib/game.rb
```diff
@@ -0,0 +1,17 @@
...
+    return "It's a draw" if @player == @computer_turn
+    return "player wins" if @player == "rock" && @computer_turn == "scissors"
+    return "player wins" if @player == "scissors" && @computer_turn == "paper"
+    return "player wins" if @player == "paper" && @computer_turn == "rock"
+    return "computer wins"
+  end
```
FayeCarter, line 16: At the moment the method does four jobs, understanding a win a draw an a loss. You could set up an array of winning cases e.g. [rock, scissors], [scissors, paper], [paper, rock] and see if the players choices are included within that array

FayeCarter: What would happen if player entered 'Rock' instead of 'rock'

spec/game_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(game.computer_turn).to eq "paper"
+    end
+
+    it "#computer_turn makes a random selection from rock, paper or scissors" do
+      expect(["rock", "paper", "scissors"]).to include(game.computer_turn)
+    end
+  end
+
+  describe "#decide winner" do
+    let(:game2) { Game.new("rock", "rock") }
```
FayeCarter, line 31: I like the way you've done these 3 tests. It's very clear what's happening

lib/game.rb
```diff
@@ -0,0 +1,17 @@
+class Game
+
+attr_reader :player, :computer_turn
+
+  def initialize(player, computer=['rock', 'paper', 'scissors'].sample)
+    @player = player
```
FayeCarter, line 6: Could you break these down into separate classes?

lib/game.rb
AlexLewis10: > What would happen if player entered 'Rock' instead of 'rock'

I will be making this into a button.

PR #1629, opened 2020-04-06
Challenge complete

tommyrharper:


lib/game.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'player'
+require_relative 'computer'
+
+class Game
+
+  attr_reader :result, :player, :computer
+
+  RULES = { rock: :scissors,
```
PhilipVigus, line 8: That's a clever way of setting up the move comparison

spec/computer_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'computer'
+
+describe Computer do
+  subject(:computer) { Computer.new }
+
+  it 'selects randomly either rock, paper or scissors' do
```
PhilipVigus, line 6: An alternate way of doing this would be to use https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/include-matcher#array-usage

spec/features/features_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  scenario 'fill in form and submit' do
+    enter_name
+    expect(page).to have_content 'Tom'
+  end
+end
+
+feature 'I want to be able to choose rock' do
+  scenario 'select name and then rock' do
+    enter_name
+    expect(page).to have_selector(:link_or_button, 'rock')
```
PhilipVigus, line 11: You can use to have_field('rock) here I think?

spec/features/features_spec.rb
```diff
@@ -0,0 +1,62 @@
+feature 'I want there to be a form I can fill in and see my name up in lights after' do
```
PhilipVigus, line 1: I would look at reorganising your feature tests. I generally have one feature per user story, with each feature in its own file. All the scenarios related to that feature then go within that feature block.

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,15 @@
+def enter_name
```
PhilipVigus, line 1: Good use of web helpers

spec/game_spec.rb
```diff
@@ -0,0 +1,83 @@
+require 'game'
```
PhilipVigus, line 1: Really well laid out and thorough tests

spec/player_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'player'
```
PhilipVigus, line 1: What would happen if the name specified was an empty string?

app/rps_web.rb
```diff
@@ -0,0 +1,43 @@
+require 'sinatra/base'
```
PhilipVigus, line 1: Nice clear logic with your routes. It's really easy to see how a user would use the app and the flow they would go through as they are playing.

lib/computer.rb
```diff
@@ -0,0 +1,13 @@
+class Computer
```
PhilipVigus, line 1: I like how you have encapsulated this functionality into its own class

lib/computer.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  attr_reader :move
+
+  def initialize
+    @move = [:rock, :paper, :scissors].sample
```
PhilipVigus, line 6: I'm not sure you need this line that initialises @move here? Do you access @move before you call the change_move method?

PhilipVigus:
Generally a well laid out project that I found nice and easy to understand. The only real issue I had was the way you had all of your feature tests in one file, and there's a separate comment that explains how I would have done it that might make things easier to navigate.

I'm not seeing any particular issues with your feature test coverage. The way I approached them was that they don't need to cover every single scenario, as that's what the unit tests on individual model classes do. I might be wrong here though...

PR #1630, opened 2020-04-06
rps challenge

Gina-Frankel:


app.rb
```diff
@@ -0,0 +1,44 @@
...
+class Game < Sinatra::Base
+  enable :sessions
+
+  get '/' do 
+    @welcome_message = 'Welcome to rock paper scissors'
+    erb :index
+  end 
+
+  post '/names' do
+    $player_1 = Player.new(params[:player_1])
```
Edward-Phillips, line 13: Global variables - It's possible to replace this so they're not used, as apparently they're not very popular. This is covered in the last step of last week's challenge. I recommend having a read of the walkthrough and the resources they link as they're very informative.
 https://github.com/makersacademy/course/blob/master/intro_to_the_web/killing_the_global_variable.md

lib/computer.rb
```diff
@@ -0,0 +1,8 @@
+class Computer
+  def initialize
+    
+  end
+  def select_object
```
Edward-Phillips, line 5: Nice, I like that you've got the thing which makes the computer uniquely a computer stored in it's own class, seems a much better place for it!

lib/player.rb
```diff
@@ -0,0 +1,53 @@
+require_relative 'computer.rb'
+
+class Player
+
+  attr_reader :opponent_attack, :name, :attack, :player_total, :computer_total
+  def initialize(name)
+    @name =name
+    @message
+    @player_total = 0
```
Edward-Phillips, line 9: Ah nice I like having a way of incrementing each player's score!

lib/player.rb
```diff
@@ -0,0 +1,53 @@
...
+    @computer_total = 0
+  end
+
+  def play(object)
+    @attack = object
+    compare(Computer.new.select_object)
+  end 
+
+
+  def compare(computer_object)
```
Edward-Phillips, line 19: it feels like the player class is doing a lot here, it might be worth extracting some of the logic to another class to adhere to the SRP.

spec/computer_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'computer'
+
+describe Computer do 
+
+  describe '#computer_attack_type' do  
+    computer = Computer.new
```
Edward-Phillips, line 6: Good tests, tests exactly what you need, I like the way that you've tested to see if it's one of the three, but have you forced it to be random by doing this?

spec/player_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+  end
+
+  context "#play('Rock')" do
+    it "if player selects Rock attack is set to 'Rock' " do
+      player.play('Rock')
+      expect(player.attack).to eq "Rock"
+    end
+    it 'if computer also selects Rock' do 
+      allow_any_instance_of(Computer).to receive(:select_object).and_return('Rock')
```
Edward-Phillips, line 19: nice stubbing, I like this! I'd not known about allow_any_instance_of, seems nice and powerful, I ended up doing a deprecated method instead of this so shall definitely be adding this to my list of refactoring

views/round_1.erb
```diff
@@ -0,0 +1,6 @@
+<h3> You entered <%=@attack%> </h3>
```
Edward-Phillips, line 1: I like that your ERBs are nice and skinny, no excessive logic on the view level, very nice!

spec/computer_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'computer'
```
Edward-Phillips, line 1: Solid spec files, excellent test coverage and you don't really seem to be testing for testing's sake either, nice work!

app.rb
```diff
@@ -0,0 +1,44 @@
...
+get '/play' do 
+  @player_1 = $player_1.name
+  erb :play
+end 
+
+# if this is a post it should not have erb :round_1
+# where should it go instead? 
+# or should this be post '/play'
+post '/round_1' do 
+  $player_1.play(params[:attack])
```
Edward-Phillips, line 26: is it possible to push this logic to the model level?

README.md
```diff
@@ -1,5 +1,11 @@
 # RPS Challenge
 
+## Domain model 
+Please see domain model - (https://miro.com/app/board/o9J_kux70qo=/)
```
Edward-Phillips, line 4: EXCELLENT DOMAIN MODEL, I LOVE IT! Very informative about your process and clearly describes the expected behaviour of the model

spec/computer_spec.rb
Gina-Frankel: **Ed:**
 "so I guess I'd structure it like this:
call the method,
Check to see that the result of the method is included in the rock paper scissors array,
push that to your results array,
loop that 500x or whatever number you like,
then expect it to not have 500x rock, not to have 500x paper and not to have 500x scissors
that way you know it's pushing at least two different options to the array
doesn't have to push three
I mean technically it's possible to push rock 500 times, but if you look at a normal distribution you can see that with large numbers of tests that's very unlikely
not the best as you'd be doing 4 expects in a single test
I guess you can split it up
have 1 test which is if it's included in the array
and then 3 tests for each of the not to be entirely rock/paper/scissors
though that's not perfect as it could then push 500x of the another type for each of the three tests"

lib/computer.rb
```diff
@@ -0,0 +1,8 @@
+class Computer
+  def initialize
```
tatsiana-makers, line 2: This initialize method doesn't look to be doing anything, do we need to keep it?

spec/features/choose_attack_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+scenario 'select Scissors' do
+  sign_in
+  click_button('Scissors')
+  expect(page).to have_content('You entered Scissors')
+  end
+end
+
+
+# Do I need tests for all of them?
```
tatsiana-makers, line 23: Great question! What's your gut feeling about this?

spec/player_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'player'
```
tatsiana-makers, line 1: Nice coverage of all of the possible scenarios!

PR #1633, opened 2020-05-03
Rock paper scissors lizzie

LTurns:


controller.rb
```diff
@@ -2,30 +2,30 @@
...
+  attr_reader :player, :result
 
   get "/" do
     erb :index
   end
 
   post '/game' do
-    @player = params["player"]
+    session[:player] = params["player"]
     redirect "/game?name=#{@player}"
```
Tracht, line 13: that's pretty cool, so is this the way to get parameters to persist across routes?

controller.rb
```diff
@@ -2,30 +2,30 @@
...
-    @player = params["player"]
+    @player = session[:player]
     erb :play
   end
 
   post '/result' do
-    @player = params["player"]
+    @player
     @result = params[:result]
     redirect "/result?name=#{result}name=#{@player}"
```
Tracht, line 24: do you know why you didn't need @ for result but you do for player?

PR #1637, opened 2020-05-03
Zsofi's RPS challenge solution

ZsofiaS:


app.rb
```diff
@@ -15,21 +15,22 @@ class RPS < Sinatra::Base
     redirect "/play"
   end
 
-  get "/play" do
+  before do
     @game = Game.instance
```
Tracht, line 19: Nice use of class instances

Tracht, line 19: nice use of class instance

lib/game.rb
```diff
@@ -1,6 +1,14 @@
 class Game
   attr_reader :player, :computer, :draw, :winner
 
+  def self.create(player, computer)
```
Tracht, line 4: nice refactor to use class instance

app.rb
```diff
@@ -15,21 +15,22 @@ class RPS < Sinatra::Base
     redirect "/play"
   end
 
-  get "/play" do
+  before do
```
Tracht, line 18: nice use of before hook

PR #1638, opened 2020-05-03
Catriona's RPS challenge

cmb84scd:
Unfortunately this is only part done as I just couldn't get it to save the player's choice. I tried different ways of allowing the player to make their choice but it would not save and I just could not work out why. I'm not particularly happy about this and feel I could have done a lot better.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,8 @@
 require 'sinatra/base'
 
-class RPS < Sinatra::base
+class RPS < Sinatra::Base
   get '/' do
-    
+    'Welcome to RPS!'
   end
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -2,7 +2,12 @@ require 'sinatra/base'
 
 class RPS < Sinatra::Base
   get '/' do
-    'Welcome to RPS!'
+    erb(:index)
+  end
+
+  post '/play' do
+    @player = params[:player]
+    erb(:play)
   end
 
   run! if app_file == $0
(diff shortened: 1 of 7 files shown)


app.rb
```diff
@@ -0,0 +1,32 @@
...
+
+class RPS < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/play' do
+    session[:player] = params[:player]
```
dpwdec, line 11: Good use of sessions hash to persist player state.

spec/features/enter_name_spec.rb
```diff
@@ -0,0 +1,28 @@
+feature 'Enter name' do
+  scenario 'User can enter name and see it' do
+    visit('/')
+    fill_in :player, with: 'Catriona'
+    click_button 'Submit'
```
dpwdec, line 5: Code extract these duplicated feature actions into a helpers file.

views/play.erb
```diff
@@ -0,0 +1,12 @@
...
+</style>
+<h2>Welcome <%= @player %>!</h2>
+
+<h3>To play, please click on your choice</h3>
+
+<form action="/result" method="post">
+  <input type="submit" name="move" value="rock">
+  <input type="submit" name="move" value="paper">
+  <input type="submit" name="move" value="scissors">
+</form>
```
dpwdec, line 12: Nice and clear keeping everything in one form.

PR #1640, opened 2020-05-04
Submit rps-challenge

will-head:


app.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  post '/move' do
+    verbose_output(request.url) if $verbose
+ 
+    session[:players][0].move = params[:commit]
+    session[:players][1].move = ['🗿', '📄', '✂'].sample
+
+    redirect to('result')
+  end
+
```
mbrad26, line 44: This controller is a bit too fat :). I think you can move some of the logic in the view if  you keep it as a simple if else statement.

app.rb
```diff
@@ -0,0 +1,75 @@
...
+  get '/result' do
+    verbose_output(request.url) if $verbose
+
+    game = RockPaperScissors.new
+    result = game.result(session[:players][0], session[:players][1])
+
+    case result
```
will-head, line 51: Move this logic into RockPaperScissors class

PR #1644, opened 2020-05-04
RPS with option for 2 player :)

ellistrickett:


lib/calculate.rb
```diff
@@ -0,0 +1,27 @@
+class Calculate
```
tatsiana-makers, line 1: I wonder if something like "Game" might be a clearer name for this class?

spec/scoreboard_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'scoreboard'
```
tatsiana-makers, line 1: Really nice thorough testing 🙌

PR #1646, opened 2020-05-04
Week 3 weekend challenge submission

hiboabd:


lib/game.rb
```diff
@@ -0,0 +1,40 @@
+class Game
```
cjm106, line 1: I like your game code!
Its really clear what each part does and how they're linked to each other.

app.rb
```diff
@@ -0,0 +1,60 @@
+require 'sinatra/base'
```
cjm106, line 1: At present its possible for the user to type in a random string and loose the game.  I'd consider adding an input checker to your Game class that prompts the user to type in Rock Paper or Scissors.

views/index.erb
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
```
cjm106, line 1: I really like the fact that you've spent some time on the front end as well as the game.  
I particularly like the RPS gif which is a very nice addition!

spec/features/outcome_spec.rb
```diff
@@ -0,0 +1,32 @@
+# require 'spec_helper'
```
cjm106, line 1: Without running the code its not clear why this test has been hashed out.  
Is it a depreciated test or is something not working?
If its not working I guess a note might be helpful otherwise I would remove it to avoid confusion.

spec/features/web_helper.rb
```diff
@@ -0,0 +1,6 @@
+def sign_in_and_play
```
cjm106, line 1: The addition of a web_helper for your feature tests is a really nice addition. It really helps keep your tests free of clutter and easy to read.

PR #1647, opened 2020-05-04
Very Basic RPS challenge

collier-jo:


app.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  get "/outcome_rock" do
+    @move = "Rock"
+    testing = Rps.new
+    @result = testing.calculate_result(@move)
+    @comp_move = testing.compmove
+    erb(:outcome)
+  end 
+
+  get "/outcome_paper" do
```
AndrewHulme, line 31: Would be more DRY to have one page called /outcome which combines rock, paper and scissors (can be done by adding a 'name' parameter to the submit button in the view)

lib/rps.rb
```diff
@@ -0,0 +1,40 @@
+class Rps
```
AndrewHulme, line 1: Good use of extracting responsibilities to another class. Could rethink the name of the class as it is similar to the controller class name

lib/rps.rb
```diff
@@ -0,0 +1,40 @@
...
+    elsif user_move == "Rock" && @cm == "Rock"
+      return "Draw"
+    elsif user_move == "Paper" && @cm == "Paper"
+      return "Draw"
+    else
+      return "Loser"
+    end
+  end 
+
+  def compmove
```
AndrewHulme, line 36: Could change the name of @cm to @compmove then access it through the accessor (to get rid of the method)

spec/features/move_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end 
+
+  scenario "player picks rock" do
+    sign_in_and_play
+    click_button("Scissors")
+    expect(page).to have_content("You have selected Scissors")
+  end 
+
+  scenario "Play again button" do 
+    sign_in_and_play
```
AndrewHulme, line 24: Good use of web helper

PR #1649, opened 2020-05-31
My rock, paper, scissors web app

Sumner1185:
Struggled with class methods so had to resort to sticking with sessions in order to re-use variables :( 
However done some more reading and feel I could tackle this again with more success this time!


lib/game.rb
```diff
@@ -0,0 +1,23 @@
+class Game
+  attr_reader :player_turn, :computer_turn, :player_name, :choices
+  WINNERS = [["Scissors", "Paper"], ["Paper", "Rock"], ["Rock", "Scissors"]]
+  LOSERS = [["Paper", "Scissors"], ["Rock", "Paper"], ["Scissors", "Rock"]]
+
+  def initialize(player_name, player_turn, computer_turn)
```
tatsiana-makers, line 6: Rather than passing in strings into this `initialize` method, could we pass in instances of a `Player` and a `Computer`? This way, these two instances own their information rather than exposing it and we're making use of the encapsulation offered by having these classes. 

This should help clear up some of the logic in the `'/first_game'` route 🙂

lib/player.rb
```diff
@@ -0,0 +1,8 @@
+class Player
+  attr_reader :choice
+
+  def initialize(choice)
```
tatsiana-makers, line 4: What other information does a Player contain in addition to a choice?

spec/features/first_game_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'computer'
+
+feature 'game of rock, paper, scissors' do
+  let(:computer) { Computer.new.computers_turn }
```
tatsiana-makers, line 4: You could explore using `srand` to seed the randomness
https://gist.github.com/Kotauror/8f4a9913c53dd479df91909f306e6731 provides an example

Once you know which option the computer will pick, you can strengthen the assertions in these tests

README.md
```diff
@@ -36,51 +27,16 @@ Hints on functionality
 - the game will choose a random option
 - a winner will be declared
 
+# Screenshots of final site:
```
tatsiana-makers, line 30: It's really nice that you included these!

lib/game.rb
Sumner1185: That makes much more sense!

PR #1652, opened 2020-05-31
RPSSL working

RaeRachael:


app.rb
```diff
@@ -0,0 +1,56 @@
...
+    @name_2 = session[:game].player2
+    erb(:move2)
+  end
+
+  post '/selected_move2' do
+    session[:game].player_2_move = params[:move_2]
+    redirect('/result')
+  end
+
+  get '/result' do
```
ad13380, line 44: Good use of sessions to carry over information between routes, maybe it would help to create a Player class to reduce the number of instance variables you need to define here

lib/game.rb
```diff
@@ -0,0 +1,39 @@
...
+    return "#{player1} won" if who == "player1"
+    return "#{player2} won" if who == "player2"
+  end
+
+  private
+
+  def random_move
+    return ["Rock", "Paper", "Scissors"].sample if @rpssl == "false"
+    
+    ["Rock", "Paper", "Scissors", "Spock", "Lizard"].sample
```
ad13380, line 29: Huge props for making it to bonus 2, if only I had the same level of commitment

lib/game.rb
```diff
@@ -0,0 +1,39 @@
...
+    who = result
+    return "draw" if who == "none"
+    return "#{player1} won" if who == "player1"
+    return "#{player2} won" if who == "player2"
+  end
+
+  private
+
+  def random_move
+    return ["Rock", "Paper", "Scissors"].sample if @rpssl == "false"
```
ad13380, line 27: May be a good idea to make these symbols instead of strings, it saves on memory (apparently)

spec/features/multiplayer_spec.rb
```diff
@@ -0,0 +1,35 @@
+feature 'multiplayer' do
```
ad13380, line 1: feature tests are very well set out, easy to see exactly what you're testing

spec/features/multiplayer_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  feature "get '/move'" do
+    scenario 'has selection for RSP choice' do
+      p1_name
```
ad13380, line 5: using a method in a web_helpers file could help keep these tests more DRY

spec/features/setup_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    scenario 'has selection for RSP choice' do
+      p1_name
+      submit
+      expect(page).to have_content('Rae what is your move?')
+    end
+  end
+
+  feature "get /result" do
+    scenario 'output of who won' do
+      allow_any_instance_of(Array).to receive(:sample) { "Scissors" }
```
ad13380, line 19: Nice idea! I was stuck on this problem for a while

spec/game_spec.rb
```diff
@@ -0,0 +1,17 @@
+require_relative '../lib/game.rb'
+
+describe Game do
+  let(:game) { Game.new("false", "Rae") }
+
+  describe '#result_string(player_move)' do
+    it "returns won, drew or lost for player" do
+      allow_any_instance_of(Array).to receive(:sample) { "Rock" }
```
ad13380, line 8: as tedious as it sounds, probably these should be broken into 3 separate tests, each with their own individual expect statement

spec/features/multiplayer_spec.rb
ad13380: I just realised you already have one, could be useful to create additional methods that have p1_name, p2_name etc inside them

views/move.erb
```diff
@@ -0,0 +1,23 @@
+<% if @rpssl == "true" %>
```
ad13380, line 1: Nice idea using an if statement so that you can store the erb of both game modes

PR #1653, opened 2020-05-31
Complete challenge

ChrisCooney05:
Finally managed to get the CSS file to work! instead of moving onto the bonus challenges, I spent time making sure I fully understood what I was doing and how everything works.


lib/play_game.rb
```diff
@@ -0,0 +1,53 @@
+class PlayGame 
+
+  RPSARRAY = ['Rock', 'Paper', 'Scissors']
```
nwmbutler, line 3: Computer class?

lib/play_game.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  attr_reader :player_choice, :computer_choice
+
+  def initialize(choice)
+    @player_choice = choice
+    @computer_choice = RPSARRAY.sample 
+  end
+
+  def play
```
nwmbutler, line 12: Is there a more concise way?

PR #1654, opened 2020-05-31
Nikita's RPS Challenge

NikitaDouglas:


README.md
```diff
@@ -1,86 +1,42 @@
...
+2. Install bundler (take a look [here](https://bundler.io/) if you don't know how to do that)
+and then run `bundle install` to install the dependencies.
+3. In the app's root directory, run `rackup -p 4567` on the command line to start the server.
+4. Visit http://localhost:4567 in your browser  to play the game.
 
-```
-As a marketeer
-So that I can see my name in lights
-I would like to register my name before playing an online game
+The tests were created with RSpec and Capybara. Run `rspec` on the command line when you're in the root directory to run the tests.
```
BeneArinci, line 21: Cool. You're using your readme file very well. Here but not only!

lib/rps_game.rb
```diff
@@ -0,0 +1,39 @@
...
+    elsif choice == 'Paper' && computer_choice == 'Rock'
+      message(:win)
+    elsif choice == 'Scissors' && computer_choice == 'Paper'
+      message(:win)
+    else
+      message(:loss)
+    end
+  end
+
+  def message(outcome)
```
BeneArinci, line 29: I would suggest to split this class in multiple classes. Everything is happening here and it has too many responsibilities. Player and, maybe, a random choice classes might help.

spec/rps_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    describe '#play_game(choice)' do
+      it 'returns a statement showing player has won' do
+        player_wins_outcomes = { "Rock" => "Scissors", "Paper" => "Rock", "Scissors" => "Paper" }
+        player_wins_outcomes.each do |choice, computer_choice|
+          allow(game).to receive(:random_choice) { computer_choice }
+          expect(game.play_game(choice)).to include("You win")
+        end
+      end
+    end
+  end
```
BeneArinci, line 23: Cool. I am going to copy this :) Really interesting

PR #1658, opened 2020-06-01
Well, at least it's something...

TimCPB:
Many issues:

1. Uses global variables
2. Couldn't figure out how to allow for randomness and different outcomes in a capybara test so Some of the feature spec testing isn't great
3. Within the Player class the @result variable and attr_reader are probably now unnecessary
4. The entire CompMove class and spec file is now unnecessary
5. There's almost no HTML formatting at all!


lib/comp_move.rb
```diff
@@ -0,0 +1,9 @@
+class CompMove
```
sofyloafy, line 1: You've named your files well and in accordance with the dictum from on high

lib/comp_move.rb
```diff
@@ -0,0 +1,9 @@
...
+
+  attr_reader :move
+
+  def initialize
+    @move = ["rock", "paper", "scissors"].sample
```
sofyloafy, line 6: This would be better as constant with symbols:

lib/player.rb
```diff
@@ -0,0 +1,32 @@
+class Player
```
sofyloafy, line 1: I think this class could have been split out as it is performing two functions - I think your name could have just gone into a name erb form and stayed within that. And then this file could _just_ be performing your if-elses.

views/results.erb
```diff
@@ -0,0 +1,2 @@
+<h1>The results are in aaaaaaaand...</h1>
+<h2><%= @results %></h2>
```
sofyloafy, line 2: It seems as though they want to see this function happening in the previous method:

Screenshot 2020-06-01 at 10 42 20

spec/player_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  #     allow_any_instance_of(CompMove).to receive(:move).and_return("scissors")
+  #     john.play_rps("rock")
+  #     expect(john.result).to eq("Congrats, your rock beat scissors!")
+  #   end
+  # end
+
+  describe "#play_rps" do
+    it "Returns a winner" do
+      john.play_rps("rock")
+      expect(john.result).to eq("Congrats, your rock beat scissors!").or eq("You both chose rock... It's a draw!").or eq("Unlucky, your rock lost to paper :(")
```
sofyloafy, line 24: I believe you could have stubbed this to test for different outcomes:

Screenshot 2020-06-01 at 10 45 19

views/index.erb
```diff
@@ -0,0 +1,20 @@
+<head>
+<style>
+h1 {text-align: center;}
```
sofyloafy, line 3: Looks neat here, nice CSS

app.rb
```diff
@@ -0,0 +1,35 @@
...
+class Rps < Sinatra::Base
+
+  enable :sessions
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/name' do
+    $player_name = Player.new(params[:player_name])
```
sofyloafy, line 14: Maybe this line could have gone in the method below?

PR #1659, opened 2020-06-01
Weekend Challenge, RPS

sofyloafy:


spec/unit_test_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect(page).to have_content("What is your move?\n🤘 Rock 🤘\n📰 Paper 📰\n✂️ Scissors ✂️")
+  end
+
+  scenario "Player chooses rock, and the computer chooses paper." do
+    visit '/play'
+    allow_any_instance_of(Computer).to receive(:move).and_return(:Rock)
+    click_button('Paper')
+    expect(page).to have_content("The computer chose: Rock\nAnd you chose: Paper\nWhich means...")
+  end
+end
```
TimCPB, line 22: File structure? This is perhaps supposed to be inside a "features" folder with some of these separated into different spec files within that "features folder"

spec/unit_test_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  click_button("Let's play!")
+  expect(page).to have_content("Welcome,\n💡 Sophie 💡")
+  end
+
+  scenario "Move options" do
+    visit '/play'
+    expect(page).to have_content("What is your move?\n🤘 Rock 🤘\n📰 Paper 📰\n✂️ Scissors ✂️")
+  end
+
+  scenario "Player chooses rock, and the computer chooses paper." do
```
TimCPB, line 16: Apparently we're meant to have tests checking for the different win, lose or draw scenarios...

spec/result_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'result'
+
+describe Result do
+  it 'should print a draw' do
```
TimCPB, line 4: Empty test here

spec/computer_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'computer'
+
+describe Computer do
+  it "should return a random move" do
+    allow(subject).to receive(:move) { "Rock" }
+    expect(subject.move).to eq("Rock")
+  end
```
TimCPB, line 7: Apparently we're supposed to test all the outcomes, though that seems like overkill

lib/result.rb
```diff
@@ -0,0 +1,22 @@
...
+    elsif @comp_choice == "Rock" && @move == "Scissors"
+      "You lost!"
+    elsif @comp_choice == "Rock" && @move == "Paper"
+      "You won!"
+    elsif (@comp_choice == "Paper" && @move == "Scissors") || (@comp_choice == "Scissors" && @move == "Rock")
+      "You won!"
+    elsif (@comp_choice == "Scissors" && @move == "Paper") || (@comp_choice == "Paper" && @move == "Rock")
+      "You lost!"
+    end
+  end
```
TimCPB, line 21: Think we're supposed to have the results string separated out into another layer/class, but I'm not totally sure

README.md
```diff
@@ -1,23 +1,27 @@
...
-
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Task
+This challenge involved meeting the user stories below and implementing the web app game in our own interpretation.
+<br>
```
tatsiana-makers, line 3: I love the styling of your web app! ✨

app.rb
```diff
@@ -0,0 +1,32 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    @player_1 = session[:player_1]
+    erb :play
+  end
+
+  post '/game' do
+    $comp_choice = Computer.new.move
```
tatsiana-makers, line 25: Do the variables inside this route need to be global variables?

spec/result_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'result'
+
+describe Result do
```
tatsiana-makers, line 3: What are the methods that we can test for the Result class?

lib/result.rb
```diff
@@ -0,0 +1,22 @@
+class Result
+  attr_reader :comp_choice, :move
+
+  def initialize(comp_choice, move)
```
tatsiana-makers, line 4: Rather than passing in the computer's choice, which is a string, what do you think about passing in the instance of the Computer class? This allows us to make use of the encapsulation provided by having that class.

lib/result.rb
```diff
@@ -0,0 +1,22 @@
...
+    @comp_choice = comp_choice
+    @move = move
+  end
+
+  def results
```
tatsiana-makers, line 9: What do you think about trying to capture this logic using hashes? It might make the method more readable

PR #1664, opened 2020-06-01
Joshua's RPS Challenge pull.

joshuaabrookuk:
This challenge was goof for me. I ran into the same problem that I did last week with Battle. 
I need to make progress with this and plan to put extra time in to understand and hopefully complete this.


spec/feature/name_on_form_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'sinatra'
+
+feature "Enter name on form" do
+  scenario "display form" do
+    visit '/'
+    fill_in :player, with: 'Jeff'
```
perrinjack, line 6: This test will 'not be able to find the form and fill in the player field' since the views directory is not located within the projects root. Instead it was located within the spec directory. Easy error, but just emphasises importance of setting up the project correctly at the start.

spec/feature/testing_infrastructure_spec.rb
```diff
@@ -0,0 +1,8 @@
+# require 'sinatra'
+#
+# feature "Testing infrastructure working!" do
```
perrinjack, line 3: As mentioned I would keep this test and make it relevant to the index page i.e ' Welcome to the game!' Think you wouldve had done this had you not been concentrating on fixing the previous mentioned bug.

PR #1665, opened 2020-06-01
rps-challenge

ooduola:
Hi there,

Here is my partially completed challenge. 

Thanks,
Ola


views/play.erb
```diff
@@ -0,0 +1,4 @@
+<h1>Online Game</h1>
+<h2>Rock, Paper, Scissors</h2>
+
+Player name: <%= @player %>
```
RichEwin: good use of HTML here

PR #1666, opened 2020-06-01
Week 3 RPS Weekend challenge

ad13380:
Single and Multiplayer functionality included


app.rb
```diff
@@ -0,0 +1,71 @@
...
+  before do
+    @game = Game.instance
+  end
+  
+  get '/' do
+    erb :index
+  end
+
+  get '/start' do 
+    erb "start_#{params[:mode]}".to_sym
```
RaeRachael, line 16: I like this idea a lot (not looked at anything else yet so no idea how its used but exciting)

lib/game.rb
```diff
@@ -0,0 +1,66 @@
...
+    @game
+  end
+
+  # instance methods
+
+  attr_reader :player_1, :player_2
+
+  def initialize(player_1, player_2, winner = Winner.new)
+    @player_1 = player_1
+    @player_2 = player_2
```
RaeRachael, line 20: do we need these and the array @players?

spec/computer_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      allow(Game::MOVES).to receive(:sample) { :paper }
+      computer.pick_move
+      expect(computer.move).to eq :paper
+    end
+
+    it 'returns the move :paper' do
+      allow(Game::MOVES).to receive(:sample) { :scissors }
+      computer.pick_move
+      expect(computer.move).to eq :scissors
+    end
```
RaeRachael, line 36: I'm not sure what having three tests here achieve that one of them doesn't

lib/winner.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+  end
+
+  private
+
+  def check_moves(players)
+    players.each do |player| 
+      fail "Invalid move given in Winner class" unless Game::MOVES.include?(player.move)
+    end
+  end
```
RaeRachael, line 27: do we need to check the moves, if the only possible options are valid? (button selected?)

spec/features/result_multi_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    click_button 'Rock'
+    expect(page).to have_content "IT'S A DRAW"
+  end
+
+  scenario 'has an option to return to play page' do
+    sign_in_and_play_moves
+    click_button 'Play Again'
+    expect(page).to have_content 'STEVE, SELECT YOUR MOVE'
+  end
+end
```
RaeRachael, line 29: super clean :)

spec/game_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it 'returns the Player instance of player_2' do
+      expect(game.player_2).to eq player_2
+    end
+  end
+
+  describe '#assign_move' do
+    it 'assigns a move to a player' do
+      expect(player_1).to receive(:move=).with(move)
+      game.assign_move(player_1, move)
+    end
```
RaeRachael, line 31: is this a test?

spec/player_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  describe '#move' do
+    it 'can be assigned a move and returns it' do
+      player.move = :rock
+      expect(player.move).to eq :rock
+    end
+
+    it 'raises an error when an invalid move is assigned' do
+      expect { player.move = "" }.to raise_error("Invalid move assignment in Player class")
+    end
```
RaeRachael, line 27: again I don't see how this is a possible state to get in

views/play_multi.erb
```diff
@@ -0,0 +1,61 @@
...
+  <form action="/move-multi" method="post">
+    <button type="submit" name="move" value="rock">Rock</button>
+    <button type="submit" name="move" value="paper">Paper</button>
+    <button type="submit" name="move" value="scissors">Scissors</button>
+  </form>
+</div>
+<div>
+<form action="/">
+  <button type="submit" name="move" value="rock">Main Menu</button>
+</form>
```
RaeRachael, line 59: this has confusing labelling. is name = "move" a good name? and value = "rock" ?

views/play_single.erb
```diff
@@ -0,0 +1,52 @@
...
+      <button type="submit" name="move" value="paper">Paper</button>
+      <button type="submit" name="move" value="scissors">Scissors</button>
+    </form>
+  </div>
+  <div>
+  <form action="/">
+    <button type="submit" name="move" value="rock">Main Menu</button>
+  </form>
+  </div>
+</body>
```
RaeRachael, line 52: looks very similar to play_muli.erb maybe the same erb could have been used for both?

views/result_multi.erb
```diff
@@ -0,0 +1,66 @@
...
+  <h1><%= @winner.name.upcase %> WON</h1> 
+  <h2>&#129395;</h2>
+  <h3><%= @winner.name.upcase %> PICKED <%= @winner.move.to_s.upcase %></h3>
+  <h3><%= @loser.name.upcase %> PICKED <%= @loser.move.to_s.upcase %><br></h3>
+  <div>
+    <form action="/play-multi">
+      <input type="submit" value="Play Again">
+    </form>
+  <div>
+<% end %>
```
RaeRachael, line 65: would having the logic here in the Model part be better, maybe in the Winner Class?

app.rb
```diff
@@ -0,0 +1,71 @@
...
+    erb @game.result(@player)
+  end
+
+  get '/result-multi' do
+    @player_1 = @game.player_1 # in the case of a draw
+    @player_2 = @game.player_2 # in the case of a draw
+    @winner = @game.winner
+    @loser = @game.loser
+    erb :result_multi
+  end
```
RaeRachael, line 68: I just wonder if both multiplayer and singleplayer could have utilised some of the same paths? (maybe it would be more confusing here, but could have fewer erb files to manage?)

lib/game.rb
```diff
@@ -0,0 +1,66 @@
...
+    playing_as == player_2 ? '/play-multi' : '/result-multi'
+  end
+
+  def result(player)
+    case @winner.decide(player_1, player_2)
+      when player then :win
+      when :draw then :draw
+      else :lose
+    end
+  end
```
RaeRachael, line 65: Curious if some of these methods could be made private?

spec/computer_spec.rb
RaeRachael: Game::MOVES[1] ?

app.rb
```diff
@@ -0,0 +1,71 @@
...
+    Game.create(player_1, player_2)
+    redirect '/play-single'
+  end
+
+  post '/add-multi' do
+    player_1 = Player.new(params[:player_1_name])
+    player_2 = Player.new(params[:player_2_name])
+    Game.create(player_1, player_2)
+    redirect '/play-multi'
+  end
```
tatsiana-makers, line 31: I expect that we don't need both single and multiplayer paths in the logic with some refactoring of what you already have.

Imagine from the home page, you could have a single "Play button" which leads to a page which looks like the current `start?mode=multi`  one. Then if the user wants to play single player against a computer, they just leave the `Player 2` field in the form blank. They click Submit. 

This leads us to a route where we use a ternary operator to determine what `player_2` should be:
`player_2 = params[:player_2_name].empty? ? "t" :  "f"`
where instead of `"t"` and `"f"` we correctly assign the variable to either a Player or a Computer

There might be some more changes to the Game class that you would need to do, but essentially we can make use of [Duck Typing](https://en.wikipedia.org/wiki/Duck_typing) to have the Game class not care if the second player is an instance of Player or Computer

spec/features/result_single_spec.rb
```diff
@@ -0,0 +1,30 @@
+feature 'result page single player' do
+  scenario 'displays players move' do
+    sign_in_and_pick_rock
+    expect(page).to have_content "YOU PICKED ROCK"
+  end
+
+  scenario 'states if player has won' do
+    srand(4)
```
tatsiana-makers, line 8: 🙌 Nice! I think this use of `srand` allows you to thoroughly feature test a single player game 🙂 

Another potential alternative is to mock the `sample` method. An example can be found here: https://github.com/makersacademy/rps-challenge/blob/master/docs/review.md#stub-out-random-behaviour

PR #1667, opened 2020-06-01
Completed but lack more tests..

victorvallet:


spec/features/play_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  before do
+    visit('/')
+    fill_in :name, with: 'Victor'
+    click_button 'Play'
+  end
+  scenario "the player can choose paper" do
+    expect(page.find("input#paper")).not_to be nil
+  end
+
+  scenario "the player can choose rock" do
```
sophiewo, line 11: you could merge all of these expects into one scenario (like we discussed)

views/results.erb
```diff
@@ -0,0 +1,29 @@
+<div class="game">
+<p>You chose <%= @move %></p>
+<p>Your opponent's move: <%= @opponent_move %></p>
+<div>
+<h3>Results:
```
sophiewo, line 5: you could move this into a game.rb file to remove logic from the views.

PR #1676, opened 2020-06-28
RPS Weekend Challenge - Dhara

Dhara-95:
- Both user stories have been completed (asks player to register name and allows player to play Rock, Paper, Scissors game against the computer).

- As mentioned in my README_NEW.md, I would like to refactor my code further to enable sessions and make the results method in the rps.rb file more concise. I would also like to use CSS for formatting my web app and potentially add functionality that allows two players to play the game rather than just one player vs the computer.


spec/computer_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect(subject).to respond_to(:computer_move)
+  end
+
+  it 'randomly selects the move scissors' do
+      srand(4)
+      expect(subject.computer_move).to eq('scissors')
+    end
+
+    it 'randomly selects the move paper' do
+      rand(3)
```
alicelieutier, line 14: This should be `srand`, right? You should be stubbing randomness in each test that relies on it, otherwise the success of your test will depend on the order in which the tests are run.

spec/computer_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+      expect(subject.computer_move).to eq('paper')
+    end
+
+    it 'randomly selects the move rock' do
+      rand(2)
```
alicelieutier, line 19: same.

spec/features/game_spec.rb
```diff
@@ -0,0 +1,12 @@
+feature 'Entering a move' do
+  scenario 'Player enters a move: rock, paper or scissors' do
+    visit '/'
```
alicelieutier, line 3: You should also be using `srand` in this test.

rps_web.rb
```diff
@@ -0,0 +1,23 @@
...
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    @player_name = params[:player_name]
+    erb :game
+  end
+
+  post '/game' do
```
alicelieutier, line 16: You asked whether this route has too many lines. It doesn't, really. However, if you wanted, you could separate logic even more. For example, you could have a game class, that takes the player move, is responsible for picking a computer move, and is responsible for formatting an answer that gives the result. It could delegate everything to do with the player to another class, and the formatting to a third one. This way, that "game" class would act as the interface of your programme, and then this controller only interact with that interface.
You can think of this as if you had an "irb app", and then this controller uses that irb app, the way you would use it as a player directly in IRB.

Let me know if that isn't clear.

rps_web.rb
```diff
@@ -0,0 +1,23 @@
...
+    @player_name = params[:player_name]
```
alicelieutier, line 17: Why is the player name still part of the params here 🤔

alicelieutier: It's not sent by the form to this route.

rps_web.rb
```diff
@@ -0,0 +1,23 @@
+require 'sinatra/base'
+require './lib/rps.rb'
+require './lib/computer.rb'
+
+class RPSWeb < Sinatra::Base
```
alicelieutier, line 5: Both your post routes don't use the 'post, redirect, get' pattern. It's not a must, but it's good practice.

lib/rps.rb
```diff
@@ -0,0 +1,16 @@
+class RPS
+
+  def result(player_move, computer_move)
```
alicelieutier, line 3: This code works for me. There are more compact ways to write it, for sure, but at the same time, this is really easy to read and understand, so I don't mind it.

spec/features/game_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+    fill_in :player_name, with: 'Dhara'
+    click_button 'Submit'
+    fill_in :player_move, with: 'rock'
+    click_button 'Submit'
+    expect(page).to have_content "Your move is rock"
+    expect(page).to have_content "Computer move is rock"
+    expect(page).to have_content "It's a draw!"
+  end
```
alicelieutier, line 11: It would be good to test every possible outcome.

spec/computer_spec.rb
Dhara-95: I initially thought from reading this resource https://gist.github.com/Kotauror/8f4a9913c53dd479df91909f306e6731 that you start with srand then only need rand for the other two tests. However after testing this out I now see that you don't and have amended this.

spec/computer_spec.rb
Dhara-95: Same as above.

spec/features/game_spec.rb
Dhara-95: Amended to use srand as suggested.

rps_web.rb
Dhara-95: Yes thanks for clarifying, that makes sense :)

rps_web.rb
Dhara-95: Good point, not sure why I thought I needed it there - line 17 has been removed now.

rps_web.rb
Dhara-95: Yes I did think about this, the PRG pattern is something that I'm still unsure on how to use correctly and it's purpose, so will look into it further and try to amend my post routes accordingly for best practice.

lib/rps.rb
Dhara-95: Ok thanks. I also agree that although my code is clear, there are probably ways to condense this into less lines of code, so will look into this further.

spec/features/game_spec.rb
Dhara-95: Noted, thanks :)

PR #1685, opened 2020-07-26
Ralph Mallett - RPS Challenge

ralphm10:
- The user stores are satisfied in this challenge
- The code does include global variables which are not best practice so the next step would be to refactor to remove these
- The result method inside the game class is too long with the use of conditionals and also needs to be refactored
- I would like to add some more CSS styling to the views by using a style sheet


views/play.erb
```diff
@@ -0,0 +1,24 @@
...
+
+}
+</style>
+</head>
+
+<h2>Welcome <%= $player.name %>!</h2>
+
+<h3>Make your selection:</h3>
+
+<form action="/result" method="post">
```
rjkviegas, line 16: really like the use of option here

lib/game.rb
```diff
@@ -0,0 +1,17 @@
+class Game
+
+  def result
+    if $option == $opponent
```
rjkviegas, line 4: I had something similar and rubocop suggested using a guard condition so: return "It's a draw" if ...

lib/opponent.rb
```diff
@@ -0,0 +1,9 @@
+class Opponent
```
rjkviegas, line 1: very concise class, makes it obvious what behaviour it exhibits

views/index.erb
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+<head>
+<style>
+body {
+  background-color: paleturquoise;
```
rjkviegas, line 6: I like the addition of a background colour for the views

app.rb
```diff
@@ -0,0 +1,29 @@
...
+class RPS < Sinatra::Base
+  get '/' do
+    erb(:index)
+  end
+
+  get '/test' do
+    'Testing RPS!'
+  end
+
+  post '/play' do
```
rjkviegas, line 15: Perhaps you could HAVE utilised a POST-redirect loop, so a GET request is used to render the view opposed to a POST. This is something I saw detailed in the Battle walkthrough.

lib/game.rb
```diff
@@ -0,0 +1,17 @@
...
+      "It's a draw..."
+    elsif $option == "Rock" && $opponent == "Scissors"
+      "You win!"
+    elsif $option == "Paper" && $opponent == "Rock"
+      "You win!"
+    elsif $option == "Scissors" && $opponent == "Paper"
+      "You win!"
+    else
```
rjkviegas, line 12: Great use of else condition to capture when the user loses

spec/player_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'player'
+
+describe Player do
+
+  subject(:player) { Player.new('Ralph') }
```
rjkviegas, line 5: nice refactoring of the subject object

PR #1687, opened 2020-07-26
rock paper scissors in progress, couldnt fix bug

melanieanderson1995:


app.rb
```diff
@@ -3,57 +3,48 @@
...
-    @game = Game.create(player_1, player_2)
+    $game = Game.new(player_1, player_2)
     erb :play_singleplayer
   end
 
   get "/rock_computer" do
-    @player_1_name = $player_1.name
-    @computer_name = $computer.name
-    @game = Game.new
     @computer_move = @game.random_selection
```
Pi-hils, line 38: I like the change of names here (computer_name -->computer_move) as it makes it easier to understand.

spec/game_spec.rb
```diff
@@ -17,8 +19,8 @@
 
   describe "#computer_move" do
     it "returns random selection" do
-      allow(mock_computer).to receive(:random_selection).and_return("rock")
-      expect(mock_computer.random_selection).to eq("rock")
+      allow(player_2).to receive(:random_selection).and_return("rock")
```
Pi-hils, line 22: nice use of the edge case here. I like how you separated each player and did a separate test on both.

lib/game.rb
```diff
@@ -8,14 +8,6 @@ def initialize(player_1, player_2)
...
-  def self.instance
-    @game
-  end
-
-  def self.create(player_1, player_2)
-    @game = Game.new(player_1, player_2)
-  end
-
   def random_selection
     @choices.sample
```
Pi-hils, line 12: Great use of DI here and your code is DRY so well done for that.

PR #1691, opened 2020-07-26
done

Gonciarov:


lib/views/index.erb
```diff
@@ -0,0 +1,11 @@
+
```
phump81, line 1: You seem to have two copies of your 'views' folder. One in your main folder, and one in the lib folder. I believe the lib folder should only contain your models, so I think you can remove these copies.

lib/winner.rb
```diff
@@ -0,0 +1,15 @@
+
```
phump81, line 1: I like this method a lot. I used a really long and ugly case statement, but this is much better!

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def sign_in_and_play
```
phump81, line 1: Nice use of a web helper!

lib/app.rb
```diff
@@ -0,0 +1,35 @@
...
+require './lib/winner.rb'
+
+class Rps < Sinatra::Base
+  enable :sessions
+
+get '/' do
+  erb :index
+end
+
+post '/names' do
```
phump81, line 11: Nice use of sessions. I didn't think to use them. Controller looks skinny too!

lib/views/play.erb
```diff
@@ -0,0 +1,12 @@
+<%=@player_1_name%> vs. Computer
+
+<form action="/choices" method="post">
+
+  <select name="choice">
```
phump81, line 5: I like the drop down menu for the options. A good choice!

config.ru
```diff
@@ -0,0 +1,2 @@
+require_relative "./lib/app"
```
phump81, line 1: All tests pass, with 100% coverage! Nice work!

PR #1706, opened 2020-08-24
RPS challenge

ZishJawaid:


app.rb
```diff
@@ -0,0 +1,38 @@
...
+    erb :index
+  end
+
+  post '/name' do
+    session[:name] = params[:name]
+    redirect('/play')
+  end
+
+  get '/play' do
+    @name = session[:name]
```
fraserbrookhouse, line 19: You could have kept the controller more skinny by extracting a Player class and assigning the name to it, then you could have stored them in the Game class.

lib/game.rb
```diff
@@ -0,0 +1,19 @@
+class Game
+
+  WINNING_COMBINATIONS = {
```
fraserbrookhouse, line 3: I like this method of storing the winning/losing move combinations, much cleaner and easier to change than a large conditional tree (like I did).

views/play.erb
```diff
@@ -0,0 +1,20 @@
...
+<html>
+<head>
+<link rel="stylesheet" type="text/css" href="style.css" />
+</head>
+
+<h1>Let's play, <%= @name %>!</h1>
+
+<p>Choose your move</p>
+
+<form class="container" action="/turn" method="post">
```
fraserbrookhouse, line 11: I think a drop down menu is not the most intuitive UI decision for a game, there aren't a million options. So I think buttons are a better, more accessible choice.

public/style.css
```diff
@@ -0,0 +1,71 @@
+html, body {
+    background-color: rgb(68, 95, 129);
```
fraserbrookhouse, line 2: I like your colour scheme.

PR #1709, opened 2020-09-16
ROCK PAPER SCISSORS LIZARD SPOCK

vikjusko:


.history/lib/game_20200921092231.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require './lib/player.rb'
+
+class Game
+  attr_accessor :player, :comp
+
+  WIN = {
```
JYip93, line 8: I like this way of storing your choices and then later choosing the winner. I would have never thought of it like that. Really interesting.

.history/spec/game_spec_20200921081942.rb
```diff
@@ -0,0 +1,35 @@
+require 'game'
+require 'player'
+
+
+describe Game do
```
JYip93, line 5: If there were more time, you can add more test cases to get a higher percentage in test coverage

PR #1710, opened 2020-09-18
Catzkorn RPS 

catzkorn:


lib/bot.rb
```diff
@@ -0,0 +1,13 @@
+class Bot
+  attr_reader :name, :move
+
+  def initialize
+    @name = "Bot"
+    @moves = ["Rock", "Paper", "Scissors"]
+    @move = nil
```
Katerina-codes, line 7: Generally, I avoid setting a value as nil. If nil gets thrown as an error/or response somewhere else you'll know it's not because of anything you have done and that there is an issue in the code. If you want to represent nil you could do so with **""**, **0** or any other datatype that is consistent with others you are using in that particular domain. 
Let me know if that makes sense or if you have questions? :)

lib/player.rb
```diff
@@ -0,0 +1,14 @@
+require_relative "game"
+
+class Player
```
Katerina-codes, line 3: Is Bot not also a type of player?
I would name this class Human to make the naming here more explicit. 

Once you have your Bot and Human classes you'll make it very clear that both are types of players and that they have the same behaviour in which case perhaps you could reduce some duplication that occurs in Game with Polymorphism. Do you know how you might do this?

XavierDefontaine:
I think this needs reopening for posterity 😆

catzkorn:
I feel I've let myself down not continuing the excessive ASCII art in all my readmes :(

PR #1711, opened 2020-09-18
Paper disproves Spock

TheDancingClown:
Initial commit


TheDancingClown:
Useful functionality to include - a score reset, an end game (which resets score and redirects to login), and multiplayer.

PR #1714, opened 2020-09-20
RPS Challenge - Johnny Yip

JYip93:


app.rb
```diff
@@ -0,0 +1,54 @@
...
+        @choice = params[:choice]
+        redirect '/result'
+        
+    end
+
+    get '/result' do
+        @game = Game.instance
+        @player = session[:name]
+         @result = @game.get_result     
+        if @result == "Draw"
```
vikjusko, line 40: Really great conditional move that redirects you to different pages depending on the result! will definitely use that as well!

lib/choices.rb
```diff
@@ -0,0 +1,13 @@
+class Choices
```
vikjusko, line 1: You could create a choices as a constant and run it within your game class, instead of creating a separate class for it

lib/game.rb
```diff
@@ -0,0 +1,81 @@
...
+     #   @cpu_choice = @choices.get_choices.sample
+    #end
+    
+    private 
+
+    def get_cpu_choice
+        @choices.get_choices.sample
+    end
+    
+    def decide_winner
```
vikjusko, line 60: Can also have a WINNING_HAND hash that is a constant to make the conditional statement shorter

public/index.css
```diff
@@ -0,0 +1,52 @@
+
+body {
```
vikjusko, line 2: Enjoy how many images and gifs you added to different pages of your game!

spec/feature/register_name_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'Register name for game' do
```
vikjusko, line 1: You could create a sign-in method to make your feature tests quicker

PR #1750, opened 2020-11-15
Rock, Paper, Scissors, Lizard, Spock!

horthbynorthwest:
So I managed to get the 2nd bonus feature but not the multiplayer one due to the way I'd coded it. No global variables needed though

I'm also aware that my  p1_wins method is also a bit long, and suggestions on how I could improve that is welcome.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,13 @@
+require 'sinatra/base'
+
+class Game < Sinatra::Base
+
+  set :session_secret, 'rps'
+
+  get '/' do
+    erb :index
+  end
+
+  run! if app_file == $0
+
+end
--- a/app.rb
+++ b/app.rb
@@ -1,22 +1,25 @@
 require 'sinatra/base'
 require './lib/player'
+require './lib/game'
 
-class Rps < Sinatra::Base
+class Clicker < Sinatra::Base
 
   set :session_secret, 'rps'
+
+  before do
+    @player = Player.instance
+  end
 
   get '/' do
     erb :index
   end
 
   post '/names' do
-    $player_1 = Player.new(params[:player_1_name])
+    @player = Player.create(params[:player_1_name])
     redirect '/play'
   end
 
   get '/play' do
-    @player_1 = $player_1
-
     erb :play
   end
 
@@ -60,15 +63,15 @@ class Rps < Sinatra::Base
   end
 
   get '/winner' do
-    "Matt wins!"
+    erb :winner
   end
 
   get '/loser' do
-    "Matt lost to an AI!"
+    erb :loser
   end
 
   get '/draw' do
-    "Matt & AI drew!"
+    erb :draw
   end
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -21,15 +21,54 @@ class Rps < Sinatra::Base
   end
 
   post '/rock' do
-
+    p1 = 'rock'
+    p2 = ['rock', 'paper', 'scissors'].sample
+    @game = Game.new(p1, p2)
+    if @game.result == 'rock'
+      redirect '/winner'
+    elsif @game.result == 'draw'
+      redirect '/draw'
+    else
+      redirect '/loser'
+    end
   end
 
   post '/paper' do
-
+    p1 = 'paper'
+    p2 = ['rock', 'paper', 'scissors'].sample
+    @game = Game.new(p1, p2)
+    if @game.result == 'paper'
+      redirect '/winner'
+    elsif @game.result == 'draw'
+      redirect '/draw'
+    else
+      redirect '/loser'
+    end
   end
 
-  post 'scissors' do
+  post '/scissors' do
+    p1 = 'scissors'
+    p2 = ['rock', 'paper', 'scissors'].sample
+    @game = Game.new(p1, p2)
+    if @game.result == 'scissors'
+      redirect '/winner'
+    elsif @game.result == 'draw'
+      redirect '/draw'
+    else
+      redirect '/loser'
+    end
+  end
 
+  get '/winner' do
+    "Matt wins!"
+  end
+
+  get '/loser' do
+    "Matt lost to an AI!"
+  end
+
+  get '/draw' do
+    "Matt & AI drew!"
   end
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -1,7 +1,7 @@
 require 'sinatra/base'
 require './lib/player'
 
-class Game < Sinatra::Base
+class Rps < Sinatra::Base
 
   set :session_secret, 'rps'
 
@@ -29,7 +29,7 @@ class Game < Sinatra::Base
   end
 
   post 'scissors' do
-    
+
   end
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -1,4 +1,5 @@
 require 'sinatra/base'
+require './lib/player'
 
 class Game < Sinatra::Base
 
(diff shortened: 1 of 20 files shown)


app.rb
```diff
@@ -0,0 +1,13 @@
+require 'sinatra/base'
+
+class Game < Sinatra::Base
+
+  set :session_secret, 'rps'
```
joshuaretallick, line 7: I know you had a chat with Tatsiana the other week, how does this secret session actually work? And how do you call the password 'rps'? Do you type in shotgun 'rps' or something from IRB or am I generally misunderstanding what secret session even does?

app.rb
```diff
@@ -8,6 +9,17 @@ class Game < Sinatra::Base
     erb :index
   end
 
+  post '/names' do
+    $player_1 = Player.new(params[:player_1_name])
```
joshuaretallick: Ok see you are using Global Variables now.  I had these running on mine and then reverted back to sessions.  They keep saying global variables are bad, are sessions? What is the advantage of one over the other (I dont actually know, this isnt me challenging you! haha)

views/play.erb
```diff
@@ -1 +1,11 @@
...
+
+<form action="/rock" method="post">
+  <input type="submit" value="Rock">
+</form>
+<form action="/paper" method="post">
+  <input type="submit" value="Paper">
+</form>
+<form action="/scissors" method="post">
+  <input type="submit" value="Scissors">
+</form>
```
joshuaretallick, line 11: I defo should have done this, for the life of me I couldnt figure this out and just opted for typing the input and sending everyone to /player_attack rather than individual rock, paper, scissors pages.  Reason being is my testing kept failing when I would send them to one page with buttons (expected Rock, got Scissors etc)

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def scissors_win
+    return 'true' if @p1 == 'scissors' && @p2 == 'paper'
+  end
+
+  def paper_win
+    return 'true' if @p1 == 'paper' && @p2 == 'rock'
+  end
+end
```
joshuaretallick, line 58: I think you mentioned this in your opening comments but perhaps you could tidy this all up in to one method
eg return "You win!" if @p1 == "Rock" && @p2 == "Scissors"

spec/unit/game_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+  end
+
+  describe '#a losing game' do
+    it 'allows the user to lose' do
+      lose_game = Game.new('rock', 'paper')
+      expect(lose_game.result).to eq 'lose'
+    end
+  end
+end
```
joshuaretallick, line 51: This is great testing, out of curiosity would we be expected to test all outcomes here (allows paper to win, allows scissors to win etc) or could we get away with just stubbing an outcome and assuming all tests are secure (again Im not questioning, I dont actually know, I would have opted for your method as well)

app.rb
```diff
@@ -21,15 +21,54 @@ class Rps < Sinatra::Base
...
+    p2 = ['rock', 'paper', 'scissors'].sample
+    @game = Game.new(p1, p2)
+    if @game.result == 'rock'
+      redirect '/winner'
+    elsif @game.result == 'draw'
+      redirect '/draw'
+    else
+      redirect '/loser'
+    end
   end
```
joshuaretallick, line 37: Would we typically run the if, elsif, else from app.rb? I guess this is necessary as you have different redirects based on outcome?

spec/features/paper_spec.rb
```diff
@@ -3,4 +3,25 @@
     sign_in_and_play
     expect(page).to have_button 'Paper'
   end
+
+  scenario 'Player chooses paper and wins' do
+    srand(67810)
```
joshuaretallick, line 8: You will see I used srand, tbh I have no idea.
I couldnt find a good article about the ins and outs, some refered to adding additional code to spec_helper too

app.rb
```diff
@@ -1,22 +1,25 @@
...
+    @player = Player.instance
+  end
+
   get '/' do
     erb :index
   end
 
   post '/names' do
-    $player_1 = Player.new(params[:player_1_name])
+    @player = Player.create(params[:player_1_name])
```
joshuaretallick, line 18: Very cool, I only just started on dumping global variables last week so wasnt confident enough to try this

spec/spec_helper.rb
```diff
@@ -10,7 +10,7 @@
 ])
 SimpleCov.start
 require File.join(File.dirname(__FILE__), '..', 'app.rb')
-Capybara.app = Rps
+Capybara.app = Clicker
```
joshuaretallick, line 13: Can you run me through the change here and why?

app.rb
```diff
@@ -25,7 +25,7 @@ class Clicker < Sinatra::Base
 
   post '/rock' do
     p1 = 'rock'
-    p2 = ['rock', 'paper', 'scissors'].sample
+    p2 = ['rock', 'paper', 'scissors', 'spock', 'lizard'].sample
```
joshuaretallick, line 28: Damn I should have done this, I guess it wasnt that much extra changing required?

lib/game.rb
joshuaretallick: Haha I see you changed this later after all

joshuaretallick (commented):
Ace code Georgie, super easy to read and understand

app.rb
```diff
@@ -0,0 +1,105 @@
...
+    p1 = 'rock'
+    p2 = ['rock', 'paper', 'scissors', 'spock', 'lizard'].sample
+    @game = Game.new(p1, p2)
+    if @game.result == 'rock'
+      redirect '/winner'
+    elsif @game.result == 'draw'
+      redirect '/draw'
+    else
+      redirect '/loser'
+    end
```
tatsiana-makers, line 36: Whenever we have an `if/else` statement, ideally we want the things we use in the conditions to be the same "type of thing".

In this case, we are either comparing `@game.result` to "rock" or to "draw", which aren't the same type of thing because "rock" is a move", while "draw" is the result of comparing two moves. 

This suggests that `@game.result` is occasionally returning the result of the comparison and occasionally returning which move the player made. Can we separate these concerns out into two different methods  - in this specific case, can we make use of an attribute reader to read out which move was made and have `.result` only be responsible for determining the outcome of the move?

app.rb
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+  post '/paper' do
```
tatsiana-makers, line 39: A lot of the code across the 5 post routes `post /rock`, `post /paper`, etc. ends up repeated, which means the code isn't following the DRY principle. Rather than having 5 routes which all essentially do the same thing, but for the different moves, can you see a way of having 1 route which can handle all the potential moves?

lib/game.rb
```diff
@@ -0,0 +1,58 @@
...
+    if rock_win
+      'rock'
+    elsif scissors_win
+      'scissors'
+    elsif paper_win
+      'paper'
+    elsif spock_win
+      'spock'
+    elsif lizard_win
+      'lizard'
```
tatsiana-makers, line 28: In all of these possible win scenarios, we're returning the move that player1 made. Rather than returning a new string, where is this player's move already stored?

app.rb
```diff
@@ -0,0 +1,105 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post '/rock' do
+    p1 = 'rock'
+    p2 = ['rock', 'paper', 'scissors', 'spock', 'lizard'].sample
```
tatsiana-makers, line 28: We could extract out this sampling logic into a model, for example a `Computer` model

tatsiana-makers (commented):
This is a nice go at the RPS challenge! Great work adding in the lizard and spock move options! I've left a few comments. Happy to chat about any of these further 🙂

PR #1751, opened 2020-11-15
RPS Weekend Challenge Submission

margmartelo:
User stories covered.
Bonus level 2 included. 
100% coverage


app.rb
```diff
@@ -0,0 +1,34 @@
...
+    @player_name = session[:player_name]
+    erb :play
+  end
+
+  get '/game' do
+    @player_name = session[:player_name]
+    @player_choice = params[:option]
+
+
+    @computer_choice = ['rock', 'paper', 'scissors', 'spock', 'lizard'].sample
```
alexleesonmill, line 26: Could you have had this in the game class for better single responsibility?

spec/features/play_spec.rb
```diff
@@ -0,0 +1,16 @@
+feature "playing" do
+  scenario "presented with the different choices" do
+    register_and_play
+    expect(page).to have_link(nil, href: "/game?option=rock")
```
alexleesonmill, line 4: Haven't seen have_link before - this is really good!

spec/game_spec.rb
```diff
@@ -0,0 +1,44 @@
+require_relative '../lib/game'
+
```
alexleesonmill, line 2: Good for full coverage but it feels like all these tests are testing broadly the same thing? i.e if one test passes they all should as the logic is sound?

views/game.erb
```diff
@@ -0,0 +1,10 @@
+
+<div>
+<%= @player_name %> chose <%= @player_choice %>.
```
alexleesonmill, line 3: Good use of global variables to pull through player choices!

views/play.erb
```diff
@@ -0,0 +1,10 @@
+<div>
+<b><%= @player_name %></b> vs <b>Computer</b>
+</div>
+<div>
+<a href="/game?option=rock"><img src="/rock.png"/></a>
+<a href="/game?option=paper"><img src="/paper.png"/></a>
```
alexleesonmill, line 6: Also like the use of the link controlling the players move, I didn't think of this as an option so interesting to see it work in practice!

alexleesonmill (commented):
This is really good! 

Interested to hear your thoughts as to how you might implement a multiplayer option and a player class to have the controller be more 'skinny'!

PR #1753, opened 2020-11-15
Michael Schats - RPS Challenge

Michaelschats:


app.rb
```diff
@@ -0,0 +1,38 @@
+require 'sinatra/base'
```
malachispencer, line 1: Minimal logic in the controller, nice!

app.rb
```diff
@@ -0,0 +1,38 @@
...
+require './lib/game.rb'
+
+class RPS < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
```
malachispencer, line 11: How does using the session work? What are the advantages compared to using a class variable?

spec/features/attack_spec.rb
```diff
@@ -0,0 +1,28 @@
+feature 'Attacking' do
```
malachispencer, line 1: Good that you did feature tests for all the different scenarios and set strong expectations.

spec/game_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'game'
```
malachispencer, line 1: For your tests your could use another describe above your it blocks, so you can easily know what method the set of tests are referring to. For example `describe '#move' do`

spec/game_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+describe Game do
+  subject(:dave) { Game.new("Dave", "Rock") }
```
malachispencer, line 4: 2 different Game objects for one Game between 2 players?

PR #1755, opened 2020-11-16
RPS Challenge 

rafael-oelmann:
Basic RPS game that works (single player), I want to continue working on this.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -7,16 +7,22 @@ class RPS < Sinatra::Base
     erb :index
   end
 
-  post '/game' do
-    session[:player_1_name] = params[:player_1_name]
-    redirect '/game'
+  post '/name' do 
+    session[:name] = params[:name]
+    redirect '/play'
   end
 
-  get '/game' do 
-    @player_1_name = session[:player_1_name]
-    erb :game
+  get '/play' do 
+    @name = session[:name]
+    @choice = session[:choice]
+    erb :play
   end
 
+  post '/play' do 
+    session[:choice] = params[:choice]
+    redirect '/play'
+  end
+   
   # start the server if ruby file executed directly
   run! if app_file == $0
 end
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,16 @@
+require 'sinatra/base'
+
+class RPS < Sinatra::Base
+
+  get '/' do 
+    erb :index
+  end
+
+  post '/game' do
+    @player_1_name = params[:player_1_name]
+    erb :game
+  end
+
+  # start the server if ruby file executed directly
+  run! if app_file == $0
+end
--- a/app.rb
+++ b/app.rb
@@ -15,11 +15,13 @@ class RPS < Sinatra::Base
   get '/play' do 
     @name = session[:name]
     @choice = session[:choice]
+    @cpu_choice = session[:cpu_choice]
     erb :play
   end
 
   post '/play' do 
     session[:choice] = params[:choice]
+    session[:cpu_choice] = :Rock
     redirect '/play'
   end
    
--- a/app.rb
+++ b/app.rb
@@ -1,13 +1,19 @@
 require 'sinatra/base'
 
 class RPS < Sinatra::Base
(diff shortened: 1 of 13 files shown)


app.rb
```diff
@@ -0,0 +1,34 @@
...
+    @cpu_choice = session[:cpu_choice]
+    @result = session[:result]
+    erb :play
+  end
+
+  post '/play' do 
+    round = Rps.new
+    session[:choice] = params[:choice]
+    session[:cpu_choice] = round.cpu_choice
+    session[:result] = round.compare(session[:choice], session[:cpu_choice])
```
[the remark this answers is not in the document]
tatsiana-makers: My understanding for why we need to put `round.cpu_choice` in the session is so we can have access to it in the `get /play` route. 

Another way we could achieve this is through having this stored inside of `round` as an instance variable and then having `round` be a global variable so it's accessible across all routes. 

This isn't great practice though, I encourage you to have a read through https://github.com/makersacademy/course/blob/master/intro_to_the_web/killing_the_global_variable.md and the corresponding walk-through and have a go at making these changes:
- first, can you get it working with a global variable `$round`
- then, have a go at replacing that global variable with the Singleton patterm=n

lib/rps.rb
```diff
@@ -0,0 +1,28 @@
...
+    elsif(user_choice == "Paper" && cpu_move == "Scissors") 
+      return "Computer wins :(" 
+    elsif(user_choice == "Paper" && cpu_move == "Rock") 
+      return "You win" 
+    elsif(user_choice == "Scissors" && cpu_move == "Rock") 
+      return "Computer wins :(" 
+    elsif(user_choice == "Scissors" && cpu_move == "Paper") 
+      return "You win" 
+    else 
+      return "It's a tie!"
```
tatsiana-makers, line 24: This is quite messy to read. Are there any ways we can make use of data structures available in Ruby which would allow us to condense this if/elsif/else?

For example, can you think of a way we might have a hash which represents the winning pairs of moves and how we could use that:
```
{
 "Rock" => "Paper",
  ...
}
```

This isn't the only way to condense this code, but might get you thinking 🙂

spec/rps_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  
+  it 'provides CPU with a random choice' do
+    srand(23332)
+    expect(subject.cpu_choice).to eq "Paper"
+  end
+
+  it 'compares user choice and cpu choice, returns winner' do 
+    srand(23332)
+    this_round = Rps.new
+    expect(this_round.compare('Scissors')).to eq "You win"
```
tatsiana-makers, line 13: Why do we use `this_round` here instead of `subject`?

tatsiana-makers: This test is currently failing - can you use the error message to fix it?

views/index.erb
```diff
@@ -0,0 +1,30 @@
...
+          _______) P A P E R 
+         _______)
+---.__________)
+
+    _______
+---'   ____)____
+          ______)
+       __________) S C I S S O R S
+      (____)
+---.__(___)
```
tatsiana-makers, line 28: This is very cool!

app.rb
```diff
@@ -0,0 +1,34 @@
...
+class RPS < Sinatra::Base
+  enable :sessions
+
+  get '/' do 
+    erb :index
+  end
+
+  post '/name' do 
+    session[:name] = params[:name]
+    redirect '/play'
```
tatsiana-makers, line 13: I found it slightly confusing being shown the "You chose" message before I had actually chosen anything. Perhaps This could show us a page with only the move options? 

And then in the `play.erb` if we wanted to also show these options you could make use of Sinatra partials: http://sinatrarb.com/faq.html#partials

tatsiana-makers (commented):
Thanks for asking for feedback! I've left a few comments with links and suggestions of what to explore next. I encourage you to spend some time trying to refactor after reading those links and then reaching out if you have any questions or get blocked 🙂

PR #1760, opened 2020-11-16
RPS Challenge

samanthagottlieb:


app.rb
```diff
@@ -0,0 +1,44 @@
...
+  enable :sessions
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/names' do
+    session[:player_1] = Player.new(params[:player_1])
+    session[:computer] = Computer.new
+    session[:game] = Game.new(session[:player_1], session[:computer])
```
tatsiana-makers, line 16: Rather than storing the game instance in the session, you could get it working with the singleton pattern. The first step towards this is using a global variable for example `$game` to store the instance and use this global variable across the routes wherever you currently use `session[:game]`. Then I'd encourage you to read through [this step](https://github.com/makersacademy/course/blob/master/intro_to_the_web/killing_the_global_variable.md) and corresponding walkthrough to see how to to get rid of the global variable

app.rb
```diff
@@ -0,0 +1,44 @@
...
+class RPS < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/names' do
+    session[:player_1] = Player.new(params[:player_1])
+    session[:computer] = Computer.new
```
tatsiana-makers, line 15: Since you pass both of these into the `Game.new` instead of storing them in the session, can we create attribute readers and read them out of the Game instance whenever we need the values?

This will allow the Game class to encapsulate these values

app.rb
```diff
@@ -0,0 +1,44 @@
...
+    erb(:play)
+  end
+
+  post '/choose' do
+    @game = session[:game]
+    session[:p1_choice] = params[:choice]
+    @p1_choice = session[:p1_choice]
+    session[:p2_choice] = @game.player_2.choose
+    @p2_choice = session[:p2_choice]
+    session[:outcome] = @game.result(@p1_choice, @p2_choice)
```
tatsiana-makers, line 31: Rather than setting the value here and then reading it out in `get /outcome`, could we just call `@game.result` within the `get /outcome` route and use the outcome immediately?

lib/game.rb
```diff
@@ -0,0 +1,21 @@
+class Game
+  OPTIONS = ["Rock", "Paper", "Scissors"]
+  BEATS = { "Rock" => "Scissors", "Paper" => "Rock", "Scissors" => "Paper" }
```
tatsiana-makers, line 3: Nice!

lib/computer.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class Computer
+  attr_reader :name
+
+  def initialize(name = "The Computer")
+    @name = name
+  end
+
+  def choose
+    Game::OPTIONS.sample
```
tatsiana-makers, line 11: Here, the Computer class is implicitly relying on the Game class, rather than in the code injecting the dependency. 

Have a look at the Greeter/Smiley example in the [dependency injection practical](https://github.com/makersacademy/skills-workshops/blob/f5b4801840fe07d26ff70341652dc81dcda12289/practicals/object_oriented_design/dependency_injection.md#introduction). See how in the example code (once we refactor it), we have decoupled the classes by injecting the dependency - we can do something similar here, specifically:

In the initialize method, can we inject `Game::OPTIONS`:
```
def initialize(name = "The Computer", options = GAME::OPTIONS)
   @name = name
   @options = options
end
```

This means that in the Computer unit tests we can actually do something like
`computer = Computer.new("My computer", ["Option1", "Option2"]`
and therefore we have isolated the Computer class from the Game class.

spec/computer_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  describe '#name' do
+    it 'returns the name' do
+      expect(computer.name).to eq('The Computer')
+    end
+  end
+
+  describe '#choose' do
+    it 'returns a randomly chooses a shape' do
+      allow(computer).to receive(:choose).and_return(rock)
+      expect(computer.choose).to eq(rock) # testing anything meaningful?
```
tatsiana-makers, line 20: Instead of stubbing `choose`, what about stubbing `sample` or using `srand`. These other two options test that either `choose` uses `sample` under the hood or that there is some element of randomness.

spec/features/choose_spec.rb
```diff
@@ -0,0 +1,36 @@
+feature 'Choose' do
```
tatsiana-makers, line 1: In feature tests you don't want to be using stubs and mocks.

We want to isolate unit tests using stubs and mocks, but test complete features using the actual implementations of classes.

spec/game_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'game'
+
+describe Game do
+  subject(:game) { described_class.new(player_1, player_2) }
+  let(:player_1) { double :player }
+  let(:player_2) { double :player }
+  let(:rock) { double :rock }
+  let(:paper) { double :paper }
+  let(:scissors) { double :scissors }
```
tatsiana-makers, line 9: We create doubles to create fake versions of instances of classes or entire classes to allow us to isolate the unit tests of one class from another. 

Here making a double of players makes sense because you pass in instances of Player and Computer classes into the Game initialize method.

In our game rock is just being represented as a string "Rock" (not an instance of a class or a class itself), so you can just have 
`let(:rock { "Rock"}` if you want to, or skip this defining in a `let` block and just use `"Rock"` in the test blocks where you need to. Strings shouldn't be doubled because they're not instances of another class we've written that we need to isolate the Game class from.

tatsiana-makers (commented):
Thank you for asking for feedback! This is a nice go at the weekend challenge! I've left a few comments/explanations around testing and some suggestions for how you could change your code to rely less on sessions. I recommend having a read through and trying some things out. Reach out if you have further questions 🙂

PR #1762, opened 2020-11-16
stringiest RSP challenge

stringiest:


README.md
```diff
@@ -1,85 +1,64 @@
...
+$ git clone git@github.com:[USERNAME]/rps-challenge.git
+$ cd rps-challenge
+$ bundle
+$ rackup
+```
+
+## Sequence diagram
+
+![Sequence diagram for RPS challenge](Sequence_diagram.png)
+
```
CodeZeus-dev, line 23: Really informative sequence diagram.

app.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  post '/result' do
+    @player_name = $game.player
+    @player_choice = params[:choice]
+    @computer_choice = $game.computer_choice
+    $game.player_choice = @player_choice
+    @result = $game.result
+    erb :result
+  end
+
```
CodeZeus-dev, line 32: You could define an instance variable `@game` and set it equal to `$game` inside of your /result post request, and then access the player's name, player's choice and computer's choice from inside the HTML file by calling `@game.player` etc.

lib/game.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def rock_comparators
+    if @computer_choice == "Rock"
+      @status = "draw"
+    elsif @computer_choice == "Paper"
+      @status = "lose"
+    else @status = "win"
+    end
+  end
+
```
CodeZeus-dev, line 33: Good definition of specialised methods for each case.

lib/game.rb
```diff
@@ -0,0 +1,52 @@
+class Game
+  attr_accessor :player, :computer_choice, :player_choice
+
+  def initialize(player)
+    @player = player
+    @player_choice = nil
+    @computer_choice = ["Rock", "Paper", "Scissors"].sample
```
CodeZeus-dev, line 7: You could implement a Player class that takes care of your players so that in case you introduce a second player, it will be easier to keep your Game class clean and SRP-friendly. Otherwise, it might get quite crowded.

public/styles.css
```diff
@@ -0,0 +1,18 @@
+body {
+  background-color: darkseagreen;
+}
```
CodeZeus-dev, line 3: Assuming that all elements like h1, h2, p, etc are all part of the HTML's body, you could specify the font-family inside of the body styling.

spec/features/play_spec.rb
```diff
@@ -0,0 +1,17 @@
+feature 'Player can select rock, paper or scissors' do
+  scenario "Player can select 'rock'" do
+    sign_in_and_play
+    choose('Rock')
```
CodeZeus-dev, line 4: Clean and concise feature tests

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def sign_in_and_play
+  visit('/')
+  fill_in :player_name, with: 'Lucy'
```
CodeZeus-dev, line 3: Smart method definition so that feature tests are not repetitive.

views/index.erb
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <link type="text/css" rel="stylesheet" href="styles.css">
+</head>
+<body>
+
```
CodeZeus-dev, line 7: Try to divide the HTMLs body into smaller sections of code using either <section></section> or <div></div> so that it's more accessible and easy to navigate through.

views/index.erb
```diff
@@ -0,0 +1,25 @@
...
+<br>
+<br>
+<div>
+<form action='/names' method='post'>
+  <label for="player_name">
+    Player name
+    <input type="text" name="player_name">
+  </label>
+  <input type="submit" value="Submit">
+</form>
```
CodeZeus-dev, line 22: Good creation of an HTML form using a submit button.

views/play.erb
```diff
@@ -0,0 +1,24 @@
...
+<br>
+
+<form action='/result' method='post'>
+  <input type="radio" id='rock' name='choice' value="Rock">
+  <label for='rock'>Rock</label><br>
+  <input type="radio" id='paper' name='choice' value="Paper">
+  <label for='paper'>Paper</label><br>
+  <input type="radio" id='scissors' name='choice' value="Scissors">
+  <label for='scissors'>Scissors</label><br>
+  <br>
```
CodeZeus-dev, line 19: Good call in using radio buttons as all options are displayed to the user at once and allows the user to choose only one option.

lib/game.rb
```diff
@@ -0,0 +1,52 @@
...
+    if @player_choice == "Rock"
+      rock_comparators
+    elsif @player_choice == "Paper"
+      paper_comparators
+    else
+      scissors_comparators
+    end
+    @status
+  end
+
```
tatsiana-makers, line 22: Let's imagine we have a hash which looks like:
```
RULES = { "Rock": "Scissors",
         "Paper": "Rock",
         "Scissors": "Paper"}
```

What will we get if we do something like `RULES[@player_choice]`? We would get the move that the computer would have had to make so that the player won.

So we could compare `RULES[@player_choice]` to `@computer_choice` to decide if the player has won against the computer or not.

What are the three possibilities:
- player and computer made the same choice 
-  `RULES[@player_choice] == @computer_choice`
- `RULES[@player_choice] != @computer_choice`

What is the result of each of these three cases?

spec/game_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+describe Game do
+  let(:testgame) { Game.new("Test") }
+
+  it 'instantiates with a player' do
+    expect(testgame).to be_instance_of Game
+  end
+
+  it 'instantiates with a random computer_choice' do
+    srand(4)
```
tatsiana-makers, line 11: Nice use of `srand`!

spec/game_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    expect(testgame.computer_choice).to eq "Scissors"
+  end
+
+  describe "results" do
```
tatsiana-makers, line 15: We ideally want to minimise the amount of `attr_accessor`s our code has, as they open up more opportunity for bugs. 

We only using the `computer_choice` attribute accessor in the tests, which suggests we should be able to get rid of it in the code and make it a reader. Rather than setting the computer choice through the attribute accessor here, can we make use of `srand`? This allows us to make the code less open to errors 🙂

app.rb
```diff
@@ -0,0 +1,35 @@
...
+  attr_reader :player_choice
+
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    player = params[:player_name]
+    $game = Game.new(player)
```
tatsiana-makers, line 15: For future refactoring, there is an opportunity for us to not use global variables, which are dangerous. [This step and walkthrough](https://github.com/makersacademy/course/blob/master/intro_to_the_web/killing_the_global_variable.md) are something you may want to have a look at in the future for how we could get rid of the global `$game` variable

tatsiana-makers (commented):
I've left a hint as to how you might use the rules hash to condense your game logic as well as a few other comments! Let me know if you have any questions 🙂

lib/game.rb
stringiest: Thanks Tatsiana!  Have managed to get this working :)

spec/game_spec.rb
stringiest: Thanks!  Tests refactored using `srand`, and `computer_choice` set as an `attr_reader`.

PR #1764, opened 2020-11-16
RPS Challenge

kabligh:
Managed to complete the first two user stories. I had intended to include a "best of three" element but ran out of time so there may be a bit of legacy code relating to this.


lib/computer.rb
```diff
@@ -0,0 +1,9 @@
+class Computer
+
+  attr_reader :weapon
```
tatsiana-makers, line 3: Do we need the attr_reader here?

Instead of storing state as an instance variable in this class, could the `weapon` method below just return `["ROCK", "PAPER", "SCISSORS"].sample`?

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  attr_accessor :player, :computer
+
+  RULES = {
+    "ROCK" => "SCISSORS",
+    "PAPER" => "ROCK",
+    "SCISSORS" => "PAPER"
+  }
+
+  def initialize(player = Player.new, computer = Computer.new)
```
tatsiana-makers, line 14: Nice use of dependency injection!

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+    @game = Game.new(player, computer)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def results
+    if @player.weapon == @computer.weapon then "It's a draw!"
+    elsif RULES[@player.weapon] == @computer.weapon then "#{player.name} wins!"
```
tatsiana-makers, line 29: Every time you call `.weapon` on the `@computer` what happens?

Each time, the code in https://github.com/makersacademy/rps-challenge/pull/1764/files#diff-03595d1589bcc0550313ff8392390f633bde7f881185dbc7b038290fb0d3f25eR5-R7 will run, therefore giving a different result each time.

Instead, we probably want to call `@computer.weapon` once and store the result in a variable. Then we could use this variable wherever we need to, without regenerating the computer's move

lib/game.rb
```diff
@@ -0,0 +1,35 @@
...
+
+class Game
+
+  attr_accessor :player, :computer
+
+  RULES = {
+    "ROCK" => "SCISSORS",
+    "PAPER" => "ROCK",
+    "SCISSORS" => "PAPER"
+  }
```
tatsiana-makers, line 12: This is really neat!

rps.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  post '/paper' do
+    @game.player.choice("PAPER")
+    redirect '/result'
+  end
+
+  post '/scissors' do
+    @game.player.choice("SCISSORS")
+    redirect '/result'
+  end
```
tatsiana-makers, line 38: Each of these routes looks very similar and follows the same logic which doesn't follow the DRY principle. Can you think of a way we might have a single route which can handle all 3 moves?

spec/game_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+describe Game do
+  let(:player) { Player.new("Player double") }
+  let(:computer) { Computer.new }
+  subject(:game) { Game.create(player, computer) }
+
+  describe "#results" do
+    it "returns a draw when same weapon is used" do
+      player.choice("PAPER")
+      srand(4567)
```
tatsiana-makers, line 11: Cool use of `srand` 💥 !

spec/computer_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "computer"
+
+describe Computer do
+  it "chooses a random weapon" do
```
tatsiana-makers, line 4: Tests are a way of capturing the behaviour of our code. In this case, we want to capture that there are 3 possible outcomes for `weapon` so we could have unit tests for all three cases

tatsiana-makers (commented):
I've added a few comments, including one which points out what I suspect is the reason the app isn't behaving as you would expect it to behave. Let me know if you have any questions, happy to discuss my comments further 🙂

lib/game.rb
kabligh: Hi Tatsiana, my code reviewer this morning pointed this out so I did what you've suggested and stored @computer.weapon as a @computer_choice variable. Unfortunately, this doesn't fix the web app, though. My issue with the web app is that I can't get it to store my player's weapon – that's what I was trying to fix with the numerous post /rock/paper/scissor methods. But couldn't...

rps.rb
```diff
@@ -0,0 +1,45 @@
...
+  before do
+    @game = Game.instance
+  end
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/names' do
+    @game = Game.create(Player.new(params[:player_name]), Computer.new)
```
tatsiana-makers, line 17: This is exactly how you would connect the controller with a class in the model 🎉 

Is there a particular question you have about this?

lib/game.rb
tatsiana-makers: Can you get some visibility on what is happening?

For example, some things I might print out:
- `@game.player` in the different routes of `rps.rb` to check that in each of the places, you are using the exact same instance of Player (checking that the number after `0x` when printing the instance is the same everywhere, in #<Player:0x00007ff0f888a630>)
- `@game.player` before and after calling `.choice` to check if the choice is actually being stored in the `post` routes
-  `@game.player` in a different route after I've set the player's choice (for example in `get /result`) -> Can I see the player's choice stored? 

This should help give some visibility on what is happening and narrow down where or why the player's choice isn't being stored

kabligh: Thanks! This is really helpful. I actually managed to fix it but this would definitely have helped find the source of the problem.

PR #1766, opened 2020-11-16
Feedback please!

joshuaretallick:
Hi Coaches! I finished the first two basic user stories here (prior to adding Spock and LIzard)
Can I please have some feedback on stubbing my Game class to deal with randomness rather than using SRAND (or is one better than the other?)
Also I am pretty sure my testing isn't rigorous enough so suggestions on adding a few more.
Lastly I am storing data in sessions.  If I select to play a new game then hit submit for my move without actually typing in input it brings over the the last move I stored.  Can I reset sessions when I start a new game?
I also type in my move, 'rock', 'paper' etc, I guess adding buttons would be a better idea as it then removes the idea of the user making a typo and then crashing the game?


app.rb
```diff
@@ -0,0 +1,38 @@
...
+  get '/play' do
+    erb :play
+  end
+
+  post '/player-attack' do
+    session[:player_attack] = params[:player_attack]
+    redirect '/player-move'
+  end
+
+  get '/player-move' do
```
tatsiana-makers, line 25: Do we need this route?

What's the difference in purpose between this rout and `get /game_over`? Could we merge the two together and have one page which shows both which move each player selected and who won?

views/player_move.erb
```diff
@@ -0,0 +1,8 @@
+<% if session[:player_attack] == "Rock" %>
+<%= session[:player_1_name] %> chooses Rock!
+<% elsif session[:player_attack] == "Paper" %>
+<%= session[:player_1_name] %> chooses Paper!
+<% else session[:player_attack] == "Scissors" %>
+<%= session[:player_1_name] %> chooses Scissors!
+<% end %>
+<a href="/game_over">OK</a>
```
tatsiana-makers, line 8: Views are responsible for displaying information to the user. This logic would be better suited in a Model. 

Some steps for achieving this:
- pull this logic back into the controller
- extract into a Model out of the controller.

See my question above about whether we need this page at all though

spec/game_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    expect(dave.player_1_name).to eq "Dave"
+  end
+
+  it "initializes with teh player's move" do
+    expect(dave.player_attack).to eq "Rock"
+  end
+
+  it "calculates the win" do
+    srand(4)
+    allow(computer).to receive(:computer_move)
```
tatsiana-makers, line 17: You shouldn't use both `srand` and stubbing, as they are both meant to achieve the same thing -> make the `sample` method determinstic.

Here you're stubbing the attribute reader for `computer_move` and [returning `nil`](https://relishapp.com/rspec/rspec-mocks/v/2-99/docs/method-stubs/allow-with-a-simple-return-value#stub-with-no-return-value). So if we go with stubbing, there should definitely be an explicit return value


Potentially something that might be clearer if we do go with stubbing would be to stub the `move` method. There would be no difference in logic but in my opinion, reading
`allow(computer).to receive(:move).and_return("Scissors")` would be clearer since `move` is actually the method which determines the outcome while `computer_move` is just reading out the result

spec/game_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  it "initializes with the player name" do
+    expect(dave.player_1_name).to eq "Dave"
+  end
+
+  it "initializes with teh player's move" do
+    expect(dave.player_attack).to eq "Rock"
+  end
+
+  it "calculates the win" do
```
tatsiana-makers, line 15: We should test all the possibilities - winning, tying, and losing.

spec/game_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'game'
+
+describe Game do
```
tatsiana-makers, line 3: We should test all public methods in class. Are there any methods that are currently missing tests?

views/game_over.erb
```diff
@@ -0,0 +1,9 @@
+<%= @game_over %>
+</br>
+<%= session[:player_1_name] %>: <%= session[:player_attack] %>
+</br>
+Computer: <%= @computer_move %>
+</br>
+<a href="/play">Try Again?</a>
+</br>
+<a href="/">New Player?</a?
```
tatsiana-makers, line 9: I couldn't replicate the problem you were seeing with sessions- when I clicked this button and played a new game with a new name and move it all worked as I expected 🤔 

What I would recommend - can you add `p` statements to follow the flow of control through the routes and get visibility on what the values in sessions are. This should help you determine what is going wrong and where

tatsiana-makers (commented):
Thanks for asking for feedback - I've left some comments and suggestions! Happy to have a chat about them after you've had a chance to read through and try out the suggestions 🙂

app.rb
joshuaretallick: TBH I have no idea why I did this, I guess for added user experience, but if you type in your move I guess you don't need it read back to you!

views/player_move.erb
joshuaretallick: Yup perfect, resolved by doing the above.  I guess it was a process I began with and then could have refactored later.  I have seen some examples of people putting this sort of logic in to app.rb. I guess that would also not be the done thing?

tatsiana-makers: `app.rb` is the controller of the app so this logic could start in the controller and then be refactored out into a model, as this is where it should belong. This is the step of battle where some extraction is done: https://github.com/makersacademy/course/blob/master/intro_to_the_web/extracting_logic_to_the_model.md

PR #1776, opened 2020-12-13
Completed initial stage of development - MVP established and feature tested.

lukewickens1989:


public/css/style.css
```diff
@@ -0,0 +1,101 @@
+* {
```
Will-Helliwell, line 1: This comment goes for all your html and css - I can't read it but it is BEAutiful

spec/features/functionality_testing_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+end
+
+RSpec.feature "Choice", :type => :feature do
+
+  scenario 'checking for "Rock" option' do
+    sign_in_and_play
+    visit('/plays')
+    choose(:Rock)
+    click_button('Lets Play')
```
Will-Helliwell, line 16: Not sure if you need lines 14-16 because they are covered by sign_in_and_play?

spec/features/functionality_testing_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  scenario 'checking for "Scissors" option' do
+    sign_in_and_play
+    visit('/plays')
+    choose(:Scissors)
+    click_button('Lets Play')
+    expect(page).to have_content 'You chose Scissors.'
+  end
+
+  scenario "making a selection" do
```
Will-Helliwell, line 36: I think this test is a repeat of the test on line 12

views/result.erb
```diff
@@ -0,0 +1,36 @@
...
+  </head>
+  <body>
+  <div class="login_div">
+
+    <% @random_choice = ["Rock", "Paper", "Scissors"].sample %> 
+
+    You chose <%= @choice %>.<br>
+    The computer chose <%= @random_choice %>.<br>
+
+    <% if @choice == "Rock" && @random_choice == "Scissors" || @choice == "Scissors" && @random_choice == "Paper" || @choice == "Paper" && @random_choice == "Rock" %>
```
Will-Helliwell, line 22: At first I thought you had mis-spelled 'MVC' but now I've learned something for the day...I like the fact that you've gone for an entirely different architecture with the MVP!

My understanding is based on 5 minutes of reading and so probably wrong, but the article I have read suggests that the Ruby logic should be located in the 'presenter' rather than the 'views'. Does this mean it should be in your CSS files? It would be good to pick your brains this afternoon!

spec/features/functionality_testing_spec.rb
```diff
@@ -0,0 +1,70 @@
+require "./rps.rb"
+
+RSpec.feature "Username", :type => :feature do
+  scenario "User inputs name" do
+    sign_in_and_play
+    expect(page).to have_content 'Welcome Thrash!'
```
Will-Helliwell, line 6: A reference to the Atlanta Thrashers mascot from 1999-2011 perhaps?
https://en.wikipedia.org/wiki/List_of_NHL_mascots#Thrash

Will-Helliwell (commented):
I've run it on my machine - so beautiful! Great job, especially considering you didn't have much time this weekend!

PR #1778, opened 2020-12-13
Javi's solution

xavierloos:


app.rb
```diff
@@ -0,0 +1,83 @@
...
+
+class RockPaperScissors < Sinatra::Base
+  enable :sessions
+  get "/" do
+    erb :index
+  end
+  get "/singleplayer" do
+    erb :singleplayer
+  end
+  get "/multiplayer" do
```
PiperS52, line 13: Really nice - I like the multiplayer option

app.rb
```diff
@@ -0,0 +1,83 @@
...
+  get "/singlegame" do
+    @name = session[:name]
+    erb :singlegame
+  end
+
+  get "/multigame" do
+    @name_1 = session[:name_1]
+    erb :player_1
+  end
+
```
PiperS52, line 37: Really concise get/do post/do methods

lib/multigame.rb
```diff
@@ -0,0 +1,16 @@
+class MultiGame
+  attr_accessor :p1_selected, :p2_selected
+
+  def initialize(p1_selected = p1_selected, p2_selected = p2_selected)
+    @p1_selected = p1_selected
+    @p2_selected = p2_selected
+  end
+
```
PiperS52, line 8: Really nice separation of classes. Could maybe also put the computer/sampler into a separate class as well.

public/css/index.css
```diff
@@ -0,0 +1,56 @@
+body {
+  background-image: linear-gradient(to top, #30cfd0 0%, #330867 100%);
+
+  font-size: larger;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  font-family: "Dancing Script", cursive;
+}
```
PiperS52, line 10: Super slick design - the game looks amazing in the browser

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,44 @@
...
+end
+
+def multiplayer
+  visit("/")
+  click_button "Multiplayer"
+  fill_in :name_1, with: "Javi"
+  fill_in :name_2, with: "Lopez"
+  click_button "Play"
+end
+
```
PiperS52, line 15: Good use of different methods in web_helpers

views/results.erb
```diff
@@ -0,0 +1,48 @@
+
+
```
PiperS52, line 2: Nice structure/design/layout in your erb files

PiperS52:
Nice work. The app functions really well and looks super slick.

spec/game_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "game"
+describe Game do
```
Jestfer, line 2: We should test all public methods in class. Are there any methods that are currently missing tests?

spec/multigame_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+describe MultiGame do
+  it { is_expected.to respond_to :playing }
+  it "no moves" do
+    multigame = MultiGame.new(nil, nil)
+    expect(multigame.p1_selected).to eq(nil)
+    expect(multigame.p2_selected).to eq(nil)
+  end
+  it "both select the same" do
+    multigame = MultiGame.new("ROCK", "ROCK")
+    allow(multigame).to receive(:playing).and_return("You tied")
```
Jestfer, line 11: I am missing some expectations in these tests, what do you want to test in these?

views/index.erb
```diff
@@ -0,0 +1,28 @@
+<html lang="en">
```
Jestfer, line 1: I like that you have many views and each with a clear responsibility in terms of displaying info to the user, well done!

views/result.erb
```diff
@@ -0,0 +1,45 @@
...
+        <title>RPS</title>
+    </head>
+    <body>
+        <h1>Single Player:</h1>
+        <h2>ROCK, PAPER, SCISSORS</h3>
+        <div class="option-game col-10 col-md-6 col-sm-8">
+            <h3 class="m-4"><%= @result %></h3>
+            <div class="m-4 col-12 d-flex flex-wrap justify-content-center">
+                <div class="option">
+                    <h4>Your choice: </h4>
```
Jestfer, line 32: Why do you think you haven't been able to display the name of the winner? What did you try before deciding to keep it generic?

Jestfer:
Overall, your code reads well and does the job. 

What I would focus on is testing, do you try to follow the Arrange-Act-Assert (AAA) steps in each unit test you add to your code base? Remember that all public methods of a class (its public interface) should be tested.
Try to spend some time of your self-learning time revisiting some TDD concepts you've already seen, as this will be beneficial for you moving forward. I recommend you attend the process workshop this week too (should appear in your Cohort Calendar).

**Some more thoughts:**
- When you do not need to expose a particular method in a class, this can be private, and you can mark it as such in your class files if you add them below the `private` keyword in your file.
- Your README could be improved: some suggestions would be adding useful bits of documentation to show the user how your program works: diagrams, screenshots
- Nice CSS implementation, your views look good
- If you want to redesign your application to have a main `Game` class that connects with, for example, let's say `SingleGame` and `MultiGame` (which I think could be a good idea), I would suggest diagramming a bit more and that should make it easier for you to reimplement that, as adapting a whole codebase without making some sort of use of documentation or putting down your thoughts somewhere can get tricky.

spec/game_spec.rb
xavierloos: I am not testing the random choose from the computer, so I can't test when the computer or the player wins/match/lose

xavierloos: I am not testing the random choose from the computer, so I can't test when the computer or the player wins/match/lose

spec/multigame_spec.rb
xavierloos: If both select the same card it will be a match

views/result.erb
xavierloos: I think I need to pass it to the class as at the moment is only passing the selected card

PR #1782, opened 2020-12-13
Chris Whitehouse RPS Challenge

chriswhitehouse:


spec/units/game_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    let(:random_choice) { "Scissors" }
+    it 'should return a random choice' do
+      allow(game).to receive(:rand).and_return(2)
+      expect(game.random_choice).to eq random_choice
+    end
+  end
+
+  describe ".declare_winner" do
+    context "Player 1: Rock vs Player 2: Scissors" do
+      let(:player_2_double) { double :player, name: "Computer", choice: "Scissors" }
```
natashamcintyre, line 56: I'm learning so much from this way of stubbing 😀

views/choose.erb
```diff
@@ -0,0 +1,11 @@
+<h1><%=@game.player_1.name %>, Pick Your Weapon!</h1>
+<h2>Choose Rock, Paper, or Scissors: </h2>
+<form action='/play' method='get'>
+  <label for='rps'></label>
+  <select id='rps' name='rps'>
+    <option value='Rock'>Rock</option>
+    <option value='Paper'>Paper</option>
+    <option value='Scissors'>Scissors</option>
+  </select>
+  <input type='submit' value="Play">
```
natashamcintyre, line 10: I've just noticed that your index.erb you have an input type='submit' but then name= instead of value=. I need to look into the difference between these or whether they are the same.

spec/features/rps_spec.rb
```diff
@@ -0,0 +1,15 @@
+describe RPS do
+  feature "1. Registration" do
+    scenario 'I would like to register my name before playing an online game' do
+      sign_in
+      expect(page).to have_content "Player 1"
+    end
+  end
+
+  feature '2. Play' do
```
natashamcintyre, line 9: I might be wrong but I thought that each new feature needed to be in different files 🤷‍♀️

README.md
```diff
@@ -1,85 +1,137 @@
...
 ```
 
+## Installation
+Fork repository. Run bundle to install gems.
+
+## Tests
+100% Coverage. All tests passing.
+
+### Feature Tests:
```
natashamcintyre, line 96: Great idea to do a summary of your tests here

natashamcintyre (commented):
I've really enjoyed reading through your code. I thought it was easy to read with concise methods. I loved how easy your website was to use, including the drop down box. I've made a suggestion about the feature specs but Eoin has just made some stubbing suggestions too which may help with coverage of feature tests 👍

lib/player.rb
```diff
@@ -0,0 +1,11 @@
+class Player
+  attr_reader :name, :choice
+
+  def initialize(name)
+    @name = name
+  end
+
+  def add_choice(choice)
+    @choice = choice
```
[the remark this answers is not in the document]
chriswhitehouse: Nice

app.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    # $game = Game.new
+    player_1 = Player.new(params[:player_1_name])
+    player_2 = Player.new("Computer")
+    @game = Game.create(player_1, player_2)
```
tatsiana-makers, line 17: If the second player is always going to be a computer, we could use a default value for the second input parameter into `Game.create` so this line would look like `@game = Game.create(player_1)`. This would remove some of the implementation detail overhead from a reader of the code.

app.rb
```diff
@@ -0,0 +1,35 @@
+require 'sinatra/base'
+require './lib/player.rb'
+require './lib/game.rb'
+
+class RPS < Sinatra::Base
```
tatsiana-makers, line 5: You have a really nice separation into MVC components. Each is responsible for separate parts of the web app and all interact nicely

lib/game.rb
```diff
@@ -0,0 +1,51 @@
...
+    if player_1.choice == player_2.choice then
+      "Draw"
+    else
+      @players[winner].name + " Wins"
+    end
+  end
+
+  private
+
+  def winner
```
tatsiana-makers, line 43: This is minor, but perhaps a clearer name for this would be `winner_index` to indicate that this is going to return the index of the winner in the @players array

spec/features/rps_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+    scenario 'I would like to be able to play rock/paper/scissors' do
+      allow_any_instance_of(Game).to receive(:rand).and_return(2)
+      sign_in_and_choose
+      expect(page).to have_content "Fight! Rock vs Scissors Chris W Win!"
```
tatsiana-makers, line 13: As Natasha mentioned, currently the feature tests only cover the scenario when Chris W wins. It'd be great to cover a tie and loss as well to make sure the app works in all these possible situations

spec/units/game_spec.rb
```diff
@@ -0,0 +1,76 @@
+describe Game do
+  let(:player_1_double) { double :player, name: "Player 1", choice: "Rock" }
+  let(:player_2_double) { double :player, name: "Computer" }
+  let(:game_class) { described_class }
+  let(:game) { game_class.new(player_1_double, player_2_double) }
```
tatsiana-makers, line 5: Nice mocking!

spec/units/game_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    it 'should create a new instance of a game' do
+      expect(described_class.create(player_1_double, player_2_double)).to be_instance_of described_class
+    end
+  end
+
+  describe "self.instance" do
+    it 'should return instance of a game' do
+      described_class.create(player_1_double, player_2_double)
+      expect(described_class.instance).to be_instance_of described_class
```
tatsiana-makers, line 20: Can we make a stronger assertion here, that `described_class.instance` actually returns the exact instance created in the line above?
```
g = described_class.create(player_1_double, player_2_double)
expect(described_class.instance).to eq(g)
```

spec/units/game_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+  end
+
+  describe ".new" do
+    it 'should accept two player arguments' do
+      expect(game_class).to respond_to(:new).with(2).arguments
+    end
+
+    it 'should record both players in a players array' do
+      expect(game.players).to eq [player_1_double, player_2_double]
```
tatsiana-makers, line 30: Is this [testing behaviour or is it testing state](https://github.com/makersacademy/skills-workshops/blob/master/practicals/testing/behaviour_not_state.md#testing-behaviour-rather-than-state-and-why-you-should-care)? We should be testing end-user behaviour so that our tests don't tie us too closely to the exact current implementation. You could imagine that we could rewrite the code so that player2 comes first in the array and update other code so that the program still performs the end-user behaviour. If we did this, this would make the test case fail because it is state-driven rather than focussing on the behaviour

views/index.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Welcome to Rock, Paper, Scissors!<h1>
```
tatsiana-makers, line 1: https://medium.com/arren-alexander/adding-css-to-your-sinatra-project-f82190e26ca0 is a nice article about adding css to a sinatra project 🙂

tatsiana-makers (commented):
Thanks for asking for feedback! This is a nice submission for the RPS challenge, showing off a really nice separation of concerns in the MVC components. I've left a few other comments and suggestions. Let me know if you have any questions, happy to chat through them 🙂

README.md
```diff
@@ -1,85 +1,137 @@
...
+  * *should accept a name parameter*
+* .name
+  * *should return the player name*
+* .add_choice
+  * *should record the players choice*
+* .choice
+  * *should return the player choice*
+
+## How to use?
+Run rackup. Then type 'localhost:' with assigned port number into web browser.
```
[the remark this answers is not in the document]
chriswhitehouse: Ta!

app.rb
chriswhitehouse: Agree, that would be a neat solution. I did grapple with how to create the computer player a bit.

lib/game.rb
chriswhitehouse: Agree, I could make this a bit more human readable. And possibly update how the hash actually works.

spec/units/game_spec.rb
chriswhitehouse: Thanks!

spec/units/game_spec.rb
chriswhitehouse: Agree

spec/units/game_spec.rb
chriswhitehouse: Ok, thanks Tatsiana, I think I need to work on this a bit more. I'm not sure I've definitively got my head around it.

views/index.erb
chriswhitehouse: Thanks, I'll try this out on the next challenge 🤘

PR #1784, opened 2020-12-13
RPS App

iantfleming:
Due to illness, I was unable to complete as much as I would have liked, though I am happy with what I did get done.


views/namepage.erb
```diff
@@ -0,0 +1,13 @@
+<link rel="stylesheet" href="index.css">
+
+<h1><center>
+  Hi <%= @player_name %>!
+</center></h1>
+
```
ollienorman, line 6: Great styling/html - made me laugh!

spec/game_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'game'
+
+describe Game do
+  it 'creates an instance of Game' do
+    expect(subject).to be_kind_of(Game)
+  end
+
+  describe '#rps' do
```
ollienorman, line 8: I think this is not actually testing the .rb file - the code is just testing itself (via the 'allow' line). I tried breaking the code in the .rb file and it still passed. I find this quite difficult with mocks/stubs - I've given myself false positives quite a few times - I got round it on this occasion by using the 'srand' function

public/index.css
```diff
@@ -0,0 +1,8 @@
+body {
+  background-color: powderblue;
+}
+
+h1 {
+  color: black;
+  font-family: "Lucida Console", "Courier New", monospace;
```
ollienorman, line 7: Have you written HTML before or did you look this up while doing the challenge? It's really neat!

lib/game.rb
```diff
@@ -0,0 +1,11 @@
+class Game
+
+  def initialize
+    @game_array = ['rock', 'paper', 'scissors']
```
ollienorman, line 4: I was initially going to make the possible moves/shapes their own class but ended up deciding that was too bloated and went with the same route as you here - so I'm glad you agree!

views/namepage.erb
```diff
@@ -0,0 +1,13 @@
+<link rel="stylesheet" href="index.css">
```
ollienorman, line 1: And including css! This is really good

app.rb
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Rps < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do #box appears and stored in '/' page
```
ollienorman, line 12: Code very readable with comments etc

PR #1786, opened 2020-12-14
RPS challenge

PiperS52:


xavierloos (commented):
Excellent, I like the way you implement your TDD.

PR #1791, opened 2021-01-23
Rps challenge

MegGosnell:


app.rb
```diff
@@ -0,0 +1,30 @@
+require 'sinatra/base'
+
```
Reeshul, line 2: You're missing a `require './lib/opponent'` here
 
 The page breaks after selecting an option!

spec/features/opponent_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'opponent'
+
+describe Opponent do 
+    subject(:opponent) {described_class.new}
+
+    describe '#shape' do 
+        it 'returns a random shape' do 
+            expect(Opponent::SHAPES).to include opponent.shape
```
Reeshul, line 8: This tests that the opponent returns a shape, but not that it is random.

spec/features/opponent_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+        end 
+    end 
+    end
```
Reeshul, line 11: Indentation

spec/play_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+feature 'playing a game' do
+    PLAY_SEED = 221563
+    before do 
+        visit ('/')
+        fill_in :player_name, with: 'Meg'
+        click_button 'submit'
+        expect(page).to have_content 'Meg'
```
Reeshul, line 8: Is this line necessary? You've already tested for the name to  be there in the `registration_spec.rb` file.

spec/play_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+feature 'playing a game' do
+    PLAY_SEED = 221563
+    before do
```
Reeshul, line 4: Woop keep things DRY!

spec/play_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end 
+
+    scenario "Game chooses a random option" do 
+        srand(PLAY_SEED)
+        click_button "Rock"
+        expect(page).to have_content 'Opponent chose Scissors'
+    end 
+
+
+    def possible_messages
```
Reeshul, line 35: This could've gone in a different method helpers file. Feels a bit awkward to declare methods like this at the end of in a spec file.

spec/play_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end 
+
+    scenario "Game chooses 'Rock'" do 
+        click_button "Rock"
+
+        message = find("#opponent").text
+        expect(possible_messages).to include message
+    end 
+
+    scenario "Game chooses a random option" do
```
Reeshul, line 28: Now this tests for randomness!

Reeshul (commented):
Overall really good I think! Just pointed out some small things.

PR #1792, opened 2021-01-24
Rock Paper Scissors

mscwilson:


lib/game.rb
```diff
@@ -0,0 +1,38 @@
+require_relative "player"
+require_relative "playerbot"
+
+class Game
+
+  attr_reader :players, :results
+
+  def initialize(player1)
+    @player1 = player1
+    @player2 = PlayerBot.new(["RPSbot", "Murderbot", "Opponentbot", "Badloserbot", "Goodgamebot"].sample)
```
GlykeriaStr, line 10: I love the idea that the computer has a name! I might steal it!

public/style.css
```diff
@@ -0,0 +1,31 @@
+body {
```
GlykeriaStr, line 1: I had no idea you can put it in one file and call it...! Stealing this too! :)

lib/views/results.erb
```diff
@@ -0,0 +1,25 @@
...
+<% else %>
+
+<%= @game.players[0].name %> chose <%= @game.results[:player1] %> - <%= @game.players[1].name %> chose <%= @game.results[:player2] %>!
+<br>
+<strong><%= @game.winner.name %> is the winner</strong>!
+
+<% end %>
+<br>
+<br>
+<a href="/play"><button>Play again?</button></a>
```
GlykeriaStr, line 22: How did you do it to keep the name of the player when they play again?

PR #1797, opened 2021-01-24
Finished RPS game and edited webpages

RTurney:
I have completed a simple rock-paper-scissors game with an AI randomly selecting their choice. 
I have also edited the HTML's so that the webpages look a bit nicer. (I apologise that the colours are fairly awful!)

I welcome any comments or criticism's regarding this code. 
Thank you


lib/game.rb
```diff
@@ -0,0 +1,36 @@
+class Game
+  attr_reader :name
+
+  def initialize
+    @possibility_1 = ['Paper', 'Rock']
```
ddottyler, line 5: interesting approach

views/play_page.erb
```diff
@@ -0,0 +1,71 @@
...
+      <div style = "display: inline-block; margin: 40px;">
+        <img src="https://lh3.googleusercontent.com/8hhYjwyBpVguvmiPp2i_WknssoP_IwcH5NVY03pwW84fN5femDiioXUO0P2m-zDwsvpjFF0=s85">
+        <br>
+        <button type="submit" value="Scissors" name = "player_choice">
+          SCISSORS
+        </button>
+      </div>
+    </form>
+  </section>
+</body>
```
ddottyler, line 71: really impressive!

ddottyler (commented):
Really impressive running of the programme. The websites look great and everything runs really smoothly. Particularly like the option to play again and all the detail shared when the outcome has been decided. 
The testing looks really good too. Very thorough and very clear to read and understand. 
If I had one suggestion it would be to split the Game class up a bit as it looks like it's doing quite a lot of work.
Great overall!

app.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  post '/name' do
+    session[:player_name] = params[:player_name]
+    redirect 'play-page'
+  end
+
+  # main game page
+  get '/play-page' do
+    @player_name = session[:player_name]
+    $computer = Game.new
```
tatsiana-makers, line 20: Global variables are dangerous as it opens up the risk that we could accidentally change them from many places. [This is a step](https://github.com/makersacademy/course/blob/master/intro_to_the_web/killing_the_global_variable.md) which goes through removing the global variable which you might want to check out (:

lib/game.rb
```diff
@@ -0,0 +1,36 @@
...
+    @possibility_2 = ['Paper', 'Scissors']
+    @possibility_3 = ['Rock', 'Scissors']
```
tatsiana-makers, line 7: These could be class constants instead of instance variables

spec/features/results_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  scenario 'player chooses rock while AI chooses rock' do
+    srand(2)
+    sign_in_and_play
+    click_button 'Rock'
+    expect(page).to have_content('The AI has chosen Rock')
+  end
+
+  scenario 'player chooses paper while AI chooses paper' do
+    srand(1)
```
tatsiana-makers, line 16: Nice use of srand for testing!

lib/game.rb
```diff
@@ -0,0 +1,36 @@
...
+    possibility_checker
+  end
+
+  def player_wins?(player_choice, result)
+    player_choice == result 
+  end
+  private #-------------------------
+
+  def possibility_checker
+    if @game_array.sort == @possibility_1
```
tatsiana-makers, line 28: By sorting the array of the choices, we're losing track of which choice belongs to which player, so this might be something which might need to be changed if we wanted to pull more logic into the Model where the model determines which player won

app.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  # results page
+  post '/result' do
+    @choice = params[:player_choice]
+    @computer_choice = $computer.random_selection
+    @result = $computer.game_results(@choice, @computer_choice)
+    # declares winner name:
+    $computer.player_wins?(@choice, @result) ? (@winner = session[:player_name]) : (@winner = $computer.name)
+    # declares noone a winner if it's a draw 
+    @winner = 'Noone' if @result == "Noone! It's a Draw"
```
tatsiana-makers, line 32: The determination of which player is the winner feels like logic that should belong inside of the Model (the Game class). In order to do this, the Game would need to know the player names and have a method which returned the name of the winning player. This would mean that the code here might look something like:
```
@winner = $computer.determine_winner(session[:player_name], @choice, @computer_choice)
@result = $computer.winning_move
```

lib/game.rb
```diff
@@ -0,0 +1,36 @@
...
+    @game_array = [player_choice, game_choice]
+    return "Noone! It's a Draw" if player_choice == game_choice
+
+    possibility_checker
+  end
+
+  def player_wins?(player_choice, result)
+    player_choice == result 
+  end
+  private #-------------------------
```
tatsiana-makers, line 25: Nice use of private methods!

tatsiana-makers (commented):
This is a nice go at the challenge! I've left a few comments. Let me know if you want to chat further about any of them 🙂

app.rb
RTurney: Completely agree. I know global variables were meant to be big no-no's but I did not know a better approach to deal with my calling variables across pages. Thank you for forwarding this step however, I will read through it and edit my code with it later. Thank you!

app.rb
RTurney: Yes good point. I feel as though my SRP-ing could have been done alot better here. I will try to extract that into my Game class.

lib/game.rb
RTurney: Thank you for the feedback Tatsiana, much appreciated. I understand all your feedback and will try to implement it into my code some point soon! 
Thanks again!

PR #1799, opened 2021-01-24
RPS Challenge

HolsDuckett:
RPS weekend challenge - week 3 Makers.

### DIFF
--- a/Domain_Modelling
+++ b/Domain_Modelling
@@ -1,0 +1,73 @@
+
+Nouns
+------------------------------
+Property or owner of property?
+------------------------------
+Name/player  ->    Owner of property
+computer     ->    Owner of property
+game         ->    Owner of property
+Weapon (rock/paper/scissors)  -> Owner of property
+
+Actions
+------------------------------
+Owned by?
+------------------------------
+Add player name   ->  computer
+play              ->  game
+Choose weapon     ->  Computer & player
+Result      -> game
+
+
+Actions
+------------------------------
+Property it reads or changes
+------------------------------
+Add player name      ->   Submissions in server/ params > @player_1
+
+players_choice       ->   Players choice of weapon
+computer             ->   Random choice of weapons
+result               ->    Return winner
+
+
+Class diagrams for these user stories:
+Summarise what we have learned from the three tables in the previous section into class diagrams.
+
+Class
+------------------------------
+Game
+------------------------------
+Properties (instance variables)
+------------------------------
+@game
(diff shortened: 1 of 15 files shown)


Domain_Modelling
```diff
@@ -0,0 +1,46 @@
+
```
charliefischer, line 1: nice domain modelling to simplify what classes will be needed

lib/turn.rb
```diff
@@ -0,0 +1,24 @@
...
+
+attr_reader :player_name, :player_weapon, :computer_weapon
+
+RULES = { rock: :scissors,
+          paper: :rock,
+          scissors: :paper }
+
+  def initialize(options)
+    @player_name = options["player_name"]
+    @player_weapon = options["player_weapon"]
```
charliefischer, line 11: might be better SRP if these were moved into a Player class so that deals with storing the players name and their move?

app_rps.rb
```diff
@@ -0,0 +1,33 @@
+require 'sinatra/base'
```
charliefischer, line 1: controller seems much cleaner than my attempt - good use of sessions and no repetition of code

lib/turn.rb
```diff
@@ -0,0 +1,24 @@
...
+    @computer_weapon = options["computer_weapon"]
+  end
+
+#method to determine the winner of rock paper scissors using the RULES constant which returns the winners name
+  def winner
```
charliefischer, line 16: could simplify this to be RULES[@player_weapon] == @computer_weapon as this can be the only case for the player to win

spec/feature_tests/game_play_spec.rb
```diff
@@ -0,0 +1,67 @@
+# As a marketeer
```
charliefischer, line 1: just for ease of reading might be worth deleting out the comments once the tests pass or shortening them down?

spec/feature_tests/turn_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'turn'
+
+describe Turn do
+  subject(:turn) {described_class.new(options)}
+  #the doubles are going to be a hash of the session information stored in the instance variables within the get /lets_fight requestion from the game_play_spec.rb file
+  let(:options) {{"player_name" => "Holly", "player_weapon" => :rock, "computer_weapon" => :paper, "winner" => "Holly"}}
```
charliefischer, line 6: nice stubing

PR #1803, opened 2021-01-24
Joe's RPS challenge

JosephSpraggon:
Got the game running and working correctly. However didn't have time to write as many tests or expanding the game as far as I would have liked! Got stuck on on trying to get my sessions to work but managed to overcome & then got stuck using randomness in Ruby 🥴


JosephSpraggon:
Got the game running and working correctly. However didn't have time to write as many tests or expanding the game as far as I would have liked! Got stuck on on trying to get my sessions to work but managed to overcome & then got stuck using randomness in Ruby 🥴

PR #1805, opened 2021-01-25
RPS Challenge Oscar

OscarB89:


app.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class RPS < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    $name = Game.new(params[:name])
```
jshields123, line 11: I am not sure if we were supposed to use the global variables or not. Could use instance variable instead?

spec/feature/play_spec.rb
```diff
@@ -0,0 +1,41 @@
+feature 'Select Rock, Paper or Scissor' do
+
+  before do
+    visit('/')
```
jshields123, line 4: could use a web helper for this process like in the week challenge to keep it DRY.

spec/game_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  describe '#name' do
+    it 'returns the name' do
+      expect(oscar.name).to eq 'Oscar'
+    end
+  end
+
+  # describe '#option' do
+  #   it 'returns the option' do
+  #     expect(oscar.option).to eq :Rock
```
jshields123, line 14: Is this in here to be implemented or old code?

views/play.erb
```diff
@@ -0,0 +1,13 @@
+<%= @name %>
+
+<form action="/play" method="post">
+  <input type="submit" name="option" value="Rock">
+  <input type="submit" name="option" value="Paper">
```
jshields123, line 5: nice clean HTML and erb here.

jshields123 (commented):
Looks good for where you have got to, its been neatly tidied up and I like the structure of the play spec file. It shows where the app will progress to.

PR #1806, opened 2021-01-25
RPS Challenge

jshields123:
Logic in place but some has issues with displaying the wrong winner at times...

### DIFF
--- a/rps.rb
+++ b/rps.rb
@@ -1,5 +1,5 @@
 require 'sinatra/base'
-
+require './models/player.rb'
 class Rps < Sinatra::Base
   enable :sessions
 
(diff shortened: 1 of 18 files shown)


rps.rb
```diff
@@ -0,0 +1,35 @@
...
+class Rps < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb:index
+  end
+
+  post '/name' do
+    $player_1_name = Player.new(params[:player_1_name])
+    p params
```
OscarB89, line 14: Could do without putting params on screen

rps.rb
```diff
@@ -0,0 +1,35 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+
+    @player_1_name = $player_1_name.name
+
+    erb:play
+  end
+
```
OscarB89, line 24: Unnecessary blank lines

rps.rb
```diff
@@ -0,0 +1,35 @@
...
+  get '/move' do
+    @player_1_name = $player_1_name.name
+    @player_move = $player_1_name.player_move(params[:player_move])
+    $computer_move = Computer.new.computer_move
+    $result = Game.new.show_winner($computer_move, @player_move)
+    p $result
```
OscarB89, line 30: Think you should have created a post method as well as you are modifying the state of the server with computer and player moves (store the params with post and render with get).

spec/computer_spec.rb
```diff
@@ -0,0 +1,13 @@
+require './models/computer'
+
+describe Computer do
+
+  describe '#computer move' do
+    it 'generates a random move for computer' do
+      stub = double('computer')
+      allow(described_class).to receive(:computer_move) do
+        ["Rock", "Paper", "Scissors"].sample
+    end
```
OscarB89, line 10: Not sure about this stubbing. Shouldn't it return ["Rock", "Paper", "Scissors"].sample?

OscarB89:
Looks quite good to me. Could have run rubocop to format your code.

PR #1808, opened 2021-01-25
DT rps-challenge week 3 

ddottyler:


app.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  get '/scissors' do 
+    @player_1 = $player_1
+    @player_1.move("scissors")
+    $game = Game.new(@player_1)
+    @outcome = $game.turn(@player_1)
+    erb :scissors
+  end
+
```
RTurney, line 50: I see what you mean regarding the three separate pages for each outcome now. 
As this has resulted in three pages with the 'same' code on it (`<%= @outcomes %>`), I would suggest that instead of using links for each option, maybe use a submit button within a form? This would allow you to get a different value returned for each button pressed. Which could then be fed into your `@player_1.move(params[button])` method.

views/scissors.erb
```diff
@@ -0,0 +1 @@
+<%= @outcome%>
```
RTurney, line 1: This applies to all of your results pages but maybe add a link back to the main play page so that you can play another round? Otherwise, happy with these pages.

views/play.erb
```diff
@@ -0,0 +1,7 @@
+<h1>Welcome to Rock, Paper, Scissors Online</h1>
+
+<h2><%= @player_1%> vs CPU</h2>
```
RTurney, line 3: Love the use of `<h1>` and `<h2>` headers for your play page. Really adds to the presentation of the game!

views/index.erb
```diff
@@ -0,0 +1,13 @@
...
+
+<form action="/names" method="post">
+
+<label for="P1">
+  Player 1 name
+  <input type="text" name="P1">
+  </label>
+
+    <input type="submit" value="Create player">
+    </form>
```
RTurney, line 13: Love the naming value for this button. Looks really good when you run it as well

app.rb
```diff
@@ -0,0 +1,52 @@
...
+require './lib/game'
+
+class RPS < Sinatra::Base
+
+  enable :sessions
+ 
+  get '/test' do
+    'Testing infrastructure working!'
+  end
+
```
RTurney, line 12: Good way to test the infrastructure. Wish I'd thought of this

lib/player.rb
```diff
@@ -0,0 +1,22 @@
...
+  
+  def move(move)
+    if move == "rock"
+      @turn = move
+    elsif move == "paper"
+      @turn = move 
+    elsif move == "scissors"
+      @turn = move 
+    else 
+      puts "Move not recognised, try again."
```
RTurney, line 18: Good guard clause. But when will it be used if player's are just clicking the links?

spec/features/paper_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'Player move' do 
+  scenario 'player takes turn' do
```
RTurney, line 2: Maybe add a more descriptive name here such as, "player plays paper and loses to scissors"

spec/game_spec.rb
```diff
@@ -0,0 +1,12 @@
+require './lib/game'
+
+describe Game do 
+
+  subject(:game) { Game.new(p1) }
+  let(:p1) { double :player }
```
RTurney, line 6: Good use of doubles

RTurney (commented):
Yeah looks good! I downloaded and played it as well. Can confirm it works, passes all tests. You should check the rubocop though, it returned a number of errors that you might want to look at. Nothing game-breaking though.

PR #1812, opened 2021-01-25
Hamish Arro

HamishArro:
Sorry it's not all the way there yet!

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,20 @@
+require 'sinatra'
+
+class RPS < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/name' do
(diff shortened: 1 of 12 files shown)


app.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  get '/paper' do
+    erb(:paper)
+  end
+
+  get '/sissors' do
+    erb(:sissors)
+  end
+
```
KarstenFinlay, line 34: Interesting approach, I am guessing that you would use each of this pages depending on what the user chooses, and that would decide the winner?

KarstenFinlay (commented):
Seems to me that you have a good structure to build the rest of your project on. Looking forward to seeing how it turns out!

PR #1814, opened 2021-01-25
RPS Challenge 

louiswicks:


app.rb
```diff
@@ -0,0 +1,30 @@
...
+    erb(:register)
+  end
+
+  post '/play' do
+    session[:player_shape] = params[:shape].downcase.to_sym
+    session[:opponent_shape] = Opponent.new.shape
+    redirect '/register'
+  end
+
+  run! if app_file == $0
```
day-katy, line 28: I've seen this before but don't actually know what it means! Could you explain it to me?

spec/features/play_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  scenario "choose a shape" do
+    sign_in_and_submit
+    click_button "Rock"
+    expect(page).to have_content "You chose rock!"
+  end
+
+  scenario "game to chose an option" do
+    sign_in_and_submit
+    click_button "Rock"
+    message = find(:css, "#opponent").text.strip
```
day-katy, line 18: Oooh, what does this mean?

spec/features/play_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    expect(possible_messages).to include message
+  end
+
+  scenario "game to chose a random option" do
+    sign_in_and_submit
+    srand(221563)
```
day-katy, line 24: Basically my comments are all "please explain the thing to me" - this is no different! How'd you pick the number?

lib/opponent.rb
```diff
@@ -0,0 +1,8 @@
+class Opponent
```
day-katy, line 1: I may have missed something, but you only seem to use this in app.rb - is there a reason you're not using it in your Turn class?

lib/turn.rb
```diff
@@ -0,0 +1,34 @@
+class Turn
```
day-katy, line 1: I think your code is super clear, which is great! The only thing is that I don't think you've really stuck to the single responsibility principle, because the Turn class seems to include everything - like your player and your opponent. If you separated those out, not everything would be hardcoded into this class, and then you could change things around more easily.

PR #1816, opened 2021-01-25
Anna's RPS Challenge

acavalla:


spec/player_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'player'
+
+describe Player do
+  describe 'move' do
+    subject { described_class.new("Anna") }
```
CodeRed30, line 5: Consider naming your subject to make your tests more readable

views/play.erb
```diff
@@ -0,0 +1,23 @@
+<h1><%= @player.name %> is playing RPS! </h1>
+
+<h3><%= @player.name %> vs. Computer! </h3>
+<strong><%= @player.name %>'s turn</strong>
+
+<!-- <form action="/result">
+  <input type="submit" value="Rock!">
+  <input type="submit" value="Paper!">
+  <input type="submit" value="Scissors!">
```
CodeRed30, line 9: add name = hand to each of these for t single table to work and then use params[:hand] to access it

app.rb
```diff
@@ -0,0 +1,36 @@
+require 'sinatra/base'
+require_relative './lib/player'
+require_relative './lib/game'
+require_relative './lib/computer'
+
+class RPS < Sinatra::Base
+  enable :sessions
+  set :session_secret, 'super secret'
```
CodeRed30, line 8: Consider instantiating Game and running the app through the game so you don't have to use sessions. (last step in battle explains it well)

PR #1817, opened 2021-02-20
Ian_McNicholas rps-challenge

ianmcnicholas:
As usual for me, there are a shockingly small number of commits.
I wasn't able to figure out how to create a double to control the random nature of the computer choices.  This might have been achieved if I created a game class.
I didn't create game class, which would have been beneficial to control some logic.

### DIFF
--- a/lib/computer.rb
+++ b/lib/computer.rb
@@ -1,0 +1,9 @@
+class Computer
+
+attr_reader :result
+
+def initialize
+  @result = ["rock", "paper", "scissors"].sample
+end
+
+end
--- a/lib/player.rb
+++ b/lib/player.rb
@@ -1,0 +1,9 @@
+class Player
+
+  attr_reader :playername
+
+  def initialize(playername)
+    @playername = playername
+  end
+
+end
--- a/spec/features/testing_infrastructure_spec.rb
+++ b/spec/features/testing_infrastructure_spec.rb
@@ -1,0 +1,60 @@
+feature "testing home page is working" do
+  scenario "displays welcome text" do
+    visit "/"
+    expect(page).to have_content "Rock, Paper, Scissors"
+  end
+  scenario "allows user to start a game" do
+    visit "/"
+    expect(page).to have_button("Let's go!")
+  end
+end
+
+feature "test Game page is working" do
+  scenario "greets user by name if provided" do
+    sign_in_and_play
+    expect(page).to have_content "It's Ian against the computer"
+  end
+  scenario "allows there to be no name provided" do
+    visit "/"
+    click_on("Let's go")
+    expect(page).to have_content "It's you against the computer"
+  end
+  scenario "displays 'rock' link" do
+    sign_in_and_play
+    expect(page).to have_button('Rock')
+  end
+  scenario "displays 'paper' link" do
+    sign_in_and_play
+    expect(page).to have_button('Paper')
+  end
+  scenario "displays 'scissors' link" do
+    sign_in_and_play
+    expect(page).to have_button('Scissors')
+  end
+end
+
+feature "playing the game" do
+  scenario "player chooses Rock" do
+    sign_in_and_play
+    click_on("Rock")
+    expect(page).to have_content "You chose Rock"
+  end
+  scenario "player chooses Paper" do
+    sign_in_and_play
+    click_on("Paper")
+    expect(page).to have_content "You chose Paper"
+  end
+  scenario "player chooses Scissors" do
+    sign_in_and_play
+    click_on("Scissors")
+    expect(page).to have_content "You chose Scissors"
+  end
+end
+
+feature "upon seeing the result" do
+  scenario "allows player to play again" do
+    sign_in_and_play
+    click_on("Scissors")
+    expect(page).to have_button('Play again!')
+  end
+end
--- a/spec/web_helpers.rb
+++ b/spec/web_helpers.rb
@@ -1,0 +1,5 @@
+def sign_in_and_play
+visit "/"
+fill_in('name', with: 'Ian')
+click_button("Let's go!")
+end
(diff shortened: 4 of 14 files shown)


views/playgame.erb
```diff
@@ -0,0 +1,25 @@
+<center><h1>Rock, Paper, Scissors</h1>
+<% if @@name == "" %>
+<h2>It's you against the computer</h2>
+<% elsif %>
+<h2>It's <%= @@name %> against the computer</h2>
+<% end %>
```
jai-jk, line 6: In order to align with MVC, do you think you could've moved the logic here down to the Model Layer?

views/rock.erb
```diff
@@ -0,0 +1,20 @@
...
+
+<% if @@computer == "rock" %>
+<h2>The rocks bounce off each other.  It's a tie!</h2>
+
+<% elsif @@computer == "paper" %>
+<h2>Your rock is wrapped up tight!  You lose!</h2>
+
+<% elsif @@computer == "scissors" %>
+<h2>Your rock crushes the scissors.  You win!</h2>
+<% end %>
```
jai-jk, line 14: Also here - in order to align with MVC, do you think you could've moved the logic here down to the Model Layer?

views/scissors.erb
```diff
@@ -0,0 +1,20 @@
...
+
+<% if @@computer == "rock" %>
+<h2>Your scissors are crushed by the rock.  You lose!</h2>
+
+<% elsif @@computer == "paper" %>
+<h2>Your scissors cut the paper to shreds.  You win!</h2>
+
+<% elsif @@computer == "scissors" %>
+<h2>It's a dead heat between these shearers.  It's a tie!</h2>
+<% end %>
```
jai-jk, line 14: You know how it goes, logic to Model layer?

views/paper.erb
```diff
@@ -0,0 +1,21 @@
...
+
+<% if @@computer == "rock" %>
+<h2>You wrap up the rock with your paper.  You win!</h2>
+
+<% elsif @@computer == "paper" %>
+<h2>The pages brushed past each other.  It's a tie!</h2>
+
+<% elsif @@computer == "scissors" %>
+<h2>Your paper is cut to shred by the scissors.  You lose!</h2>
+<% end %>
```
jai-jk, line 14: In order to align with MVC, do you think you could've moved the logic here down to the Model Layer?

jai-jk (commented):
Everything is thorough, clean, and succinct. Only a few little things you could've done to make your code more efficient. Good work!

lib/computer.rb
```diff
@@ -0,0 +1,9 @@
+class Computer
```
siellsiell, line 1: This class only has an initialiser and no other methods -- is it truly necessary to put this code inside a class and store state that you will only access once? How might you have encapsulated it differently?

lib/player.rb
```diff
@@ -0,0 +1,9 @@
+class Player
```
siellsiell, line 1: Same question here. What does this class add that isn't already provided by having a variable with the username in it?

spec/features/testing_infrastructure_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  scenario "allows player to play again" do
+    sign_in_and_play
+    click_on("Scissors")
+    expect(page).to have_button('Play again!')
+  end
+
+  scenario "tells the player the outcome of the game" do
+    srand(4)
+    computer = double(:computer)
+    allow(computer).to receive(:result).and_return("scissors")
```
siellsiell, line 64: This shouldn't be necesssary if you're also using `srand` to control the random outcome (see https://ruby-doc.org/core-2.4.0/Random.html#method-c-srand). Since this is a feature test, you shouldn't be mocking the `Computer` class if you can avoid it. Does your test still pass if you remove these lines?

spec/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def sign_in_and_play
+  visit "/"
+  fill_in('name', with: 'Ian')
+  click_button("Let's go!")
+end
```
siellsiell, line 5: Nice helper! Since you're only using it in one file though, it's best to put it in that file rather than separating it out. It makes your tests easier to follow (right now I have to jump to another file to figure out what your test is really testing). You should aim for tests to be as self-contained as possible.

spec/features/testing_infrastructure_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    sign_in_and_play
+    click_on("Rock")
+    expect(page).to have_content "win"
+  end
+end
```
siellsiell, line 69: I would expect some more tests to cover all other possible scenarios of the game.

views/paper.erb
```diff
@@ -0,0 +1,21 @@
...
+
+<form action="/playagain" method="post">
+<input type="submit" value="Play again!">
+</form>
+
```
siellsiell, line 19: Since this block is repeated across all of the "weapons" templates, maybe you can find a way to share it between them.

siellsiell (commented):
Hi Ian,

Thanks for working on this challenge! You've got a nice solution. Regarding your questions:

- Is my test suite sufficient?
You've got a good set of feature tests but you could have more!

- Have I followed the principle of separation of concerns?
Like Jai commented and like you mentioned in your PR description, I would argue that you could do with separating your game logic from your presentation logic by pulling it out of the views and into a its own game logic class/module.
**However**, given that you did put logic in your views, I think it was a nice approach to make separate erb files for each of the cases (rock/paper/scissor) as it makes the logic quite neat and easy to follow.

- Am I making good progress for this stage of the course? (end of week 3)
I think so, yes! But I'm more curious about how you feel about your progress and how you would assess it?

Regarding your question about `srand` -- what is your current understanding of what `srand` does?

PR #1819, opened 2021-02-21
Iain's RPS challenge

iainaitken:
Please see the README for notes on the design/use of the app and some reflection on how I feel I've done.


app.rb
```diff
@@ -0,0 +1,75 @@
+require 'sinatra/base'
+require_relative 'lib/classes.rb'
+
+class RockPaperScissors < Sinatra::Base
+
+  before do
+    @game = Game.instance
```
tatsiana-makers, line 7: Nice job getting rid of global variables!

lib/player.rb
```diff
@@ -0,0 +1,16 @@
+class Player
+  attr_reader :name, :weapon
+
+  def initialize(name)
+    if name != nil
+      @name = name
+    else
+      @name = "Computer"
+    end
```
tatsiana-makers, line 9: Can these lines can be replaces with a default value in the `initialize` input parameter?
`def initialize(name = "Computer")`

lib/calculator.rb
```diff
@@ -0,0 +1,19 @@
...
+    'scissors' => 'paper',
+    'paper' => 'rock'
+  }
+
+  def initialize(player_1_weapon, player_2_weapon)
+    @p1 = player_1_weapon
+    @p2 = player_2_weapon
+  end
+
+  def result
```
tatsiana-makers, line 13: This is a neat way of calculating the result

lib/random.rb
```diff
@@ -0,0 +1,11 @@
+class RandomWeapon
+  CHOICES = ['rock', 'paper', 'scissors']
```
tatsiana-makers, line 2: The strings `'rock'`, `'paper'`, and `'scissors'` appear both in this class and in the Calculator class. Rather than having these strings floating around, we could create constants for each of these strings in one place and use that constant in all the places we need that string. This would eliminate the chance that a typo in one of the strings in one of the classes breaks the logic of the game

spec/features/multiplayer_spec.rb
```diff
@@ -0,0 +1,11 @@
+feature 'multiplayer' do
+  scenario 'two people can play each other' do
+    multiplayer_sign_in
+    expect(page).to have_content("Kevin vs. Nigel")
+  end
+
+  scenario 'multiplayer game is possible' do
+    multiplayer_sign_in_and_play
+    expect(page).to have_button("Play again?")
```
tatsiana-makers, line 9: Should we also assert that the correct winner is declared on this page?

spec/features/play_game_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+  scenario 'player can choose rock, paper or scissors' do
+    single_player_sign_in
+    expect(page).to have_button("rock")
+    expect(page).to have_button("paper")
+    expect(page).to have_button("scissors")
+  end
+
+  scenario 'player choice takes them to the next page' do
+    single_player_sign_in_and_play
+    expect(page).to have_button("Play again?")
```
tatsiana-makers, line 11: Similarly here, we could assert that the right winner will be declared on this page.

tatsiana-makers (commented):
This is a nice go at the challenge, thanks for asking for review! Let me know if you have any questions on the comments I've left (:

lib/player.rb
iainaitken: HI Tatsiana - I tried that, but it would be overwritten by 'nil' because of the way app.rb was working. I could probably refactor all of that but didn't have time, this was just my workaround :)

lib/random.rb
iainaitken: Great idea!

lib/player.rb
tatsiana-makers: Ah, good point! Makes sense (:

PR #1821, opened 2021-02-21
Joey's rock paper scissors ✂️ 

jshorns:


README.md
```diff
@@ -1,5 +1,15 @@
...
+* The Player class does not currently have any responsibility other than storing a name. My thought is that this class could be expanded, and the game history stored in the game log made use of, if the game were to be expanded to involve multiple players, tournaments, etc etc.
+* I have opted for having the game all take place on a single page, rather than taking the user to a results page.
+* For storing the player move, I initially used independent get requests that then put the move into the model logic and redirect to the play page with the result.  i had intended to use a post request and store the move as a parameter, but I couldn't figure out how to. Ultimately I realised I could just type the parameter into the form action manually!
+* My chief problem is that rubocop doesn't like my logic for getting the result of the game as it thinks it's over complicated. I changed it from a case..when formula to its current form but this didn't make rubocop any happier. I can't think of any other way to do it!
+* The first player to 10 wins, and the player has the option to start again. if they win, they get $100,000...maybe
+
```
tatsiana-makers, line 12: Really cool to read about your approach to the challenge, thanks for including it here!

README.md
```diff
@@ -1,5 +1,15 @@
...
+* The Player class does not currently have any responsibility other than storing a name. My thought is that this class could be expanded, and the game history stored in the game log made use of, if the game were to be expanded to involve multiple players, tournaments, etc etc.
+* I have opted for having the game all take place on a single page, rather than taking the user to a results page.
+* For storing the player move, I initially used independent get requests that then put the move into the model logic and redirect to the play page with the result.  i had intended to use a post request and store the move as a parameter, but I couldn't figure out how to. Ultimately I realised I could just type the parameter into the form action manually!
+* My chief problem is that rubocop doesn't like my logic for getting the result of the game as it thinks it's over complicated. I changed it from a case..when formula to its current form but this didn't make rubocop any happier. I can't think of any other way to do it!
+* The first player to 10 wins, and the player has the option to start again. if they win, they get $100,000...maybe
```
tatsiana-makers, line 11: 😂

app.rb
```diff
@@ -0,0 +1,33 @@
+require 'sinatra/base'
+require './lib/rps.rb'
+require './lib/player.rb'
+require './lib/game_log.rb'
+
+class RockPaperScissors < Sinatra::Base
+  enable :sessions
+  before do
+    @game = RPS.game
```
tatsiana-makers, line 9: Nice use of the Singleton pattern to avoid global variables!

lib/game_log.rb
```diff
@@ -0,0 +1,40 @@
+class GameLog
+  attr_reader :score, :games
+
+  def initialize
+    @games = []
+    @score = { player: 0, computer: 0 }
+  end
+
+  def add_game(player_move, computer_move, result)
+    @games << [player_move, computer_move, result]
```
tatsiana-makers, line 10: The risk with representing the games as arrays is that it is easy to mix up the order of the moves input into the `add_game` method and this will mess up everything. Is there another way of representing a game that you can think of which will remove this risk?

lib/game_log.rb
```diff
@@ -0,0 +1,40 @@
...
+    get_score(result)
+  end
+
+  def game_over?
+    true if @score[:player] == 10 || score[:computer] == 10
```
tatsiana-makers, line 15: 10 is currently a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)). We could create a well-named constant to represent this value instead

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
+class RPS
+  attr_reader :game_log, :player, :result
+
+  def self.create(player, game_log)
```
tatsiana-makers, line 4: We could add a default value for `game_log`

spec/features/playing_a_game_spec.rb
```diff
@@ -0,0 +1,21 @@
+feature 'user can play a game of rock paper scissors' do
```
tatsiana-makers, line 1: Nice work testing the three possible outcomes

spec/unit/rps_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rps'
+
+describe RPS do
+  let(:bimini) { double(:player, :name => "BIMINI BON BOULASH") }
+  let(:game_log) { double(:game_log, :games => [], :add_game => []) }
```
tatsiana-makers, line 5: Great use of doubles

lib/rps.rb
```diff
@@ -0,0 +1,41 @@
...
+private
+
+  def random_move
+    ["rock", "paper", "scissors"].sample
+  end
+
+  def get_result(player, comp)
+    if player == comp
+      :draw
+    elsif (player == "rock" && comp == "scissors") ||
```
tatsiana-makers, line 32: [Using a hash ](https://github.com/makersacademy/rps-challenge/blob/master/docs/review.md#use-of-ifelsif-conditionals-for-business-logic)is a cool way of representing the rules and determining the winner of the game. This is probably similar to the way that you mentioned someone showed you 🙌

tatsiana-makers (commented):
Thanks for asking for a review, Joey! I've left a few comments. Do reach out if you have questions about any of them 🙂

lib/game_log.rb
jshorns: I changed this to just take a single argument, and then changed the RPS class itself to represent the moves and result of the game as a hash - is that what you're getting at? just want to be sure I'm not barking up the wrong tree 😅

lib/game_log.rb
jshorns: I've amended this - the only trouble I'm having with refactoring it out is in the features tests for the game being over. I can't figure out how to access the GameLog constance from within them. Could I refer to the game instance variable like I have done in the view files? I.e. just have it as @game.game_log.first_to?

lib/rps.rb
jshorns: I actually did this initially and then got rid of it, I don't know why! have changed it back now

lib/rps.rb
jshorns: Yes that's pretty much it! I hope you'll forgive me if I don't actually implement this change - it felt like a bit too much of an undertaking for nearly 6pm on a Monday! Maybe one day...

lib/game_log.rb
tatsiana-makers: What I was thinking of as alternatives to an array:
- a hash because you have named key-value pairs
- instances of classes because you'll have named instance variables

The RPS class could potentially work for this, though the class currently has a `@game_log` instance variable inside of it which might make things confusion with a sort of cyclic dependency. Perhaps creating a new class like a `GameOutcome` class might be an easier change to make at first? and then see if it makes sense to connect this back to the RPS class?

lib/game_log.rb
tatsiana-makers: You should be able to access the constant from the test files in the same way as from other ruby files. I suspect you'll have to require the file that the constant is defined in and then use the format `ClassName::CONSTANTNAME`

PR #1822, opened 2021-02-21
Jai's RPS

jai-jk:
What I did well: 

- Followed OOP principles well i.e. separation of concerns, skinny controllers, SRP, Dependency Injection, etc.
- Completed the first two user stories.
- Added design features.
- Wrote code that is easy to read and change.

What I would improve on:

- Even though I completed the basic version, I didn't refactor my global variables into Class Instance variables and that's not really negotiable..
- Could've done the advanced version.
- Could've made my game class more effective i.e. break the results method down further to be easier to change and read, and to adhere to OOP principles more closely.


ianmcnicholas:
* Good notes on the pull request - this is the first thing I saw.
* Good number of commits with good explanations
* Project could benefit from checking a couple of edge cases e.g. what if the user doesn't enter their name?
* Nice style. Nice css
* Good knowledge of views and separating styling sheets out.
* In the computer spec, there is availability to implement a testing stub for the random sampling.
* 100 test coverage - great.
* Good to see srand in gameplay_spec and tested against all results.
* Good use of constant for choices in the comp.rb  - allows for change.
* game.rb - great idea to manage the outcome. 
* Project would benefit from a detailed README file instructing on use along with approach considerations.

lib/game.rb
```diff
@@ -0,0 +1,20 @@
+class Game
+
+  def result
```
tatsiana-makers, line 3: You could capture the rules in a hash such as the one shown [here](https://github.com/makersacademy/rps-challenge/blob/master/docs/review.md#use-of-ifelsif-conditionals-for-business-logic), which would remove the need to have a separate if/else if case for every pair of moves

spec/features/gameplay_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  scenario 'Player draws' do
+    click_button('Scissors')
+    expect(page).to have_content "You both drew the same hand!"
+  end
+
+  scenario 'Player loses' do
+    click_button('Paper')
+    expect(page).to have_content "You lost! :("
+  end
```
tatsiana-makers, line 47: Nice work testing the win/lose/draw possibilities

tatsiana-makers (commented):
This is a good go at the challenge! I've left some comments, let me know if you have any questions 🙂

PR #1825, opened 2021-02-21
Week 3 weekend challenge submission

kerisic:
User stories completed and implemented two bonus features: RPSLS and multiplayer mode.

Testing and refactoring was less vigorous for the bonus features as not enough time - may be issues with encapsulation and visibility in RPS class code when implementing multiplayer mode. 

Wonder if paths in controller can be simplified having created a few more for multiplayer mode.

Had hoped to refine the user interface more however html proficiency currently limited.


app.rb
```diff
@@ -0,0 +1,57 @@
...
+    redirect '/choose'
+  end
+
+  get '/choose' do
+    erb :choose
+  end
+
+  post '/check' do
+    @game.player_choose(params[:weapon])
+    @game.multiplayer == true ? (redirect '/choose2') : (redirect '/result')
```
kjjwhitlock, line 27: great ternary operator!

app.rb
```diff
@@ -0,0 +1,57 @@
...
+    @game = RPS.instance
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    player = Player.new(params[:name])
+    @game = RPS.create(player)
```
kjjwhitlock, line 17: Could possibly have clearer naming of the create method e.g. def new_game

lib/weapon.rb
```diff
@@ -0,0 +1,24 @@
...
+  RULES = { rock: [:scissors, :lizard],
+            paper: [:rock, :spock],
+            scissors: [:paper, :lizard],
+            lizard: [:paper, :spock],
+            spock: [:rock, :scissors] }
+
+  attr_reader :type
+
+  def initialize(type)
+    @type = type.downcase.to_sym
```
kjjwhitlock, line 14: not sure what type means, maybe a clearer name for this one too

lib/computer.rb
```diff
@@ -0,0 +1,11 @@
+require_relative 'weapon'
+
+class Computer
+  def weapon
+    @weapon = Weapon::WEAPONS.sample
```
kjjwhitlock, line 5: this is genius please explain it to me shortly

lib/rps.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def self.create(player)
+    @rps = RPS.new(player)
+  end
+
+  def self.instance
+    @rps
+  end
+
+  def switch_to_multiplayer
```
kjjwhitlock, line 26: good method name

spec/weapon_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'weapon'
+
+describe Weapon do
+  let(:rock1) { double('rock1', type: :rock) }
+  let(:scissors) { double('scissors', type: :scissors) }
+  subject(:rock) { described_class.new('rock') }
```
kjjwhitlock, line 6: smart way of testing with doubles

kjjwhitlock (commented):
This is so good, and really smart how it all flows together. Not much to suggest in terms of improvement, but maybe a couple of clearer method names.

lib/player.rb
```diff
@@ -0,0 +1,7 @@
+class Player
+  attr_reader :name
+
+  def initialize(player)
+    @name = player.capitalize
+  end
+end
```
siellsiell, line 7: this class doesn't really do much, I wonder whether it's really needed. Perhaps it would be more useful if it could represent a computer player as well? Could be a case for using [inheritance or composition](https://github.com/pruett/ruby-patterns/blob/master/concepts/inheritance-vs-composition.md).

spec/features/welcome_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'Welcomes the user' do
+  scenario "welcomes and asks for user's name to be shown later" do
+    visit '/'
+    fill_in :name, with: 'Chrissy'
+    click_button 'Ready'
+    expect(page).to have_content 'Chrissy'
```
siellsiell, line 6: You could be using your `sign_in_and_play` helper here

lib/rps.rb
```diff
@@ -0,0 +1,48 @@
+require_relative 'computer'
+require_relative 'weapon'
+require_relative 'player'
+
+class RPS
+  attr_reader :computer, :player_choice, :player2_choice, :player, :player2, :multiplayer
+
+  def initialize(player, weapon_class = Weapon, computer = Computer.new)
```
siellsiell, line 8: It would be nice to find a way to design this class in such a way that it doesn't have to care whether a game is multiplayer or not. It just receives two players that can make choices and doesn't need to know whether the choices come from a computer or not. The concern of whether the player is a computer or not and how the computer makes choices should be encapsulated inside another piece of code.

Also see my comment on your `Player` class, which relates to this.

spec/rps_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+  end
+
+  describe 'when player has made a choice' do
+    it 'creates a new weapon from the choice' do
+      expect(weaponclass).to receive(:new)
+      subject.player_choose(rock)
+    end
+
+    context 'check the result with #result' do
```
siellsiell, line 37: nice use of doubles in these tests!

spec/weapon_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'weapon'
+
+describe Weapon do
+  let(:rock1) { double('rock1', type: :rock) }
```
siellsiell, line 4: It's confusing to have both `rock1` and `rock`. What's the difference between them? Also, if `rock1` is supposed to represent a `Weapon`, you shouldn't be doubling it because you are trying to test the `Weapon` class in this file.

views/choose2.erb
```diff
@@ -0,0 +1,31 @@
+<head>
```
siellsiell, line 1: This file is practically the same as `choose.erb`.  You could try and consolidate them into a single file OR find a way to share more code between them?

Sinatra's `layout.erb` feature can help to make your template code more DRY: https://learn.co/lessons/sinatra-yield-readme. This approach also saves you from having to repeat all the HTML boilerplate (e.g. the stuff in`<head>`) in every one of your template files.

views/lose.erb
```diff
@@ -0,0 +1,32 @@
...
+<p style="text-align:center">&nbsp;</p>
+<p style="text-align:center">
+<span style="font-size:28px">
+  <span style="font-family:Arial,Helvetica,sans-serif"><%= @game.player.name %> chose <%= @game.player_choice.type %>
+  <br>
+  <% if @game.multiplayer == true %>
+    <%= @game.player2.name %> chose <%= @game.player2_choice.type %>
+    <% else %>
+    Computer chose <%= @game.computer.type %>
+  <% end %>
```
siellsiell, line 16: Similarly to my comment about the RPS class: it would be good to find a way so that the template file doesn't need to know whether the opponent was a computer or a person. It could just receive two player names. This might come naturally if you find a way to refactor your RPS and Player classes as I suggested above.

siellsiell (commented):
Hi Christine,

You got really far, nice! Your website is also very nice to use -- liked your gif!

You found a good way to push logic down into your model and you tried to separate it into different classes. You could improve on that separation even more. The comments I left aim to steer you in that direction. A secondary improvement is to think about your naming of endpoints and methods a little more so that they are more descriptive. 

Please feel free to ask follow up questions!

PR #1832, opened 2021-02-22
Unfinished RPS Challenge

Laura-Keen:


charlierdm:
Code Review RPS:

- Good planning stages, the sequence diagram and planning notes are nice inclusions. 
- A README would be nice to know how it went or how to use the app. Did you create that instructions file, if so that should be the README as it's not immediately obvious but the README would appear right away, it's a very clear instructions!
- Testing looks well written and thorough, it's good you've separated them into multiple files and that included the web helper! Very clear to read through.  
- The app.rb file is clear(as is the rest of the code), it looks like everything is routed correctly. It would be nice to see the game logic implemented to see it all come together, you mentioned time management though so I'm sure you're aware of this.  Even without the game logic in there, you can see that you knew how it would work overall with your page routing. Overall it looks like a good understanding of the week3 concepts, good work on it!

PR #1834, opened 2021-02-22
Weekend Challenge Rock Paper Scissors

NicolasBermejo:
Only managed to get throught the singleplayer option as I struggled with the testing. mostly looking for feedback as I know my code needs refactoring...


app.rb
```diff
@@ -0,0 +1,44 @@
...
+class Rps < Sinatra::Base
+attr_reader :player2_choice
+enable :sessions
+  get '/' do
+    erb(:index)
+  end
+
+  post '/names' do
+    session[:player1] = Player.new(params[:player1])
+    session[:bot] = Player.new(AI_NAME)
```
siellsiell, line 16: Is it necessary to put this information in the session? The server will always know the bot's name since it's a constant. We put things into a session if the server would otherwise forget about them between requests.

lib/game.rb
```diff
@@ -0,0 +1,24 @@
...
+      result = "Paper covers rock, you win!"
+    end
+  elsif choice1 == "scissors"
+    if choice2 == "rock"
+      result = "Rock smashes scissors, you lose!"
+    elsif choice2 == "paper"
+      result = "Scissors cut paper, you win!"
+    end
+  end
+  return result
```
siellsiell, line 23: Can you think of a data structure that would allow you to avoid having to use so many if-statements to determine the outcome? 

Hint: you need a data structure that allows you to map a weapon to the weapon it would beat and the appropriate message to display.

spec/game_spec.rb
```diff
@@ -0,0 +1,11 @@
+require './lib/game.rb'
+require './app.rb'
+
+describe Rps do
+  it 'shows the game results' do
+    player1 = instance_double('Player', name: 'Nico', move: 'rock' )
+    bot = instance_double('Bot', name: 'Bot', move: 'rock')
+  @results = results(player1.move,bot.move)
```
siellsiell, line 8: There is no need to create an instance variable here (you don't need the `@`)

spec/game_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+  expect(@results).to eq("You tied!")
+  end
```
siellsiell, line 10: I'd expect more tests, one for every possible combination of moves that could occur

spec/game_spec.rb
```diff
@@ -0,0 +1,11 @@
+require './lib/game.rb'
+require './app.rb'
+
+describe Rps do
+  it 'shows the game results' do
+    player1 = instance_double('Player', name: 'Nico', move: 'rock' )
+    bot = instance_double('Bot', name: 'Bot', move: 'rock')
```
siellsiell, line 7:
```suggestion
    bot = instance_double('Player', name: 'Bot', move: 'rock')
```
As far as I can see you have no class called `Bot`. In an `instance_double` the string you pass should be the name of the class you're doubling

spec/features/choose_an_action_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    expect(page).to have_content "You tied!"
+  end
+
+  scenario 'it shows Player losing' do
+    srand(1)
+    sign_in_and_play
+    make_a_choice
+    expect(page).to have_content "Paper covers rock, you lose!"
+  end
+end
```
siellsiell, line 44: Nice feature tests!

To increase your test coverage, it would be good to test that inputs other than "rock" work too and that invalid inputs are handled.

lib/Player.rb
```diff
@@ -0,0 +1,10 @@
+class Player
+attr_reader :name, :move
+  def initialize(args)
+    @name = args
+  end
+
+  def choice(args)
```
siellsiell, line 7: The name makes it sound like it's going to return a choice but actually it's recording a choice. Using `args` as a parameter name is also not very descriptive (here and above).

```suggestion
  def record_choice(choice)
```

app.rb
```diff
@@ -0,0 +1,44 @@
...
+    session[:player1].choice(params[:choice1])
+    session[:bot].choice(AI_MOVE.sample)
+    redirect '/results'
+  end
+
+  get '/results'do
+    @player1 =session[:player1].name
+    @bot = session[:bot].name
+    @player1_choice = session[:player1].move
+    @player2_choice = session[:bot].move
```
siellsiell, line 38: It would be nice to be more consistent with your naming -- either call the weapon the user picks a "choice" or a "move" but not both :)

views/result.erb
```diff
@@ -0,0 +1,10 @@
+<h3> You chose <%= @player1_choice %> </h3>
+<h3> The Bot chose <%= @player2_choice %> </h3>
+<h2> <%= @result %> </h2> <br>
+
+<h4> Play Again? </h4>
+<h4> Select your move </h4>
+<form action="/choices" method="post">
+    <input type="text" placeholder="rock,paper,or scissors?" name="choice1">
```
siellsiell, line 8: Allowing the user a free-form text field to enter their moves means that you need to handle the possibility that they enter an invalid move, which you don't currently. You could prevent that by creating buttons for each option.

siellsiell (commented):
Hi Nicolas!

Thanks for sending your code for review! You've nicely avoided putting too much logic into your views by handling all of the game logic on the backend.

Regarding tests you could improve by testing a wider variety of cases to cover everything that could happen in the game, including edge cases. I've also left some suggestions as to how you could simplify your code.

Let me know if you have questions by replying on this PR.

PR #1835, opened 2021-02-22
RPS-Challenge

raf-swiderski:


lib/game.rb
```diff
@@ -0,0 +1,30 @@
+class Game
+    attr_reader :player_choice, :cpu_choice
+    def initialize(choice)
+        @player_choice = choice
+        @cpu_choice = ['Rock', 'Paper', 'Scissors'].sample
+    end
+
+    def cpu_choice
```
rhc07, line 8: May not need to call the cpu method explicitly considering you have an attribute reader and an initialize method? Good way of keeping your code DRY

PR #1837, opened 2021-02-22
Solution for Week 3 Challenge

peter-james-allen:
Complete
Covers all user stories
Added lizard and spock

Unfinished
No multiplayer, but I don't think it would be too hard to implement

To Improve
Need to look at class instances
I'm not sure if the game class is too large and needs splitting into multiple classes


app.rb
```diff
@@ -0,0 +1,75 @@
+require 'sinatra/base'
+require './lib/game'
+require './lib/player'
+require './lib/random_pick'
+
+class RPS < Sinatra::Base
```
sandyMax974, line 6: Silly question... but is that ok to put a class name in full capital like a constant?

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def set_game_type(type, random_class)
+    @type = type
+    type == 'RPS' ? @rand = random_class.new(CHOICES[0..2]) : @rand = random_class.new(CHOICES)
+  end
+
+  def won
+    @score1 += 1
+    @result = :win
```
sandyMax974, line 54: That's really cool that you added a score count!

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+    @pick1 == @pick2
+  end
+
+  def win?
+    TO_WIN[@pick1].include?(@pick2)
+  end
+
+  def lose?
+    !draw? && !win?
+  end
```
sandyMax974, line 76: Very good separation of responsibilities - very good SRP

app.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  post '/lizard' do
+    session[:game].turn('Lizard')
+    redirect '/result'
+  end
+
+  post '/spock' do
+    session[:game].turn('Spock')
+    redirect '/result'
```
sandyMax974, line 71: I like your approach on creating separate routes for each player choice.

spec/game_spec.rb
```diff
@@ -0,0 +1,76 @@
+require 'game'
+
+describe Game do
+
+  let(:player_class) { double :player_class, new: player }
+  let(:player) { double :player, name: "Pete", score: 0, won: true } 
+  let(:random_class) { double :random_class, new: random }
+  let(:random) { double :random, pick: 'Rock' }
+  let(:subject) { Game.new(player, "RPS", "The computer", player_class, random_class) }
```
sandyMax974, line 9: Nice use of double.

spec/web_helpers.rb
```diff
@@ -0,0 +1,12 @@
...
+  fill_in('name', with: 'Pete')
+  click_button('Play')
+end
+
+def sign_in_and_play_rps
+  visit '/'
+  fill_in('name', with: 'Pete')
+  click_button('Play')
+  click_button('rps')
+end
```
sandyMax974, line 12: Good implementation DRY

sandyMax974 (commented):
Code easy to read, each responsibility is well defined. I like how you've user a hash to manage the choices and results. Class responsibility break down is great, you move the `RandomPick` which is a great choice I think.
Really great effort on the CSS and presentation - I love it! 
Good use of the session. 
I would be curious to see how you would implement a 2 players mode with the current design.

app.rb
```diff
@@ -0,0 +1,75 @@
...
+    redirect '/type'
+  end
+
+  get '/type' do
+    @name = session[:name]
+    erb :type
+  end
+
+  post '/rps' do
+    session[:game] = Game.new(session[:name], 'RPS')
```
tatsiana-makers, line 29: Rather than use sessions to store the Game instance, you could apply the singleton pattern which is shown in this step of the Battle challenge: https://github.com/makersacademy/course/blob/master/intro_to_the_web/killing_the_global_variable.md

lib/game.rb
```diff
@@ -0,0 +1,82 @@
+class Game
+
+  CHOICES = ["Rock", "Paper", "Scissors", "Lizard", "Spock"]
+  R, P, S, L, SP = CHOICES
+  TO_WIN = { R => [S, L], P => [R, SP], S => [P, L], L => [P, SP], SP => [R, S] }
+
+  attr_reader :result, :type
+
+  def initialize(player1, type, player2 = "The computer", player_class = Player, random_class = RandomPick)
```
tatsiana-makers, line 9: Nice use of dependency injection!

tatsiana-makers, line 9: Perhaps calling the input parameters `player1_name` and `player2_name` would be more descriptive?

lib/game.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def set_scores
+    @score1 = 0
+    @score2 = 0
+  end
+
+  def set_game_type(type, random_class)
+    @type = type
+    type == 'RPS' ? @rand = random_class.new(CHOICES[0..2]) : @rand = random_class.new(CHOICES)
```
tatsiana-makers, line 49: It's common to have lines like these inside of the body of the `initialize` method and the methods aren't called from elsewhere so I'm wondering what your motivation was for extracting the logic into these methods.

I wonder if putting the logic back into the body of the `initialize` method would make the code slightly easier to read because a reader wouldn't have to jump to lower down in the file to see the logic

lib/game.rb
```diff
@@ -0,0 +1,82 @@
+class Game
+
+  CHOICES = ["Rock", "Paper", "Scissors", "Lizard", "Spock"]
+  R, P, S, L, SP = CHOICES
+  TO_WIN = { R => [S, L], P => [R, SP], S => [P, L], L => [P, SP], SP => [R, S] }
```
tatsiana-makers, line 5: This variable name is nice and clear ✨

lib/random_pick.rb
```diff
@@ -0,0 +1,11 @@
+class RandomPick
```
tatsiana-makers, line 1: It looks like this class's purpose at the moment is to wrap the Ruby array's `sample` method, which doesn't feel like sufficient responsibility currently to me to balance out the additional complexity of another class.

I wonder if converting this class into a `Computer` class, which stores the possible moves in a constant as well as returns the move that the computer chooses might be a nice way forward, as it would remove some responsibility from the Game class while making this class responsible for more things

spec/features/game_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  scenario 'takes players move - scissors' do
+    sign_in_and_play_rps
+    click_button "scissors"
+    expect(page).to have_content "Pete chose Scissors"
+  end
+
+  scenario 'shows computers move' do
+    sign_in_and_play_rps
+    click_button "rock"
+    expect(page).to have_content "The computer chose "
```
tatsiana-makers, line 24: Can you make use of `srand` here to make the assertion validate that the page actually shows the user's choice: https://gist.github.com/Kotauror/8f4a9913c53dd479df91909f306e6731

spec/game_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    context 'and you win' do
+
+      it 'should give a winning result' do
+        subject.turn("Rock", "Scissors")
+        expect(subject.result).to eq :win
+      end
+
+      it 'should increase the score for player 1' do
+        expect { subject.turn("Rock", "Scissors") }.to change { subject.stats[:score1] }.by 1
```
tatsiana-makers, line 39: We could also check that player 2's score isn't incremented

tatsiana-makers (commented):
Thanks for asking for a review! I've left some comments. Let me know if you have any follow-up questions, I'd be happy to chat further 🙂

PR #1840, opened 2021-02-22
not finished

joneslee111:


models/computer.rb
```diff
@@ -0,0 +1,8 @@
+class Computer
```
a-nawrocka, line 1: I like the idea of creating  computer class

spec/game_spec.rb
```diff
@@ -0,0 +1,12 @@
+require './models/game.rb'
+
+describe Game do
+  subject(:game) { described_class.new(options) }
+  let(:options) { {"player_name" => "Lee", "player_move" => :rock, "ai_move" => :scissors} }
```
a-nawrocka, line 5: nice let syntax

PR #1846, opened 2021-03-21
RPS Challenge!

louballantyne:


app.rb
```diff
@@ -0,0 +1,49 @@
...
+      Game.stored_game.play(params[:choose_turn], params[:choose_turn2])
+    else
+      Game.stored_game.play(params[:choose_turn])
+    end
+    redirect '/play'
+  end
+
+  run! if app_file == $0
+
+end
```
Yates101, line 49: Nice skinny controller, good use of session

lizard.rb
```diff
@@ -0,0 +1,21 @@
...
+  def explanation(player2turn)
+    case player2turn
+    when "Rock" then @explanation = "Rock crushes Lizard!"
+    when "Paper" then @explanation = "Lizard eats Paper!"
+    when "Scissors" then @explanation = "Scissors cut Lizard's head off!"
+    when "Spock" then @explanation = "Lizard poisons Spock!"
+    when "Lizard" then @explanation = "Great minds think alike!"
+    end
+  end
+end
```
Yates101, line 21: Interesting use of creating classes for each move for the logic. Perhaps this becomes a lot longer than it needs to be and an array of the answers would perhaps be better? Though it does make each outcome assured and I assume that becomes more complex when you start lizzard spocking

player.rb
```diff
@@ -0,0 +1,8 @@
+class Player
+  attr_accessor :name, :turn, :score
```
Yates101, line 2: I can't see the score used elsewhere? perhaps it is left over from the original game?

game.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def self.stored_game
+    @@this_game
+  end
+
+  def create_object(choice1)
+    case choice1
+    when "Rock" then @player1.turn = Rock.new
+    when "Paper" then @player1.turn = Rock.new
```
Yates101: Should these be the different move classes?

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,24 @@
+require_relative "../../app"
```
Yates101, line 1: Nice range of web helpers, good for dry tests

spec/units/game_spec.rb
```diff
@@ -0,0 +1,41 @@
+require_relative '../../game'
+
+describe Game do
+  describe '#new_game' do
+    it 'sets up a game with two player objects' do
+      expect(Game).to respond_to(:new_game).with(4).arguments
+    end
+    it 'allows the game to be started with one player' do
+      expect(Game).to respond_to(:new_game).with(3).arguments
+    end
```
Yates101, line 10: Nice, clear effective tests

spec/units/lizard_spec.rb
```diff
@@ -0,0 +1,38 @@
+require_relative '../../lizard'
+
```
Yates101, line 2: Good exhaustive unit testing

Yates101 (commented):
Really like your layout and execution of the different logic. While it is a little lengthy, it's very clear and thoroughly tested! On the feature front, the fact that the game was creating all Rock instances (which you caught and fixed), but the tests were still all passing, might suggest that a more extensive use of mocking the different moves might expose that?
Either way, it looks great, it's fun, functional and its clear a lot of work has gone in to it, great job!

PR #1847, opened 2021-03-21
RPS - human vs computer

cspoppuppy:
Web interface with functions: register name, play, display result


lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def player_choice(choice)
+    @player1.make_choice(choice)
+    @player2.make_random_choice
+  end
+
+  def result
+    rps = { "rock" => 0, "paper" => 1, "scissors" => 2 }
+    return "#{@player1.name} won!" if (rps[@player2.choice] + 1) % 3 == rps[@player1.choice]
```
eatruman, line 18: I LOVE this way of finding a winner, much better than a bit if elsif statement for all possibilities! Very clever!

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+    return "Draw!" if @player1.choice == @player2.choice
+
+    return "#{@player2.name} won!"
+  end
+
+  def self.create(player)
+    @game = Game.new(player)
+  end
```
eatruman, line 26: What does .create do? I've never seen this before!

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def self.instance
```
eatruman, line 28: I've also never seen this before, what is .instance doing here??

app.rb
```diff
@@ -0,0 +1,37 @@
...
+  post '/choose' do
+    @choice = params.keys[0].split(".")[0]
+    @game = Game.instance
+    @game.player_choice(@choice)
+    redirect to('/result')
+  end
+
+  get '/result' do
+    @game = Game.instance
+    @player1_choice_img = "/img/#{@game.player1.choice}.png"
```
eatruman, line 32: I like that you chose an image to represent what the player chose.

spec/features/register_name_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature 'Register name' do
+  scenario 'register name' do
+    sign_in_and_play
```
eatruman, line 3: Great use of a web helper.

spec/unit/game_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      allow(player).to receive(:choice).and_return("scissors")
+      allow(computer).to receive(:choice).and_return("paper")
+      expect(game.result).to eq "#{player.name} won!"
+    end
+
+    it 'player: scissors vs computer: scissors, draw' do
+      allow(player).to receive(:choice).and_return("scissors")
+      allow(computer).to receive(:choice).and_return("scissors")
+      expect(game.result).to eq "Draw!"
+    end
```
eatruman, line 77: This is super thorough but I don't know if you need a test for every single scenario based on the way you wrote your method. Since you wrote your method in a really clear/ simple way I think you probably could have done fewer tests but they still would have covered all possibilities.

eatruman (commented):
Sha, I'm sorry this isn't more helpful as far as ways to improve but I feel like you're done a really great job! I've learned a lot just by looking over your code so thank you for that!

PR #1850, opened 2021-03-21
rps challenge

fg24davies:
Please see README.md for my notes on this challenge. One glaring fault is the method #play in the game class being so long!
Would like to extract classes and implement multiplayer in the future and tidy up the design.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,33 @@
+require 'sinatra'
+require_relative './lib/player.rb' 
+require_relative './lib/game.rb'
+
+class RPS < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end 
+
+  post '/play-rps' do
+    session[:player] = params[:player]
+    @player_1 = session[:player]
+    erb :play
+  end
+
+  get '/play-rps' do
+    @player_1 = session[:player]
+    erb :play
+  end
+
+  post '/result' do
+    @player_1 = session[:player]
+    @game = Game.new(@player_1)
+    session[:player_1_move] = params[:move]
+    @player_1_move = session[:player_1_move]
+    @player_2_move = @game.computer
+    @result = @game.play(@player_1_move, @player_2_move)
+    erb :results
+  end 
+
+end
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -1,0 +1,37 @@
+require_relative './player.rb'
+
+class Game
+  
+  attr_reader :player_1, :player_2
+  
+  def initialize(player_1)
+    @player_1 = player_1
+    @player_2 = "Computer"
+  end   
+
+  def play(player_1_move, player_2_move)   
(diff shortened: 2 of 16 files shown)


app.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  get '/play-rps' do
+    @player_1 = session[:player]
+    erb :play
+  end
+
+  post '/result' do
+    @player_1 = session[:player]
+    @game = Game.new(@player_1)
+    session[:player_1_move] = params[:move]
```
TataSher, line 26: Seems to be too much logic going on that can be brought to Model layer

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+    elsif player_1_move == 'Paper' && player_2_move == 'Scissors' 
+      @player_2
+    elsif player_1_move == 'Scissors' && player_2_move == 'Rock' 
+      @player_2
+    elsif player_1_move == 'Scissors' && player_2_move == 'Paper' 
+      @player_1
+    end 
+  end 
+
+  # play method is way too long and complicated
```
TataSher, line 30: This could be done with a hash potentially. I think it is cool though that the return you get is the winner variable and this fits straight into the result

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def computer
+    moves = ["Rock", "Paper", "Scissors"]
```
TataSher, line 33: Could be another class as it does create a new player

lib/game.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def play(player_1_move, player_2_move)   
+    return nil if player_1_move == player_2_move
+
+    if player_1_move == 'Rock' && player_2_move == 'Scissors' 
+      @player_1
+    elsif player_1_move == 'Rock' && player_2_move == 'Paper' 
+      @player_2
+    elsif player_1_move == 'Paper' && player_2_move == 'Rock' 
+      @player_1
```
TataSher, line 20: The computing could be another class as well maybe that can contain rules to offload the tasks of the Game class.

views/index.erb
```diff
@@ -0,0 +1,18 @@
+<head>
+<style>
+body {background-color: lightblue;
+font-family: Helvetica, sans-serif;
```
TataSher, line 4: This looks very very nice!

TataSher (commented):
Your app looks very nice and also is very pleasant to play. I like that you did the buttons instead of the type in text for 'rock', 'paper' and 'scissors'. Maybe a bit of refactoring of Game class methods and generally splitting Game class will  help to make it more concise. I noticed also with the help of your comments that some of your test don't really test the actual behaviour, but it's a tricky one with all the random sampling, we should exchange some useful info that we find on this subject, cause I am also struggling :)
Well done! It's an awesome Rock Paper Scissors app!

PR #1851, opened 2021-03-21
Rock(/Paper/Scissors) & Roll

inbarvernia:


lib/player.rb
```diff
@@ -0,0 +1,16 @@
+class Player
```
ralphbartley, line 1: I think this is an elegant way to have the computer player but could possibly be extracted into a dedicated class of its own.

app.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  post '/outcome' do
+    @game = session[:game]
+    @game.player1.make_move(params[:player_move])
+    @game.player2.make_move
+    redirect '/outcome'
+  end
+
+  get '/outcome' do
+    @game = session[:game]
```
ralphbartley, line 32: I like how you're passing the game object into the session information rather than using the class instance variable for Game.  I wonder if one is more appropriate than the other?  Convention?

ralphbartley: Perhaps using the class instance variable doesn't work for multiple users accessing a site? Got me thinking!

lib/player.rb
ralphbartley: also love Deep Thought

spec/features/choices_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+feature 'Possible moves' do
```
ralphbartley, line 3: Good test.  Useful application of web_helpers, very readable.

spec/features/computer_move_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'spec_helper'
+
+feature 'Computer move' do
```
ralphbartley, line 3: Concise expectation given you have three options.  You could expect only one option if you specify an srand seed before the computer makes it move, thereby forcing the same answer each time but still testing the methods.

spec/units/game_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  describe '#initialize' do
+    it 'takes players as arguments and saves them as instance variables' do
+      game = described_class.new(rock_player, paper_player)
+      expect(game.player1).to eq(rock_player)
+      expect(game.player2).to eq(paper_player) 
+    end
+  end
+
+  describe '#winner' do
```
ralphbartley, line 16: Thorough!  This seems like the territory of input validation?  Could possibly have that in the rb code or app.rb?

views/outcome.erb
```diff
@@ -0,0 +1,13 @@
+<h2><%= @game.player1.name %> played <%= @game.player1.move.capitalize %></h2>
```
ralphbartley, line 1: short and sweet, nice and clean

spec/features/computer_move_spec.rb
ralphbartley: I see you do this in players!

lib/game.rb
```diff
@@ -0,0 +1,26 @@
+class Game
```
ralphbartley, line 1: I think this is a good set of responsibilities, perhaps Game could also aid move validation by containing a collection of allowed moves?

ralphbartley (commented):
Nice work!  I have confidence that running this code would give me a clean RPS experience :)  The testing is comprehensive and there's not much I can recommend!

PR #1853, opened 2021-03-22
incomplete

ralphbartley:
So I'm really happy with how thoroughly I planned with my README, but as you can see there I kind of lost focus on the minimum specifications of the program.  While I'm really happy with how my testing (much higher confidence with mocking) and code/structure turned out (used a subclass in a hopefully useful way! score!), most of the work has gone towards creating the backbone for the multiplayer bonus goal.  Since I actually outlined so much I think I'll be able to come back and finish this quickly with fresh eyes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,60 +1,184 @@
 # RPS Challenge
 
-Instructions
--------
+How to Play
+===========
+ - clone repo and bundle up!
+ - rackup and visit your shiny new sinatra ditty in the home route '/'
+ - enter your name and specify game parameters
+ - multiplayer can be played in two ways:
+  * choose on screen setting
+  * choose remote setting where you are able to provide remote access to the server -- load the site  in incognito windows or from separate clients so you can have separate cookies to test locally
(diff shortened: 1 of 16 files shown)


inbarvernia (commented):
Some interesting ideas for additional functionality and allowing for more complex game structures in here! I did struggle a little bit to understand how things fit together, in part because, as you wrote, you hadn't gotten around to writing the Game class yet, which I feel probably would have been the one to put things together a bit more. I think if you had a bit more time to write the rest of it, this would work, but as a reviewer, it made it a bit hard to understand.

I would mainly suggest, it would be good to start writing some feature tests before/alongside unit testing. From the perspective of a stranger looking at your code, it would make it easier to understand how you want this code to behave in a real-world application. It might also make things easier for you as you're writing the model code, for example (and I know I'm biased on this because this is how I approached it): if you feature test for giving the player a choice between rock, paper, scissors, you don't have to write any code to check that their choices are valid, because they are only able to make valid choices.

PR #1855, opened 2021-03-22
RPS challenge

taran314:


jackmcc08 (commented):
Overall, great!. 

Code is very clean, with easily understood methods and logic, as well as a clean flow in the router. 

There was some opportunity for refactoring, in particular applying polymorphism principles and a few other comments as per the detailed comments.

PR #1856, opened 2021-03-22
Rock, Paper, Scissors

dmgreaves:


app.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  post '/move' do
+    $rock = params[:ROCK]
+    $paper = params[:PAPER]
+    $scissors = params[:SCISSORS]
+    @player_move = Move.new.player_move
+    @computer_move = Move.new.computer_move
+    @win_lose = Move.new.win_lose
+    erb :move
```
StuBehan, line 30: I think you shouldn't direct to an erb from a post method so we need a get method and a `redirect '/result' `  here or something along those lines.

views/play.erb
```diff
@@ -0,0 +1,9 @@
+<h1> Player: <%=  @player %> </h1>
+
+<h2> Make your move... </h2>
+
+<form action="/move" method="post">
+<input type="submit" value="ROCK">
+<input type="submit" value="PAPER">
+<input type="submit" value="SCISSORS">
+<form>
```
StuBehan, line 9: this works, but I think the app.rb doesn't do its part properly.

lib/move.rb
```diff
@@ -0,0 +1,47 @@
...
+  elsif @player_move == "Paper" && @computer_move == "Paper"
+    @win_lose = "Draw!"
+  elsif @player_move == "Rock" && @computer_move == "Rock"
+    @win_lose = "Draw!"
+  elsif @player_move == "Scissors" && @computer_move == "Scissors"
+    @win_lose = "Draw!"
+  else
+    @win_lose = "You win!"
+  end
+end
```
StuBehan, line 45: this method is super long, perhaps there is a better way of looking at these possibilities? perhaps a 2d array to iterate over or something?

StuBehan (commented):
I think it's a solid base that with a bit of rejig and removal of the global variable and use of the class instance variable.

PR #1857, opened 2021-03-22
rps-challenge

willspencer16:
Basic completion of the user stories. Opportunity to improve code and interface.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,43 @@
+require 'sinatra/base'
+
+class RPS < Sinatra::Base
+	enable :sessions
+	set :sessions, true
+	set :session_secret, '7389575'
+
+	get '/' do
+		erb(:index)
+	end
+
+	post '/user' do
+		session[:user] = params[:user]
+		redirect to('/play')
+	end
+
+	get '/play' do
+		@user = session[:user]
+		erb :play
+	end
+
+	post '/rock' do
+		@user = session[:user]
+		erb :rock
+	end
+
+	post '/paper' do
+		@user = session[:user]
+		erb :paper
+	end
+
+	post '/scissors' do
+		@user = session[:user]
+		erb :scissors
+	end
+
+	post '/play' do
+		@user = session[:user]
+		erb :play
+	end
+
+	run! if app_file == $0
+end
--- a/app.rb
+++ b/app.rb
@@ -20,16 +20,22 @@ class RPS < Sinatra::Base
 	end
 
 	post '/rock' do
+		@opponent_choice = ["Rock", "Paper", "Scissors"].sample
+		@user_choice = "Rock"
 		@user = session[:user]
 		erb :rock
 	end
 
 	post '/paper' do
+		@opponent_choice = ["Rock", "Paper", "Scissors"].sample
+		@user_choice = "Paper"
 		@user = session[:user]
 		erb :paper
 	end
 
 	post '/scissors' do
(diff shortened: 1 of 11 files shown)


spec/features/app_spec.rb
```diff
@@ -47,5 +47,14 @@
     find_button('Rock').click
     expect(page).to satisfy {|page| page.has_content?("Rock") or page.has_content?("Paper") or page.has_content?("Scissors")}
```
Emmapr123, line 48: Nice test! I didnt know this syntax, but looks really good

views/play.erb
```diff
@@ -0,0 +1,14 @@
+
+<h1>Player: <%= @user %></h1>
+
+<form action ='/rock' method='post'>
```
Emmapr123, line 4: I think you could have put these three submit buttons in one form, that way they could have all gone to one action instead of all to a different one. If you do want to try that, I think you need a value and a name.

views/scissors.erb
```diff
@@ -3,6 +3,10 @@
 
 <h2> Scissors </h2>
 
+<h1>Opponent: Edward Scissor Hands</h1>
```
Emmapr123, line 6: Love the name

app.rb
```diff
@@ -0,0 +1,49 @@
...
+		redirect to('/play')
+	end
+
+	get '/play' do
+		@user = session[:user]
+		erb :play
+	end
+
+	post '/rock' do
+		@opponent_choice = ["Rock", "Paper", "Scissors"].sample
```
Emmapr123, line 23: I think you could have made a ruby file that would have done this, to keep things dry - but it's pretty cool that this works too without too many lines of code

views/scissors.erb
loushark: I'm looking forward to seeing more opponent names you come up with! Love this

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def sign_in_and_play
```
loushark, line 1: Now that you have this web helper, you'll be able to update your tests to use this, to help keep your test code DRY

views/paper.erb
```diff
@@ -0,0 +1,30 @@
...
+
+<h2><%= @opponent_choice %></h2>
+
+<% if @opponent_choice == "Paper" %>
+
+<h1>It's a draw!</h1>
+
+<% end %>
+
+<% if @opponent_choice == "Scissors" %>
```
loushark, line 16: I wonder if you could remove the empty lines here to make the blocks of code a little more readable? i.e. which lines belong together and which should be separated. You could use line breaks with `<br>` in areas of your code where you visually want there to be empty lines on the web app e.g between lines 4 and 6.

views/play.erb
loushark: agreed on merging the forms into one form block. again you can still use <br> to add the buttons to new lines. And later with CSS you can style them with separate IDs without them needing to be in a separate form block (but that's just a bonus thing to do and not required to meet the user story need)

loushark (commented):
Love how simple your code is to achieve the requirements of the user stories.

You could add a ruby file with a class or two that enables you to create instances of the players and game itself that allows you to to keep your controller code DRY, but also will make things easier for you to expand the game further (e.g. more players added, players that have win points, adding lizard spock etc). 

Really enjoy that your computer has a fun name! Was really fun to play it! Nice work, Will!

PR #1859, opened 2021-03-22
RPS 

TataSher:


app.rb
```diff
@@ -0,0 +1,29 @@
+require 'sinatra/base'
```
fg24davies, line 1: really nice, 'skinny' controller. 
I see you have enabled sessions but then not used them, was this for any particular reason?

lib/score.rb
```diff
@@ -0,0 +1,24 @@
+class Score
+  attr_reader :player_weapon, :opponent_weapon
+  WIN = {
```
fg24davies, line 3: This was a really cool way to do the outcome! I also like the extracted method of score from game. Makes game very clear to read.

lib/score.rb
```diff
@@ -0,0 +1,24 @@
...
+  scissors: :paper,
+  paper: :rock,
+  }
+  def initialize(player_weapon, opponent_weapon)
+    @player_weapon = player_weapon
+    @opponent_weapon = opponent_weapon
+    @game = { @player_weapon.to_sym => @opponent_weapon.to_sym }
+  end
+
+  def calculate
```
fg24davies, line 14: will remember this trick!

app.rb
```diff
@@ -0,0 +1,29 @@
...
+    erb :play
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post '/outcome' do
+    @game = Game.new_game(@player1_name, params[:player_weapon])
+    erb @game.outcome
```
fg24davies, line 27: this is interesting! I didn't know you could just write erb and then some ruby. I'd be intrigued to know how this works.

lib/game.rb
```diff
@@ -0,0 +1,24 @@
+require './lib/weapon.rb'
+require './lib/score.rb'
+class Game
+
+   def self.new_game(name, player_weapon)
```
fg24davies, line 5: this was instead of the global variable part of the afternoon challenge? It's cool to see it in use

lib/weapon.rb
```diff
@@ -0,0 +1,10 @@
+class Weapon
```
fg24davies, line 1: All the extracted classes make your code really clean and easy to understand what's what and the methods are short within the classes also

spec/features/select_weapon_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'select weapon' do
+  scenario 'user selects weapon' do
+    sign_in_and_play
+    fill_in "player weapon", :with => "rock"
```
fg24davies, line 4: I think this test is failing because the name of your text field is player_weapon so it cannot find "player weapon" - I've added a _ to see if it works!

fg24davies (commented):
The website works nicely, looks good! It is currently telling me I'm losing after I played rock and the computer played scissors though. I think you were aware of this, maybe we can discuss in pair progrmaming later.


I ran rspec and saw the 2 failures like you said and I also ran rubocop that had 7 offenses but only indentation issues  that can be cleaned up very easily so well done for that!

Re Rspec:
I changed a typo in the select_weapon_spec --> fill_in "player weapon" should have have been "player_weapon". All your tests now passed. I did this on my local repo, I can't do it on here. 

Overall, I really liked how you designed your app. It feels very clean and clear. The classes divided up the logic nicely and you were able to keep your controller skinny! Well done 🥇

PR #1860, opened 2021-03-22
Weekend challenge review submit

Yates101:
Given more time, I would have separated responsibility of the computer generated move in to a sub-class of Player, as well as creating a sub-class for human players to accommodate the multiplayer element.
Additionally I would have used a constant or an array to store the move values and then refer to that in the calculate_winner method, so as to write a shorter and more concise method.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,20 @@
+require 'sinatra/base'
+require_relative './lib/rps'
+
+class RpsApp < Sinatra::Base
+  set :session_secret, 'secret sessions be secret'
+  enable :sessions
+
+  before do
+    @game = RPS.instance
+  end
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/name' do
+    @game = RPS.create(params[:player_1])
+    redirect '/play'
+  end
+end
--- a/app.rb
+++ b/app.rb
@@ -13,8 +13,19 @@ class RpsApp < Sinatra::Base
     erb(:index)
   end
 
-  post '/name' do
-    @game = RPS.create(params[:player_1])
+  post '/names' do
+    @player_1 = Player.new(params[:player_1])
+    @game = RPS.create(@player_1)
     redirect '/play'
   end
+
+  get '/play' do
+    erb(:play)
+  end
+
+  get '/result' do
+    @result = @game.calculate_winner(params[:move], ["Rock", "Paper", "Scissors"].sample)
+    erb(:result)
+  end
+
 end
--- a/lib/player.rb
+++ b/lib/player.rb
@@ -1,0 +1,7 @@
+class Player
+  attr_reader :name
+
+  def initialize(name)
+    @name = name
+  end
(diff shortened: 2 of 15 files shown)


spec/feature/play_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+Capybara.app = RpsApp
+
+feature "/play" do
+  scenario "when players name entered, should serve play page" do
+    sign_in_and_play
+    expect(page).to have_button("Rock")
+    expect(page).to have_button("Paper")
+    expect(page).to have_button("Scissors")
+  end
+
```
louballantyne, line 13: Nice use of scenario - I need to look into the difference between this and describe/context

lib/rps.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+  
+  def self.instance
+    @game
+  end
+
+  def initialize(player)
+    @player = player
+  end
+
```
louballantyne, line 17: This is quite a long if statement. Perhaps you could refactor your code into further classes and methods.

lib/player.rb
```diff
@@ -0,0 +1,7 @@
+class Player
+  attr_reader :name
+
```
louballantyne, line 3: Perhaps the player class could have even greater responsibility (such as storing a turn) in order to extract methods from the RPS class

spec/unit/rps_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rps'
+
+describe RPS do
+  let(:player_1_mock) { double(Player, name: "Wendy") }
+
```
louballantyne, line 5: I like your use of a double to keep the testing to a single class.

views/index.erb
```diff
@@ -0,0 +1,16 @@
...
+  <head>
+    <meta charset="utf-8">
+    <title>Welcome to RPS</title>
+  </head>
+  <body>
+    <h1>Welcome to Rock, Paper, Scissors!</h1>
+    <form action="/names" method="post">
+      <label for="name">Player, enter your name below:</label><br><br>
+      <input type='text' id="name" name="player_1" placeholder="John Doe">
+      <input type='submit' value='Submit'>
```
louballantyne, line 13: Nice use of placeholder!

views/result.erb
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+
```
louballantyne, line 2: It would be really nice to see the computer's choice here too!

app.rb
```diff
@@ -0,0 +1,31 @@
+require 'sinatra/base'
+require_relative './lib/rps'
+
```
louballantyne, line 3: Nice, skinny controller!

louballantyne (commented):
Nice code! Love the use of buttons to choose user input. Would have been super helpful to have instructions on how to run it in the README.

PR #1862, opened 2021-04-18
Rock-Paper-Scissors with beautification.

mikejeuga:
It should be looking good.


spec/features/in_game_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+feature "when playing" do
+  scenario "the player should to see in both names" do
+    login_and_play
+    expect(page).to have_text("Mike vs MATRIX")
+  end
+  
+  scenario "the player should be able to click on rock" do
+    login_and_play
+    expect(page).to have_selector(:link_or_button, 'rock')
```
KaneG9, line 11: Could you be more specfic here eg just link or just button?

lib/engine.rb
```diff
@@ -0,0 +1,14 @@
+class Engine
+  attr_reader :name, :rmove
+  attr_accessor :points
```
KaneG9, line 3: change_points method rather than an accessor,  I think they mentioned somewhere using accessors are bad practices

spec/features/end_game_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "./spec/features/web_helpers"
+
+feature "when the game ends" do
+  scenario "the player can keep playing" do
+    login_and_play
+    click_button "rock"
+    expect(page).to have_content "Keep playing"
+    click_button "Keep playing"
```
KaneG9, line 8: Do you need this line?

spec/game_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "game"
+
+describe Game do
+  let(:michael) { double :player, name: "Michael", points: 0 }
+  let(:matrix) { double :engine, name: "MATRIX", points: 0 }
+
+  context "#initialize" do
+    it "should have initiated a player" do
+      allow(michael).to receive(:name).and_return("Michael")
```
KaneG9, line 9: defined in line 4?

PR #1864, opened 2021-04-18
Jess Brand RPS challenge

tremella:
See README for outline of project. Completed base level challenge (2 user stories), did not do bonus levels however.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,32 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/player'
+# require './lib/result'
+
+class RPS < Sinatra::Base
+  enable :sessions
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    # 'hello world'
+    erb :input_name
+  end
+
+  post '/name' do
+   session[:p1_name] = params[:p1_name]
+   redirect '/choose'
+  end
+
+  get '/choose' do
+    erb :choose
+  end
+
+  post '/result'do
+    session[:choice] = params[:choice]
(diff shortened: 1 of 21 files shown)


lib/judge.rb
```diff
@@ -0,0 +1,23 @@
+# session[:choice], session[:ai_choice]
+
+class Judge
```
ahmad047, line 3: good idea of using a judge class

app.rb
```diff
@@ -0,0 +1,35 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/player'
+require './lib/ai'
+require './lib/judge'
+
+class RPS < Sinatra::Base
+  enable :sessions
```
ahmad047, line 8: consider using class instance variables instead of sessions, you could automate the saving and loading game functionality, this will make the controller skinny

spec/features/choose_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    page.choose('scissors')
+    click_button ('play')
+    expect(page).to have_content('You chose scissors')
+  end
+  scenario 'solo player clicks rock' do
+    sign_in_and_play
+    page.choose('rock')
+    click_button ('play')
+    expect(page).to have_content('You chose rock')
+  end
```
ahmad047, line 20: I think since we have tested the feature in the first test so we could get rid of these two tests

ahmad047 (commented):
overall, a very well put together piece of software. Only thing I guess to recommend will be implementing class instance variables which will take a lot of functionality out of the controller back to the model.

PR #1873, opened 2021-04-18
RPS-Challenge

fizzy-fifs:


fizzy-fifs:
Completed the challenge. Upon reflection however, I realised that some of the feature tests were poorly written, as they attempt to test specific outcomes of randomness when they should have been testing for behaviours instead.

PR #1874, opened 2021-04-19
Incomplete RPS Solution

fwill22:
WIP - my solution is really incomplete. Have been blocked on implementing the underlying ruby logic. Failing unit test for play view dependent on this.


lib/computer.rb
```diff
@@ -0,0 +1,11 @@
+require_relative 'weapon'
```
fwill22, line 1: remove line

PR #1877, opened 2021-04-19
First build of the challenge

phileeep:


views/playgame.erb
```diff
@@ -0,0 +1,28 @@
...
+  <head>
+    <meta charset="utf-8">
+    <title>Rock Paper Scissors!!</title>
+    <link href="<%= url("/styles.css") %>" rel="stylesheet" type="text/css"/>
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Oswald&display=swap" rel="stylesheet">    
+  </head>
+
+  <body>
+    <div class = "wrapper">
```
YohannTisserand, line 11: setting up some divs is quite nice 👍

app.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  get "/" do
+    erb :index
+  end
+
+  post "/names" do
+    player1 = Player.new(params[:player1])
+    computer = Player.new("Computer")
+    $game = Game.new(player1, computer)
```
YohannTisserand, line 18: you need to kill the global variable here! it's evil 🦹‍♀️ 
And set up an instance variable of the game into the main file such as 
```
def self.create
@game = Game.new(params)
end
def self.instance
@game
end 
```

PR #1879, opened 2021-05-15
Beca's Solution (Apr 21 cohort)

BecaLParker:
Thanks for reviewing! 

Don't be shy - I am receptive to feedback.


README.md
```diff
@@ -1,21 +1,43 @@
-# RPS Challenge
+# BecaLParker's Solution to Rock Paper Scissor Challenge
```
fel1xp, line 1: Clear and concise readme.

app.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  get '/result' do
+    @player_weapon = session[:player_weapon]
+    @player_name = session[:player_name]
+    game = Game.new
+    @computer_weapon = Computer.new.select_random
+    @winner = game.calculate_outcome(@player_weapon, @computer_weapon)
+    erb :result
+  end
+
```
fel1xp, line 40: A very neat way of calculating the result!

lib/game.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
```
fel1xp, line 1: I like the way you have separated the game logic into its own file.

spec/features/weapon_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'capybara/rspec'
+require_relative '../../app'
+
+Capybara.app = RPS
+
+feature "player's weapons" do
+  it 'player presented with weapon choices' do
+    sign_in_and_play
```
fel1xp, line 10: Some good re-factoring here

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
```
fel1xp, line 1: Covering all bases too, nice one

README.md
fel1xp, line 1: Overall it is a very well written and concise piece of code. You've taken the time to re-factor and your approach is clear. The only improvement I would suggest is having the game declare who won, as opposed to 'Scissors won'.

BecaLParker: Agreed! I will write some code that checks who played the winning weapon and use that output in the /result view instead.

app.rb
BecaLParker: Thanks I can't take full credit for that approach... 

After a lot of head-scratching, I referred to the code review rubric for this challenge and saw the hash showing 'weapon => weapon it beats' as key => value pairs and that was able to move forward with that.

BecaLParker:
I think I was failing Travis because of a bug with :draw not returning successfully. In fixing this, I had to add arguments to the Game.new method, which broke my unit testing. I added a double in game_spec file to get back to all tests passing.

PR #1886, opened 2021-05-16
week 3 end of unit challenge submission

rdupplaw:


lib/rps_game.rb
```diff
@@ -0,0 +1,46 @@
...
+    outcome = outcome()
+    opponent = opponent(choice, outcome)
+    { outcome: outcome, opponent: opponent }
+  end
+
+  private
+
+  def outcome
+    ['win', 'loss', 'draw'].sample
+  end
```
Wh3g, line 12: never would have thought to do it that way, it makes sense I guess and looks simple though, and if it works it works

views/layout.erb
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Rock Paper Scissors</title>
+  </head>
+  <body>
+    <%= yield %>
```
Wh3g, line 7: what does `yield` do?

lib/rps_game.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def wins_against(choice)
+    case choice
+    when 'rock'
+      'paper'
+    when 'paper'
+      'scissors'
+    when 'scissors'
+      'rock'
+    end
```
Wh3g, line 44: I like the way you've done this, very clean and minimises length of method as well

PR #1888, opened 2021-05-16
Dev's version completed

Dev-ops-true:


Dev-ops-true:
Made changes to pass Travis.yml tests

PR #1892, opened 2021-05-17
RPS challenge pull request

Wh3g:


README.md
```diff
@@ -1,85 +1,5 @@
...
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
-  SimpleCov::Formatter::Console,
-  # Want a nice code coverage website? Uncomment this next line!
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
-
-You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
+For this challenge I used the knowledge gained from the previous week learning to build web apps, the basic format for my solution to this problem was fairly simple, for the first user story, I had to save the user's name, which I did by building an HTML form, and saving the input as an instance variable of a Player class object, this Player object is in turn saved as an instance variable of a Game class object, then I took the user's choice of Rock, Paper or Scissors, saving that as an instance variable as well, and ran it, as well as a randomly generated move through a method to decide the outcome of the game, finally announcing the winner
```
rdupplaw, line 5: You could also include instructions on how to download and run the app here.

spec/game_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      expect(game.outcome).to eq game.comp_win
+    end
+
+    it "returns draw" do
+      srand(2)
+      p game.comp_move
+      expect(game.outcome).to eq game.draw
+    end
+  end
+end
```
rdupplaw, line 38: You might want to test every possible outcome here e.g. every combination of choices. E.g. it looks like there is no test that scissors beats paper.

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def comp_win
+    "The computer wins!!"
+  end
+
+  def draw
+    "It's a draw"
+  end
+end
```
rdupplaw, line 43: These presentation strings should not be handled by the model. The model should just return e.g. :win, :loss, :draw, then the view layer should control how this data is presented.

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+    when @player.move == @comp_move
+      draw
+    when @player.move == "Scissors" && @comp_move == "Rock"
+      comp_win
+    when @player.move == "Paper" && @comp_move == "Scissors"
+      comp_win
+    when @player.move == "Rock" && @comp_move == "Paper"
+      comp_win
+    end
+  end
```
rdupplaw, line 30: I did this in a similar way, but check the rubric for a more elegant way of doing this with a hash and a Weapon#beats? method.

lib/game.rb
```diff
@@ -0,0 +1,43 @@
...
+  RPS = ["Rock", "Paper", "Scissors"]
+  attr_reader :player, :comp_move
+  def initialize(player)
+    @player = player
+    @comp_move = comp_choice
+  end
+
+  def comp_choice
+    RPS[rand(3)]
+  end
```
rdupplaw, line 11: It might be better for the computer to be a separate class. Then the Game class could take two players, which might be real or computers, and compare their choices.

app.rb
```diff
@@ -0,0 +1,40 @@
...
+    register Sinatra::Reloader 
+  end
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/name' do
+    player = Player.new(params[:player_name])
+    $game = Game.new(player)
```
rdupplaw, line 19: global variables should be avoided

lib/game.rb
```diff
@@ -0,0 +1,43 @@
+class Game
+  RPS = ["Rock", "Paper", "Scissors"]
```
rdupplaw, line 2: storing this in a constant is good

app.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  get '/play' do
+    @game = $game
+    erb(:play)
+  end
+
+  post '/choice' do
+    @game = $game
+    @game.player.save_move(params[:move])
+    redirect '/outcome'
```
rdupplaw, line 31: good splitting up post and get requests

lib/player.rb
```diff
@@ -0,0 +1,10 @@
+class Player
+  attr_reader :name, :move
+  def initialize(name)
+    @name = name
+  end
+
+  def save_move(move)
+    @move = move
```
rdupplaw, line 8: you could add a guard condition here if the given move is not in Game::RPS array

PR #1894, opened 2021-05-17
Stuart McArdle - RPS Challenge

StuMcardle:


app.rb
```diff
@@ -0,0 +1,43 @@
...
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    $player_1 = Player.new(params[:player_1_name])
```
GunelMC, line 18: Could you store this in an instance variable and then session? Like 
@player1 = Player.new(params[:player_1_name]) 
session[:player_1_name] = @player1  
I think then you can use this across multiple requests within the same session.

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,9 @@
+def sign_in_and_play
+  visit('/')
+  fill_in :player_1_name, with: 'Charlotte'
+  click_button 'Submit'
+end
+
+def opp
```
GunelMC, line 7: great idea to use helpers and to put the choice pick in here!

spec/spec_helper.rb
```diff
@@ -1,6 +1,16 @@
+ENV['RACK_ENV'] = 'test'
```
GunelMC, line 1: I still need to figure out what this is for, but well done for figuring it out!

views/scissors.erb
```diff
@@ -0,0 +1 @@
+<h1><%= @player_1_name %> chose scissors and the Computer chose <%= choice = ["rock", "paper", "scissors"].sample %></h1>
```
GunelMC, line 1: great! Overall, I really like how readable and clean your code is. One thing to consider would be creating a method in your model (game or player.rb) to evaluate the result and put it in your results erb to tell the user the result of the game.  Overall, good job! ⭐

PR #1896, opened 2021-05-17
Rock, Paper, Scissors!

fel1xp:


README.md
```diff
@@ -1,4 +1,4 @@
-# RPS Challenge
+n# RPS Challenge
 
 Instructions
 -------
```
BecaLParker, line 4: Ideally it would be good to have an update readme for your code with

- what your code does
- steps to install it (from a user-perspective)
- any special instructions about how to interact with it

app.rb
```diff
@@ -0,0 +1,51 @@
...
+class RPS < Sinatra::Base
+  enable :sessions
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    "Hello! Welcome to Rock, Paper, Scissors!"
+    "Please enter your name below to begin"
```
BecaLParker, line 15: I love that the instructions for user interaction are displayed in the web interface

app.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  get '/play' do 
+    @player_1 = session[:player_1]
+    erb :play
+  end
+
+  get '/rock' do 
+    @computer_move = computer_move
+    #"You played rock, and the computer played #{computer_move}!"
+    #"You Win" if computer_move == 'scissors'
```
BecaLParker, line 32: I see where you're going with this :) 

For separation of concerns, you might find it helpful to have another ruby file in ./lib that is responsible for the game logic (rock beats scissors etc). 

Once that's in place, you can have app.rb file call that method and pass the output of the game logic to your view file.

app.rb
```diff
@@ -0,0 +1,51 @@
...
+    :rock
+  end
+
+  get '/paper' do
+    @computer_move = computer_move
+    "You played paper, and the computer played #{computer_move}!"
```
BecaLParker, line 38: Good that you stored the computer move in a variable. 

And the string interpolation works well in the web interface when I tried it out :)

For separation of concerns and MVC principles, consider moving and sort of stringyness to the erb files in view.

computer_move.rb
```diff
@@ -0,0 +1,3 @@
+def computer_move
+['rock', 'paper', 'scissors'].sample
```
BecaLParker, line 2: Elegant implementation of randomness :)

spec/features/computer_move_spec.rb
```diff
@@ -0,0 +1,6 @@
+
+describe 'computer_move' do 
+  it 'returns a random move' do
```
BecaLParker, line 3: The test you've written works and is useful as a test of 'does the computer move equal either rock, paper or scissors?'

Consider renaming this test accordingly and implementing another test if you want an approximate assertion of randomness (maybe calling the method a bunch of times, pushing the result to an array, and checking each of the three options is returned at least once in the array?)

views/play.erb
```diff
@@ -0,0 +1,15 @@
...
+  <button type"button">Rock!</button>
+</form>
+
+<form action="/paper" method="get">
+  <button type"button">Paper!</button>
+</form>
+
+<form action="/scissors" method="get">
+  <button type"button">Scissors!</button>
+</form>
```
BecaLParker, line 15: I love that you got your buttons in a vertical layout :)

app.rb
BecaLParker: Now I see you began implementing that with your winner_rock file :)

Did you try calling that method in app.rb to return the outcome and pass it to your view file?

BecaLParker (commented):
See specific comments throughout but in summary:

The web interface worked well when I cloned your repo and rackup'ed :) Nice job on the messages you give your user. 

Your next step could be to extract all game logic to a lib file with that as it's single responsibility. Test that without needing capybara, then call the method/s from your app.rb and proceed that way.

PR #1905, opened 2021-06-14
Chuka's RPS Challenge

coo990:
First user story completed, 100% coverage and passes RSpec feature test. 2nd user story not completed. I created multiple paths but struggled to combine the computer logic through links, the idea was the 'Rock, Paper, Scissors' links to be the players input and go against the computers randomisation, and another method for the outcome to be displayed whether the player won or lost. Tried to make it simpler. Screenshots in README.


lib/game.rb
```diff
@@ -0,0 +1,23 @@
...
+    @random = ["Rock", "Paper", "Scissors"].sample
+  end
+
+  def player_score
+    'Rock'
+  end
+
+  def result1
+    if player_score && @random == "Paper"
+      puts "You won!"
```
YinnyF, line 17: Can remove puts and replace with return so that your controller can work with it

spec/game_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'game'
+
+describe Game do
+  let(:game) { double :game }
+  it 'game produces random option' do
+    allow_any_instance_of(Array).to receive(:sample).and_return('Paper')
```
YinnyF, line 6: It's very smart to use a stub this way!

views/welcome.erb
```diff
@@ -0,0 +1,5 @@
+<h2>
+  Welcome, <%= @name %>!
+</h2>
+
+<a href="/win">Rock</a>  <a href="/win">Paper</a> <a href="/lose">Scissors</a>
```
YinnyF, line 5: I see where you are going with this, perhaps you could create another page in the controller to bring together the model logic and then redirect to a certain view depending on the outcome.

YinnyF (commented):
Overall really good effort on the challenge. 
The repo has a nice structure following MVC pattern. 
The code is neat easy to understand. 
I see where you are going with the direction of the challenge and have offered some comments to help you get there!

PR #1908, opened 2021-06-14
rps-challenge

Ashley-Slaney:


app.rb
```diff
@@ -0,0 +1,39 @@
+require 'sinatra/base'
```
craigea92, line 1: Nice and straight forward to read code. I can also see the elements you planned to implement later.

PR #1918, opened 2021-07-18
Luke - RPS Challenge

lushcodes:


isobelpepp:
Luke this is lovely code - super readable and very impressed with how you have dealt with separation of concernsl! The only two suggestions I would make (which are both very small) is to put the computer choices in a constant and in your outcome spec whether you could use a before ... do hook so you don't have to repeat yourself for every test.

PR #1922, opened 2021-07-19
Matt Firmston rps-challenge

YoFirmy:


app.rb
```diff
@@ -0,0 +1,41 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/player'
+require './lib/npc'
+require './lib/result'
+
```
pav0107, line 6: Nice and clean code. Very easy to understand.

lib/result.rb
```diff
@@ -0,0 +1,27 @@
...
+    elsif outcome_calculation == 1
+      'You win!'
+    else 
+      'You lose...'
+    end
+  end
+
+  private
+
+  def outcome_calculation
```
pav0107, line 24: There may be other ways of doing this that I'd personally find more intuitive, but it's a great bit of problem-solving nevertheless.

spec/result_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'result'
+
```
pav0107, line 2: Good use of doubles.

PR #1926, opened 2021-08-15
All done, looks good, feels good, done good

necsanuradu:
I've written the tests before the code for the whole exercise, it was hard


app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+  before do
+    @game = Game.instance
+  end
+
+  get '/' do
+    erb(:index)
+  end
+
+  post "/game" do
+    session[:name] = params[:name].downcase.split(" ").map(&:capitalize).join(" ")
```
leoht, line 21: This piece of logic should probably be *at least* within some method, so it's clear when reading it what it does/means:
```ruby
session[:name] = capitalize_name_parts(params[:name])
```

app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+    when :draw
+        redirect("/results_draw?wact=" << @game.player1.option.to_s << "&lact=" << @game.player2.option.to_s)
+    else
+        @error = true
+        redirect("/game")
+    end
+  end
+
+  get "/results_won" do
+    params[:name].nil? ? (@name = session[:name]) : (@name = params[:name])
```
leoht, line 60: Once again here, maybe it'd be good to move this into a controller method, something like `set_name_from_params_or_session`, since it's used for the three different endpoints. You could actually also have a single endpoint, and have a different behaviour with a param `?result=win` for example

app/views/results_draw.erb
```diff
@@ -0,0 +1,17 @@
+<hr style="border-top:rgba(0, 0, 0, .1)">
```
leoht, line 1: Likewise here, since the views are mostly the same maybe it could be one view and then some conditionals to show different messages/parts depending on the result?

It also might be better to put any CSS rules inside stylesheets .css files rather than inlining them with `style` attributes, as it improves readability and brings more flexibility into applying CSS rules to elements

lib/game.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def results_setup
+    win_or_lose(@player1.option, @player2.option)
+  end
+
+  def win_or_lose(choice1, choice2)
+    return :error unless [:rock, :paper, :scissors].include?(choice1)
+    return :error unless [:rock, :paper, :scissors].include?(choice2)
+    option = choice1.to_s[0] << choice2.to_s[0]
+    @winners.include?(option) ? ((@winners.index(option) % 2 == 0) ? :win : :lose) : :draw
```
leoht, line 20: Smart one! To make it even better, isolate this logic into a single private method in this class, that would also define the `@winners` array (it's used only for this logic, and not needed to store some "state" on the class):

```ruby
def get_win_or_lose_result(choice1, choice2)
  winners = ['', '', 'sp', 'ps', 'pr', 'rp', 'rs', 'sr']
  option = choice1.to_s[0] << choice2.to_s[0]
  winners.include?(option) ? ((winners.index(option) % 2 == 0) ? :win : :lose)
end
```

This has the advantage of isolating the logic in a method that has a clear name, so we don't have to spend too much time reading it (to someone else, it might not be clear why we do have an array such as `['', '', 'sp', 'ps', 'pr', 'rp', 'rs', 'sr']` — but by wrapping this into a method named with intent, it improves readability)

lib/player.rb
```diff
@@ -0,0 +1,26 @@
+class Player
+  attr_reader :name
+
+  def initialize(name = "Computer")
+    @score = 0
+    @name = name
+    @option = nil
+  end
+
+  def score
```
leoht, line 10: You could use `attr_reader :score` to save some lines here :) (same for `@option`)

lib/player.rb
```diff
@@ -0,0 +1,26 @@
...
+    @score
+  end
+
+  def score_plus
+    @score += 1
+  end
+
+  def set_option(option = [:rock, :paper, :scissors].sample)
+    ([:rock, :paper, :scissors].include?(option)) ? (@option = option) : (@option = nil)
```
leoht, line 19: It's usually better to use only ternary conditionals when returning/evaluating a value, such as `num % 2 == 0 ? 'even' : 'odd'` — when assigning values or executing something more complex, it's usually to expand this with `if/else` statements (most rubocop configurations in codebases usually consider using ternary conditionals as warnings or offences, because they're harder to read)

leoht (commented):
Great job! I had some comments here and there mainly to improve readability and simplify the code, but overall everything seems quite good :) 
The one thing missing could be some unit tests for the `Game` class for example, along the feature tests you've already written

necsanuradu:
Thank you Léonard, much appreciated, I will refactor with all suggested changes. On the views I kept them separate because I have the tendency of moving towards a single page app and end up with a lot on any view. Also the inline css was added thinking that it will be little of it but then just seemed necessary more and more over(I will put it in classes). Thank you again.

PR #1927, opened 2021-08-15
Weekend challenge: week three

edpackard:
Pleased with this on the whole but would like particular feedback on:

* Capybara tests - are these OK? I'm ok with setting up the feature tests, but my `expect` statements seem to be quite limited - i.e. mainly expecting the page to contain something (or not).
* app.rb - most of the game logic is in the model, but there are a couple of bits of logic left in the controller because it seemed the most straightforward place to put it - is my controller skinny enough?
* html - any comments/advice on my html gratefully received.


app.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/name' do
+    player1 = Player.new(params[:name])
+    player2 = Player.new(params[:name2])
+    @game = Game.generate(player1, player2)
```
Nicola-Carroll, line 24: Unclear what `Game.instance` and `Game.generate(..)` do and why you haven't just initiated `Game.new(..)` on line 24 and removed line 14. Not sure what these different methods do so interested to hear your reasoning

app.rb
```diff
@@ -0,0 +1,50 @@
...
+    redirect '/' if params[:name].empty?
```
Nicola-Carroll, line 25: Would be cool if this also returned a message to the user when player 1 name is left blank

app.rb
```diff
@@ -0,0 +1,50 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    erb(:play)
+  end
+
+  post '/choice' do
+    @game.current_player.choose(params[:choice])
+    @game.change_current_player(@game.player2.type)
```
Nicola-Carroll, line 35: I understand that the overall point of this is to make the game compatible with both multiplayer and single player mode, but (i personally) found the code in `game.rb` and `player.rb` quite difficult to dissect on a line by line basis. I didn't try this additional exercise so not an authority to comment on how challenging it was but I wonder if there might've been a more simple way to do it , for instance if player 2's name or RPS move was left blank then just generate a random move.

Maybe if your `Game` class was blind to the fact of whether or not player 2 was a person or a computer it would make the code cleaner

app.rb
edpackard: `Game.generate` creates the new Game instance, while `Game.instance `then provides access to this instance - line 14 is invoked in every route using `before do ... end`. It's an alternative to `session` / global variables.

app.rb
edpackard: good idea - this was one of the last bugs I discovered, so created a quick fix, but a message would be good.

app.rb
edpackard: It does complicate the logic a bit having to deal with a computer player (who simply selects a random weapon) or a human player, who needs to enter a move on a form like player 1. The `change_current_player` method checks if player 2 is a computer player, in which case it does nothing, otherwise it changes the current player and goes back to the `/play` route, allowing player 2 to take a turn (per line 35). Once player 2 has a weapon, the `/play` route logic changes the player back to player 1 (in case they play again - though I might want to change things so that players alternate in who goes first) and goes to the `/result`.

I hope this clears things up - having gone over the logic again myself, I think `current_player` and `change_current_player` are OK names, as it is dealing with who is the current player in terms of making a choice of RPS.

PR #1930, opened 2021-08-16
RPS Challenge Lizzy Ballantine

eballantine:
Would really love some feedback., particularly around logic in the views (am I using acceptable 'light' logic or should this be extracted?) vs controller (again, do I have too much logic in there?) & models (does 2 classes cover it, are they following SRP?)
Finally, my use of sinatra 'sessions' - how might someone keep track of the score?

Attached is a pic of my initial plan. Black covering main requirements, green including an additional player, red for alternative game type.


app.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  post "/name" do
+    session[:player] = Player.new(params[:name])
+    session[:player2] = Player.new(params[:name2])
+    redirect "/play"
+  end
+
+  get "/play" do
+    @player = session[:player].name
```
leoht, line 24: Maybe it'd be good to rename those variables like `@player1_name` to indicate they contain the names, and not mix it up with Player instances

leoht (commented):
Really nice work! The code is clear to read and the tests are well written as well, good use of helpers to make the testing setup better to read too :) I just had a few comments here and there to improve things, but overall most of it is already really good. The logic in views is definitely OK since it only deals with conditions on what to display with `if/else` etc.

Regarding sessions, it's usually better to store only tiny pieces of data into it, such as IDs, strings, etc, rather than entire instances of `Player` for example. When using a database, for example, you'd first insert a record for a player and use sessions to keep track of their IDs, rather than the full data. One solution for this specific challenge is to keep a singleton instance in the `Game` class that lives accross requests, but for this case it's probably fine!

PR #1933, opened 2021-08-16
CB RPS-Challenge

CorinneBosch:


README.md
```diff
@@ -1,15 +1,48 @@
...
+      (____)                  _______)             (____)
+---.__(___)        ---.__________)           ---.__(___)
+
+```
+
+## Implementation
+
+Functional representation of user requirements:
+
+[Check out my workflow mindmap](https://github.com/CorinneBosch/rps-challenge/blob/main/public/images/workflow.png)
```
Battery0, line 18: Really nicely laid out and clear README with explicit instructions. I really like the inclusion of the mind map to show your thinking of how to lay out the MVC.

lib/player.rb
```diff
@@ -0,0 +1,13 @@
+class Player
+  attr_reader :name, :p_move
```
Battery0, line 2: Maybe define what p_move is. I'm guessing this is player_move?

lib/robot.rb
```diff
@@ -0,0 +1,12 @@
+class Robot 
+  attr_reader :name, :random_move
+
+  def initialize(name = 'Elvis', random_move = MOVES.sample)
+    @name = name
+    @random_move = random_move
+  end
+
+  private
+  MOVES = ['rock', 'paper', 'scissors']
```
Battery0, line 10: MOVES doesn't have to be made private - as moves is a constant it's immutable and can't be changed anyway

lib/rps.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    (@score = score + 0) if (@player_move == @robot_move)
+    (@score = score + 1) if (signs[@player_move] == @robot_move)
+    (@score = score - 1) if (signs[@robot_move] == @player_move)
+  end
+end
+
+rps = Rps.new('rock', 'scissors')
+rps.result
+p rps
```
Battery0, line 39: I'm thinking lines 37 - 39 were for testing Ruby out put? These should probably be deleted

public/style.css
```diff
@@ -0,0 +1,62 @@
...
+  border-radius: 8px;
+  height: 5vh;
+  width: 80px;
+  font-family: courier new;
+  background-color: #fff9e6;
+  border: 2px solid #ffcccc;
+}
+
+button:hover {opacity: 0.6}
+
```
Battery0, line 62: Nicely laid out CSS, you clearly know a thing or two!

spec/features/infra_test_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature 'Testing infrastructure' do
+  scenario 'Can run app and prints content' do
+    visit('/')
+    # expect(page).to have_content 'Lets play Rock Paper Scissors!'
+  end
+end
```
Battery0, line 6: Not sure what this test is testing for?

spec/features/make_move_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    click_button 'paper'
+  end
+
+  scenario 'player chooses scissor move' do
+    visit('/game')
+    sign_in_and_play
+
+    click_button 'scissor'
+  end 
+
```
Battery0, line 22: You might be able to roll the first 3 tests into one... I think you're trying to check if there is a button named 'rock', 'paper' and 'scissors'?

app.rb
```diff
@@ -0,0 +1,45 @@
...
+require 'sinatra/reloader'
+
+class Game < Sinatra::Base
+
+  configure:development do 
+    register Sinatra::Reloader
+  end
+
+  before do
+    @game = Rps.instance
```
leoht, line 14: While it's good to use the `before` hook to assign `@game` for each request, there's probably no need to call `@game.result` (especially that the return value not assigned to any instance variable, so it is "lost" in a way)

app.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/welcome' do
+    player = Player.new(params[:name])
+    robot = Robot.new
+    @game = Rps.create(player, robot, 1)
```
leoht, line 25: Reading `Rps.create(player, robot, 1)`, we could be confused about what is this third argument `1`, that's a good example of a "magic number", and we'd need to read the class to understand more. This could be simplified by using the constant `SCORE` defined in the Rps class:
```ruby
Rps.create(player, robot, Rps::SCORE)
```

Maybe this constant could even be renamed to something more explicit like `INITIAL_SCORE`?

app.rb
```diff
@@ -0,0 +1,45 @@
...
+    erb :game
+  end
+
+  post '/move' do
+    @game.player_move.move(params[:move])
+    redirect '/result'
+  end
+
+  get '/result' do
+    @game.result
```
leoht, line 40: Here again, the view will not be able to access the result since it needs to be assigned to an instance variable, for example:
```ruby
@result = @game.result 
```

lib/robot.rb
```diff
@@ -0,0 +1,11 @@
+class Robot 
+  attr_reader :name, :random_move
+
+  def initialize(name = 'Elvis', random_move = MOVES.sample)
+    @name = name
```
leoht, line 5: That's where you could maybe implement a method e.g `def generate_random_move` for example, so you can stub it later in your tests. Then you could call it to assign the move:
```ruby
@random_move = generate_random_move
```

lib/rps.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def self.instance
+    @game
+  end
+
+  def result
+    score = { 'scissors' => 'paper', 'paper' => 'rock', 'rock' => 'scissors' }
+    # signs = { 'scissors' => 'paper', 'paper' => 'rock', 'rock' => 'scissors' }
+    
+    return 'Draw!' if @player_move == @robot_move
```
leoht, line 28: This is probably down to personal choice, but I'd usually avoid returning "visible" string messages that need to be displayed to the user from this method, but rather some simpler value, or a symbol maybe `:draw`, `:victory` etc. Maybe even better, they could be defined in constants. This avoids some potential errors when making typos, and also it makes sure that only the view is responsible for *how* to display the result to the user

leoht (commented):
Great work! I left a few comments here and there to improve the code and readability, also some tests are actually "dead" tests since they're partly commented out, so maybe something to double check. But overall that's really good :)

PR #1934, opened 2021-08-16
PR for review

Battery0:
PR for review


app.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  post '/choose_RPS' do
+    session[:player_1] = params[:player_1]
+    @player1_name = session[:player_1]
+    erb(:choose_rps)
+  end
+
+  post '/result' do
+    @player1_name = session[:player_1]
```
CorinneBosch, line 23: Would have maybe been neater to create a Game instance, especially if you create more get/post routes

lib/computer_rps.rb
```diff
@@ -0,0 +1,18 @@
+class ComputerRPS
+  
+  def random_number
+    rand(1..3)
+  end
+
+  def rock_paper_scissors
```
CorinneBosch, line 7: As a refactor input, you are able to run  [].sample form an array of strings! It will allow you to use the string in the erb files.

spec/unit/gameplay_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'gameplay.rb'
+
+describe GamePlay do
+
+  let(:computer_choice) { double("double", { rock_paper_scissors: "Scissors" }) }
+
+  subject { described_class.new(computer_choice) }
+
```
CorinneBosch, line 8: I really like the way you tested this!

views/choose_rps.erb
```diff
@@ -0,0 +1,36 @@
...
+
+  <head>
+    <title>Rock Paper Scissors Game</title>
+    <style>
+      body {
+        font-family:helvetica;
+      }
+    </style>
+  </head>
+
```
CorinneBosch, line 12: just for good practice I'd suggest making a separate CSS file for your design changes as the erb file should be kept as light as possible. Also it allows you to access all erb files at once

PR #1935, opened 2021-08-16
rps-challenge

danroche10:


app.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/name' do
+    player_1 = Player.new(params[:player_1])
+    player_2 = Player.new(params[:player_2])
+    @stored_game = Game.new_game(player_1, player_2)
```
leoht, line 19: You could remove `@stored_game` since it's not going to be used later in a view (for example), calling `Game.new_game` is enough to set the class variable `@stored_game` within the class so it persists across requests

lib/game.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'player'
+
+class Game
+
+  attr_reader :player_1, :player_2, :stored_game
+
+  def initialize(player_class_1:, player_class_2:)
```
leoht, line 7: Maybe it'd be good to change the arguments name to `player_1` and `player_2` as they'll contain instances of the Player class, rather than the classes themselves

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+    @player_1 = player_class_1
+    @player_2 = player_class_2
+  end
+
+  def winning_player
+
+    winning_moves = { "scissors" => "paper", "paper" => "rock", "rock" => "scissors" }
+
+    return 'draw' if player_1.move_choice == player_2.move_choice
```
leoht, line 16: Good use of an early return :) however, since the method is named `winning_player`, we would probably expect it to return a player instance or name — or maybe `nil` if it's a draw

spec/features/index_spec.rb
```diff
@@ -0,0 +1,28 @@
+feature 'players can correctly see everything on play page' do
+
+  before do
+    sign_in_and_play
```
leoht, line 4: Maybe it could be good to pass in arguments the player names, so it's clearer when reading the call to this helper function that e.g "Dan" and "Opponent" will be used as names, otherwise it's not super clear when reading the tests below why those names have been used

spec/features/index_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  feature 'Player form' do
+    scenario 'expects player to submit name and then see the contest' do
+      expect(page).to have_content "Dan vs Opponent"
+    end
+  end
+
+  feature 'Move options' do
+    scenario 'expects Dan to see where they should enter move' do
+      expect(page).to have_content "Dan move:"
```
leoht, line 15: Maybe an added assertion to verify we have at least the right input field/HTML input on that page — such as `expect(page).to have_selector("input[name='player_1_move']")`

spec/features/index_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    scenario 'expects opponent to see where they should enter move' do
+      expect(page).to have_content "Opponent move:"
+    end
+
+    scenario 'expects player to see list of options to choose from' do
+      sign_in_and_play
```
leoht, line 23: You could probably remove it from here since it's already called in the `before` hook

spec/game_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    it 'initializes class with store_game equal to nil' do
+      expect(game.stored_game).to eq nil
+    end
+  end
+
+  describe '#winning_player' do
+
+    it 'returns corrent outcome bases on player moves' do
+      player_1.move_choice
```
leoht, line 26: The method `move_choice` is an attribute reader, so it's only used to get the value of this attribute — since there it's called but the result not assigned to any variable, it could probably be removed

leoht: And the same seems true for calling `game.winner_player` below :)

spec/game_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  describe '#result' do
+    it 'correctly returns score message when player 1 wins' do
+      game.winning_player
+      expect(game.result).to eq 'player_1 wins'
+    end
+  end
+
+  describe '#self.new_game' do
+    it 'receives player two arguments' do
+      allow(Game).to receive(:game).with(player_1, player_2)
```
leoht, line 41: There are a few things here:
  * The method `game` is not defined, so you probably meant `allow(Game).to receive(:new_game)`?
  * Using `allow` stubs the method and we can force it to return some predictable value — to *verify* the method has been called, you'd need to use `expect(Game).to receive(...)`

leoht (commented):
Great work! I've left a few comments mainly to improve some details the code and tests as well, but overall that's on good tracks :) you can revisit the challenge in your own time to address this feedback if you'd like!

lib/game.rb
danroche10: So just: def initialize(player_1, player_2)  ?

PR #1939, opened 2021-08-16
PR for review

Battery0:
PR for review


leoht (commented):
Nice work! It would be good to have at least one feature test that tries the whole flow — user enters their name, choose a move and we get the result (you could mock the class `ComputerRPS` like you did in the unit tests, to force the "random" move). Otherwise I just had a few comments here and there, mainly to improve code readability etc :)

PR #1940, opened 2021-09-11
RPS Challenge

EMDevelop:


README.md
```diff
@@ -1,86 +1,142 @@
...
+    - play_spec.rb
+    - results_spec.rb
+    - helper.rb
+  - game_spec.rb
+  - player_spec.rb
+  - computer_spec.rb
+
+---
+
+## HTTP requests
```
leoht, line 110: Nice detailed README!

app/views/index.erb
```diff
@@ -0,0 +1,7 @@
+<h1> Welcome to RPS, Please enter your Name and click "Play" <h1>
+
+<form action="/register" method="post">
+  <input type="text" name="name" placeholder="Type Name...">
+  <input type="submit" value="Register"
```
leoht, line 5: Be careful as you're missing the closing `/>` here - browsers are usually quite permissive with this and they'll "make up" for mistakes, but this can create other kind of weird bugs that are difficult to track down. There are some online HTML validators in which you can copy the source code of the resulting page and see which errors to potentially fix

lib/rps.rb
```diff
@@ -0,0 +1,51 @@
+class RPS
+
+  def initialize(player_one, player_two)
+    @players = [player_one, player_two]
+  end
+
+  attr_reader :players
+
+  def self.new_game(player_one, player_two)
```
leoht, line 9: Conventionally, in Ruby class methods usually go at the top of the class definition (before instance methods and `initialize` too) - but that's me being picky!

lib/rps.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def show_winner(player_choice)
+    p1 = player_choice
+    p2 = get_computer_choice
+    calculate_winner(p1,p2)
+  end
+
+  private 
+
+  def calculate_winner(p1,p2)
```
leoht, line 33: This method is returning a message to display to the user depending on the game result — this would belong more on the view layer (= in the ERB template file) rather than inside this class (the model layer). To keep a good separation of concerns, this method could return the player who win (or the winner player name, maybe), so it only returns a "data" result. Maybe a symbol, like `:draw` or `:win` or `:lose`. Then the view layer can use this result value to display the appropriate message of interface :)

lib/rps.rb
```diff
@@ -0,0 +1,51 @@
...
+    @players[0]
+  end
+
+  def player_two
+    @players[1]
+  end
+
+  def show_winner(player_choice)
+    p1 = player_choice
+    p2 = get_computer_choice
```
leoht, line 27: Maybe the method `get_computer_choice` is not really needed, since it only forwards the call to `player_two.chose`. It's probably fine to call it directly (and this way, if you want later to make player 2 a human player, it's easier to change!)

lib/rps.rb
```diff
@@ -0,0 +1,51 @@
...
+      "Unlucky, your #{p1} was beaten by #{@players[1].name}'s #{p2}"
+    end
+  end
+
+  def win_matrix
+    { 'scissors' => 'paper', 'paper' => 'rock', 'rock' => 'scissors' }
+  end
+
+  def get_computer_choice
+    @player_two.chose
```
leoht, line 48: I'm not sure I've seen `@player_two` be set to a value before (but there is a method `player_two`). Be careful as otherwise `@player_two` will be `nil` if it's not set before, so running this method will throw an error. Maybe it'd be better to replace it by the method you've defined earlier:

```ruby
def get_computer_choice
    player_two.chose
end
```

spec/computer_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'computer'
+describe Computer do
+
+  context 'the game will choose a random option' do
+
+    it "checks computer has name" do
+      expect(subject.name).to eq "Computer"
+    end
+
+    # this test passess regardless of whether the `choose` method exists  on `computer.rb`, Why?
```
leoht, line 10: That's because when writing:

```
allow(subject).to receive(:choose).and_return("scissors")
```

you allow this method to be called on the given instance (even if this method was never defined by the class! So you can stub methods that do not exist, even though that's not really useful (and potentially dangerous)

spec/feature/index_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    it 'Shows name after submit' do
+      register_name
+      expect(page).to have_content(player_output)
+    end
+
+    it 'shows options' do
+      register_name
+      expect(click_button('rock'))
```
leoht, line 19: Not sure what `expect(click_button('rock'))` is verifying? That the button exists on the page I guess? Is that a pattern you've seen somewhere else maybe? 🤔

spec/rps_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rps'
+
+describe RPS do
+
+  before do
+    RPS.new_game(player, computer)
+  end
+
+  #Can't figure out how to access the instance variable "name" from a double?
```
leoht, line 9: You cannot — doubles don't have instance variables, they can only respond to method calls. So if you have a method (it can be defined with an attr_reader) that returns the name, like `player.name`, you can define this method on your double to return the name value you want

leoht (commented):
Nice work! There are a few things to improve here and there, I've left some comments to clarify things and answer where they were questions in comments too — but let me know if things aren't clear or if you'd like to follow up on it :)

lib/rps.rb
EMDevelop: Actually I was thinking of doing this, it makes sense for them to go at the top as they're grouped together! thanks :) picky is good

README.md
EMDevelop: Why thankyou!

app/views/index.erb
EMDevelop: Ah, youtube has lead me down a bad path! A video said something about html 5 onwards you didn't need the closing / on `br` and ``````input` anymore - but I didn't think about the implications of other browsers. thanks :)

lib/rps.rb
EMDevelop: That makes a lot of sense - I'll do that in the next task!

lib/rps.rb
EMDevelop: Good point, I didn't have much time this weekend so I decided to stick to the basic, I think I was getting carried away with my private methods haha.

lib/rps.rb
EMDevelop: Ah yes, I actually changed this immediately after the pull request, it should say this 
  def get_computer_choice
    @players[1].choose
  end

spec/computer_spec.rb
EMDevelop: I thought this was happening, but now I'm a little confused about how I should have tested the random outcome ?

spec/feature/index_spec.rb
EMDevelop: It's just checking that the button exists on the page exactly, I made it up to be honest, but it seemed to do the job !

spec/rps_spec.rb
EMDevelop: I see. cool thanks @leoht

PR #1945, opened 2021-09-13
rps-challenge

GianlucaAnsaldi:
Hi coaches, can you please review my challenge? I would really much appreciate if you could comment on the '/move' and '/results' routes in the app.rb as I am not sure I made the best use of the sessions to store variables and params. Also I like how I called the final views in the '/results/ route but I'm not sure if passing the parameters to the function using sessions is the right approach.

I would also appreciate some advice on the feature tests. To be honest I had to reverse engineer the results_spec feature test as I wasn't sure how to get to that last page and check if it was ok. The result can't be arbitrarily decided in advance and I wasn't sure if/how to do the mocking on these tests, so I ended up having to add some common text in all three views so that whatever the result was I could check that page had some specific text in it. I was able to check the results in the unit test for the game class, but not the in the feature test for the '/results' route.

Also, I had a feature test for select_move_spec when I was in the early stages, I used that to check that the player was able to pick a weapon and that the value was stored somewhere. When I moved on with the code I didn't feel I had the need for those tests anymore (since now I would go directly to the '/results' route with the result of the fight while earlier in that page I was showing what weapon the player picked) so I removed it, but not sure if I would still need that test and how to refactor it eventually now with the code I have.

Thanks!


app.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    player = Player.new(params[:player_name])
+    computer = Computer.new
+    session[:game] = Game.new(player, computer)
```
leoht, line 21: It's usually better to not store "complex" objects like this one into session, but rather simple values, like strings or numbers (it will probably work in this case though). Obviously we weren't using databases last week so that's not a bad thing for this challenge — going further, how you would do it using databases is that the `Game` state could be saved into a table, and you would just store the ID of that record into the session, to be able to fetch the entire data of the game using the ID later

app.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  get '/play' do
+    @game = session[:game]
+    erb :play
+  end
+
+  post '/move' do
+    @game = session[:game]
+    session[:player_move] = params[:weapon]
+    session[:computer_move] = @game.player_2.choose
```
leoht, line 33: An alternative way of doing it without putting too many things in session could be to compute the game result within this route:

```ruby
player_move = params[:weapon]
computer_move = @game.player_2.choose
result = @game.check_result(session[:player_move], session[:computer_move])

# ... redirect (or not!)
```

Then you can either set the game result in session and redirect to /results, or you could even not redirect and display a different template with `erb :win` for example

app.rb
```diff
@@ -0,0 +1,43 @@
...
+    redirect '/results'
+  end
+
+  get '/results' do
+    @game = session[:game]
+    erb @game.check_result(session[:player_move], session[:computer_move])
```
leoht, line 39: If you're using the result value as the template name, it's probably best to check first if this value is a valid one (otherwise we might end up having an error because of a view not existing). It can be as simple as `if [:win, :lose, :draw].include?(result)` for example

lib/game.rb
```diff
@@ -0,0 +1,26 @@
...
+  def player_1
+    @players.first
+  end
+
+  def player_2
+    @players.last
+  end
+
+  def check_result(player_choice, computer_choice)
+    return :win if RULES[player_choice.to_sym].include?(computer_choice)
```
leoht, line 22: Nice and neat logic here :) the only improvement I could suggest would be to rename the method to `compute_result` or just `result` (`check_***` makes use think that it's checking or verifying something)

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def register_name
+  visit('/')
```
leoht, line 2: I'd suggest having the player name passed in argument to this helper, this way you can be more explicit on the name when using it in the tests :)

spec/features/register_name_spec.rb
```diff
@@ -0,0 +1,21 @@
+feature "Enter name" do
+  scenario "register name" do
+    register_name
+    expect(page).to have_content "Gianluca you are now facing Archaon"
+  end
+
+  scenario "page has Rock button" do
```
leoht, line 7: Those are actually not three different scenarios, as you'd expect to see all three buttons on the same page after registering the name. So it could all be simplified as:

```ruby
scenario "register name" do
    register_name
    expect(page).to have_content "Gianluca you are now facing Archaon"
    expect(page).to have_selector(:link_or_button, "Rock")
    expect(page).to have_selector(:link_or_button, "Paper")
    expect(page).to have_selector(:link_or_button, "Scissors")
  end
```

spec/features/results_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature "Result" do
+
+  scenario "lands in the results page" do
```
leoht, line 3: That's where you'd use stubbing, like in unit tests, to test different scenarios outcomes (that'd be different `scenario` blocks then)

views/index.erb
```diff
@@ -0,0 +1,8 @@
+<h2><%= "Welcome to the RPS Arena" %></h2>
```
leoht, line 1: For "hardcoded" text like this, you don't need ERB tags, you could just use plain HTML/text: 
```
<h2>Welcome to the RPS Arena</h2>
```

This simplifies reading and make the code (slightly slightly) faster, as Ruby doesn't need to parse and interpret all the content that is not between ERB tags

leoht (commented):
Nice job! I added a few comments here and there but overall it's quite good. It's better to have a "full" HTML structure for your pages as well, otherwise it could cause some issues (and on the long run, it's also better for technical reasons, SEO reasons, and accessibility reasons, to have a full HTML scaffold — even if the browser compensates for it and usually adds all the missing tags and structure)

To specifically answer your questions:
 * I did comment about the session below, but generally you'd put the minimum amount of info into session that allows you to identify who is browsing (or which game ID was started, for example) and then retrieve this data. In this case using the session to store the `Game` instance is fine though. 
 * You can definitely mock and stub in feature tests as well — which means you can stub the random value like you did in unit tests, and then assert on the content of the page depending on the value you stubbed
 * You probably don't need that test indeed, usually feature tests are testing for a complete user flow (or a complete "block" of a longer user flow) — in this exercise the user flow is rather simple so it's fine to remove this previous test that was just covering the start of it. If selecting a weapon wasn't working properly, it would break the whole test anyway, as you wouldn't get the correct result :)

PR #1949, opened 2021-09-13
rps - challenge 

Odrakes1992:


app.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  post '/name' do 
+    session[:player_1_name] = params[:player_1_name]
+    redirect '/play'
+  end
+
+  get '/play' do 
+    cap_name = session[:player_1_name]
+    @player_1_name = cap_name.capitalize
```
eoinmakers, line 24: You could probably combine these two lines, to something like @player_1_name = session[:player_1_name].capitalize

app.rb
```diff
@@ -0,0 +1,43 @@
...
+  get '/game' do
+    cap_name = session[:player_1_name]
+    @player_1_name = cap_name.capitalize
+    erb(:game)
+  end
+
+  get '/result' do
+    @player_choice = params[:choice]
+    @computer = CPU.new
+    @game = Game.new
```
eoinmakers, line 37: This is along the right track, but something is missing. 

At the moment, you make instances of computer and game, but they don't return anything. 

So to finish this route:
- How can you return a result from @computer?
- The Game class will need to be able to access both the @player_choice and the computer's choice: can you see a way in which you could pass both of those choices into the game class, so the game logic can read them? 
- Your game logic currently prints out who the winner is - which won't be visible to the user in any view. You could pass the result back as a variable, and redirect to a new view detailing who the winner is.

eoinmakers (commented):
This is very much on the correct path - I just left one main comment on where you should take your implementation next, as I think it'll require some refactoring in order to be use the code you already have to determine the winner. 

Let me know if you have any questions!

PR #1951, opened 2021-09-13
Weekend Challenge Week 2

SophieBonneval:


lib/game.rb
```diff
@@ -0,0 +1,31 @@
+class Game
+  STATEMENTS = ['Scissors cuts Paper',
```
leoht, line 2: I'd suggest moving those message strings into a separate class, maybe called `ViewMessages` or something, to not have any "view" or presentation logic into the `Game` class, to respect separation of concerns. The `Game` class deals with the logic of the game, so ideally a different class or component would take care of presenting the game results to the user. You can then either get the right statement in the controller or straight from the view file. This could look like this:

```ruby
# in app.rb
game_result = game.result(choice1, choice2)
@statement = view_helper.statement(game_result)
```

or in the ERB file:
```
<%= @view_helper.statement(@game_result) %>
```

but the core idea is that game logic and presentation are not mixed in the same class :)

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+                'Spock smashes Scissors'
+                ].freeze
+  attr_reader :choices
+
+  def initialize
+    @choices = ["Rock", "Paper", "Scissors", "Lizard", "Spock"]
+  end
+
+  def result(choice1, choice2)
+    return "This is a draw" if choice1 == choice2
```
leoht, line 20: Same as above here — it'd be better if this method returned a "result" value, maybe a symbol (`:draw`, `:win`, etc) instead of a message in English, to leave this to the presentation/view layer. Imagine you later want to translate the game in different languages, it then makes a lot of sense to decouple this class from how you're going to present the results to the user

spec/features/enter_name_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'Submitting name' do
+  scenario 'User enters his name on the index page and lands on the page to play' do
+    enter_name
```
leoht, line 3: Maybe it'd be nice to pass the player name to this helper method, in order to make it more explicit which name is used:
```ruby
enter_name('Sophie')
expect(page).to have_content 'Hello Sophie! Ready to play?'
```

views/layout.erb
```diff
@@ -0,0 +1,110 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link href="https://fonts.googleapis.com/css2?family=Boogaloo&display=swap" rel="stylesheet"> 
+    <style>
```
leoht: So a way to avoid putting CSS in the HTML is to link an external CSS file. I think by default Sinatra serve static files in a `public` directory, so you could put the CSS file in there and use the `<link>` tag to include it inside the HTML file. This post explains how to do it: https://medium.com/arren-alexander/adding-css-to-your-sinatra-project-f82190e26ca0

views/play.erb
```diff
@@ -0,0 +1,20 @@
...
...4.5.1 1.3.8 2.2 1.5 3.9.2 6-4.7 12.1-3.9 17.7.7 11.8 9.6 23 20.1 33.4 31.2 11.3 11.8 13.6 24.8 8.2 39.9-2.9 8.1-6.1 16.2-9.6 24-6.6 14.4-13.6 28.4-24.7 40-4.2 4.4-6.1 9.3-5.1 15.8 1.3 9.2 2 18.5 2.2 27.8.1 4.8.9 5.4 5.3 3.5 28.5-12.4 49.8-32.3 64-60.1 14.5-28.5 17.1-58.3 9-88.7-9.5-35.5-31.2-61.8-63.7-79.1-21.5-11.3-45.9-16-70.1-13.3-36.9 4.1-67 21.2-89 51.9z"/><path class="st2" d="M68.8 94.5c22-30.7 52.3-47.6 89.6-51.5 24.1-2.6 48.5 2.2 69.9 13.9 32.5 17.4 54.1 43.8 63.4 79.4 8 30.5 5.2 60.2-9.4 88.7-14.2 27.8-35.7 47.5-64.2 59.8-4.3 1.8-5.2 1.3-5.3-3.5-.1-9.4-.8-18.6-2.1-27.8-.9-6.5 1-11.4 5.2-15.8 11.1-11.6 18.2-25.6 24.8-39.9 3.6-7.9 6.8-15.9 9.7-24 5.5-15 3.3-28.1-8-39.9-10.5-11.1-21.5-21.5-33.3-31.4-5.6-4.6-11.7-5.4-17.7-.8-1.7 1.3-2.6.6-3.9-.2-8.2-5.1-16.3-4.3-24.5-.2-1.2.8-2.9.9-4.1.2-8.1-4-15-.9-20.8 4.5-1.7 1.8-4.1 2.7-6.5 2.6-9.9-.5-17.9 3-23 11.9-1 1.7-2.2 2.7-4.4 2.6-5.2-.3-8 3.1-9.6 7.2-6.7 15.5-11.3 31.7-15.6 47.9-1.7 6.8.6 13.6 3.3 19.9 4.6 10.9 11.2 20.3 20.8 27.5 3 2.3 5.8 4.9 8.6 7.5 2.3 1.9 3.5 4.8 3.4 7.9-.3 11.9-.3 23.9-.3 36 0 2.6.2 5-3.7 2.7-18.5-10.5-34.2-24-45.9-41.9-8.9-13.6-15.1-28.6-18.3-44.5-5.4-27.6-2.1-53.9 9.9-79.2 3.2-7.1 7.2-13.5 12-19.6z"/></svg>
```
leoht, line 5: Same as for the CSS above, loading SVG images would be better since otherwise it's quite hard to read the HTML here :D

views/result.erb
```diff
@@ -0,0 +1,12 @@
+<h1><%= @result %></h1>
+
+<p>You chose: <%= @player_choice %></p>
+<p>The computer chose: <%= @computer_choice %></p>
+<p class="statement"><%= @statement %></p>
+
+
+<form id="play_again" action="/play">
```
leoht, line 8: Since this form is going to send a GET request to `/play`, adding a `<a>` link to this URL would have the same behaviour so you could simplify the code:
```
<a href="/play">Play again</a>
```

Then if you want to customise how the link looks, you can use some CSS too

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+                ].freeze
+  attr_reader :choices
+
+  def initialize
+    @choices = ["Rock", "Paper", "Scissors", "Lizard", "Spock"]
+  end
+
+  def result(choice1, choice2)
+    return "This is a draw" if choice1 == choice2 
+    STATEMENTS.find { |str| str.match(/#{choice1}.+#{choice2}/) } ? "You win" : "You lose"
```
leoht, line 21: Interesting way to code the game logic though! :)

spec/features/choose_option_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'Choose option' do
+  scenario 'User enters a name, lands on the play page, chooses an option and see result' do
+    enter_name
```
leoht, line 3: You could probably add a few scenarios to also test the game logic is working (e.g I'm choosing Rock, the computer chooses Scissors, so I win). You can use RSpec stubbing with `allow(...)`, like in unit tests, to force the random computer move to be a specific value

leoht (commented):
Great work :) I just had a few comments here and there about code design and tests, let me know if you have any question on that or if you'd like to follow up on the feedback

PR #1953, opened 2021-09-13
RPS weekend challenge, week 3

lildann:
Summary
----
* Found this challenge very challenging! Identified a lot of gaps in my learning and decided to go over the course material rather than continue with the challenge.

* Understand the MVC layout structure and feature testing process. But unsure of how input (such as a ticked checkbox) can be stored and used by the model (game.rb). 

* Wrote rspec tests for rock paper scissor game, and then wrote the logic, then tested this in irb and needed to tweak my code, but then the rspec tests failed and I couldn't figure out how to make them pass. It could be something to do with my stubbing.


spec/game_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    it 'input from a computer opponent' do
+      allow(subject).to receive(:opponent) { "rock" }
+      expect(subject.opponent).to eq("rock")
+    end
+
+    it 'answer is random from the computer opponent' do
+      expect(["rock","paper","scissor"]).to include(subject.opponent)
+    end
+  end 
+
```
florence-bain, line 23: It is really easy to tell what your method names do and their purpose. I think you were really close with these tests, it might have been easier to split them into another spec to keep it simpler.

views/play.erb
```diff
@@ -0,0 +1,15 @@
...
+  <br>
+  <label for="paper">p  a  p  e  r</label>
+  <input id="paper" name="paper" type="checkbox" value="paper">
+  <br>
+  <label for="scissor">s  c  i  s  s  o  r</label>
+  <input id="scissor" name="scissor" type="checkbox" value="scissor">
+  <br>
+</form>
+  <a href="/result">did you win or lose?</a>
+
```
florence-bain, line 16: Nice use of Html love that you're trying out different features for the UI of your app

README_LD.md
```diff
@@ -0,0 +1,43 @@
+Summary
+----
+* Found this challenge very challenging! Identified a lot of gaps in my learning and decided to go over the course material rather than continue with the challenge.
+
+* Understand the MVC layout structure and feature testing process. But unsure of how input (such as a ticked checkbox) can be stored and used by the model (game.rb). 
+
+* Wrote rspec tests for rock paper scissor game, and then wrote the logic, then tested this in irb and needed to tweak my code, but then the rspec tests failed and I couldn't figure out how to make them pass. It could be something to do with my stubbing. 
+
+
```
florence-bain: Great plan! I like that you established the requirements of your app as well as the flow in detail

lib/game.rb
```diff
@@ -0,0 +1,27 @@
+class Game 
+  attr_reader :input
+
+  def initialize(player_input = "paper") #default for testing
+    @player_input = player_input
+    @input = [player_input]
+  end
+  
+  def opponent
+    opponent_input = ["rock", "paper", "scissor"].sample
```
leoht: It'd probably be better to encapsulate this `["rock", "paper", "scissor"].sample` into a separate method — this way, you can stub it when you'll need to write your tests, to force it to return a predictable value

leoht: Also, the method could be renamed `select_opponent_input` or something like this, it make it clearer what action the method is doing (`opponent` as a method name indicates that is probably just getting the opponent, reading data, etc)

lib/game.rb
```diff
@@ -0,0 +1,27 @@
...
+    @input << opponent_input
+    opponent_input
+  end
+
+  def result
+    player, opponent = @input[0], @input[1]
```
leoht: You can also just write (since the righthand value is an array):

```ruby
player, opponent = @input 
```

spec/game_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  context "result" do
+  #   it 'rock vs. scissor' do 
+  #     allow(game).to receive(:input) { ["rock", "scissor"] }
```
leoht: Rather than stubbing the whole input, you just need to stub the move from the "computer" player (see the comment above about putting the random part into a separate method). If you stub only this part, you can give the player input to the game and you should get the expected results :)

app.rb
```diff
@@ -0,0 +1,35 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+
+class RPS < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader #registering necessary for modular applications (if development works only for 'classic applications #sinatrarb.com/contrib/reloader
+  end
+  enable :sessions
+
```
leoht, line 10: In order to pass form data through the controller to the model, you can use the `params` hash. It can be used for text inputs, checkboxes, and other kind of form controls. You can debug its contents after sending a form to see what values are in there:

```
p params
```

Then you can assign these values to the controller instance variables (or to regular variables) in order to pass them to the model layer through methods. Here's an example:

```ruby
@name = params[:player_name] # params contains form/request values (POST or GET)
Game.register_name(@name)
```

It's important to note that you should set values on the controller in order to pass them into the view. You shouldn't call the model classes directly from the view, like this:
```ruby
<body>
  <p>Your name is <%= Game.player_name %></p>
</body>
```

But rather call the model inside app.rb, set the values you need to instance variables, so the view can access them

lib/game.rb
```diff
@@ -0,0 +1,27 @@
...
+    case 
+      when opponent == player then "Draw!"
```
leoht: It'd be better if this method returned a "data" result (maybe a symbol like `:win`, `:lose` etc) rather than a message string. That's because strings that the user will see belong in the view layer (the ERB templates). So this method would just return a result that we can then pass along to the view — and the view decides what message or interface to display. This is in order to have better separation of concerns and separate the logic and data manipulation from how it's displayed

leoht (commented):
That's a great start, you're just missing a few things (on stubbing randomness, and getting values from the request) that will help you unblock and complete the challenge further :) I've left comments on those different points and hopefully that'll answer your questions too, but let me know if something isn't clear or if you need to follow up on it!

PR #1958, opened 2021-09-13
Jakes RPS challenge

Jjake540:
Hi coaches, Here's my RPS challenge. I feel like I was going in the right direction with this. I hit a few blockers on the way. I tried to test to see if I could get opponent RPS to appear on a /results page and knowing globle variables is a sin I wanted to avoid this. Also struggled to get my head around linking the Classes with the instances on the app.rb file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,9 @@ source 'https://rubygems.org'
 ruby '3.0.2'
 
 gem 'sinatra'
+gem 'sinatra-contrib'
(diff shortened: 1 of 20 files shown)


jordanveness (commented):
Overall, a good systematic approach with the use of instance variables and classes, the only thing to add of note would be the change of spelling of 'choises' to choices as they differ throughout the program, it may make it confusing for someone else interpreting this!

PR #1963, opened 2021-10-09
RPS Challenge - Jacob

jtc27:


app.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  post '/names' do
+    player_1 = Player.new(params[:player_1_name])
+    player_2 = Computer.new('RPS Computer')
+    @game = Game.create(player_1, player_2)
+    redirect '/play'
+  end
+
+  get '/play' do
+    @game
```
leoht: This line is probably not necessary (since you're not assigning anything new or calling any method), so it could be removed

app.rb
```diff
@@ -0,0 +1,47 @@
...
+require './lib/computer'
+
+
+class BookmarkManager < Sinatra::Base
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  before do
```
leoht, line 15: Nice use of the `before` hook here :)

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  RULES = {
+    :rock     => {:rock => :draw, :paper => :lose, :scissors => :win},
+    :paper    => {:rock => :win, :paper => :draw, :scissors => :lose},
+    :scissors => {:rock => :lose, :paper => :win, :scissors => :draw}
+    }
+  
+  attr_reader :player_1, :player_2
+  
+    def initialize(player_1, player_2)
```
leoht: Minor — the indentation seems slightly off here, you could use a tool like rubocop to automatically fix that :)

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+      @player_1 = player_1
+      @player_2 = player_2
+    end
+
+    def result(p1_choice, p2_choice)
+      RULES[@player_1.weapon(p1_choice)][p2_choice]
```
leoht: Minor — you might want to add some code here to prevent returning a `nil` value, just in case the choice doesn't end up being a valid key in this hash

lib/player.rb
```diff
@@ -0,0 +1,15 @@
+
+
+class Player
+  attr_reader :name, :weapon
+
+  def initialize(name)
+    @name = name
+  end
+
+  def weapon(weapon)
```
leoht: I'd probably rename this method to something like `set_weapon` to indicate it's setting the value, just to make it a bit clearer when calling the method

spec/features/web_methods.rb
```diff
@@ -0,0 +1,6 @@
+def sign_in_and_play
```
leoht: It might be good to pass the player name used in argument here, so it's more explicit when the helper method is called in your tests, example:
```ruby
sign_in_and_play('Timmy')
expect(page).to have_content 'Timmy'
```

spec/features/result_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  subject(:computer) { Computer }
+
+  scenario 'P1 chooses rock' do
+    sign_in_and_play
+    click_button 'rock'
+    expect(page).not_to have_content 'You chose: scissors'
+    expect(page).to have_content 'You chose: rock'
+  end
+
+before do
```
leoht: Minor — it'd be good to move the `before` block on the top, before the different scenarios (also  the indentation is slightly off)

views/draw.erb
```diff
@@ -0,0 +1,29 @@
...
+</h1>
+
+<form action="/play" method="get">
+  <div id="centerbuttons">
+    <input type="submit" value="Play again">
+  </div>
+</form> 
+
+<style>
+#centerbuttons{
```
leoht, line 22: Generally it's better to use external CSS stylesheets (with the `<link />` HTML tag) rather than using `<style>` tag or with the style attribute (it makes it easier to change CSS rules and so the design/style of your page later). It's probably fine for this challenge, but usually a good rule of thumb is to add your CSS in an external file and load it on your HTML page (you then use HTML ids and classes to target the elements with CSS rules)

leoht (commented):
Nice work! I mainly have a few minor comments to improve some things in the code, but overall the code is readable, the tests are well written and stubbing is well used. Another comment I'd have would be on the HTML code, using external CSS stylesheets and having a "full" HTML skeleton (with a `<!doctype html>`, `<html>`, `<head>` and `<body>` tags, etc) is better (even if browsers compensate for this and the page still displays normally). If you have any question on that feedback, let me know :)

PR #1964, opened 2021-10-10
RPS Challenge Submission

alexboyling:
## Approach

* Read through the two user stories
* Created a model of the classes and methods required
* Used the singleton method on the player class to make attributes available
* Borrowed some of the code from the Battle challenge
* Ran rubocop before each git commit to keep the code clean
* Added some CSS to make it look nicer

## Thoughts
* I'm happy with how this turned out but I know adding multiplayer and the other game types would have led to a lot more complexity

## Outstanding issues
* Did not attempt multiplayer or addition game type


app.rb
```diff
@@ -0,0 +1,46 @@
...
+  enable :sessions
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  before do
+    @player = Player.instance
+  end
+
```
bernardleanse, line 16: keeping controller skinny!! nice

app.rb
```diff
@@ -0,0 +1,46 @@
...
+  post '/submit' do
+    session[:choice] = params[:choice]
+    game = Game.new
+    session[:npc] = game.random
+    session[:result] = game.result(session[:choice], session[:npc])
+    @player.update_score(session[:result])
+    redirect '/result'
+  end
```
Matt-Warnock, line 37: Nice work! I like the score, nice touch.
Maybe, runing the `Game` object in resullt would allow you to have a skinnier contoler.

app.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    @player = Player.create(params[:player_1_name])
+    redirect '/play'
+  end
+
```
bernardleanse, line 25: clear code, it's creating a player from an input field (name="player_1_name") in a post form, then redirecting.

app.rb
```diff
@@ -0,0 +1,46 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/player'
+require './lib/game'
+
+class App < Sinatra::Base
```
Matt-Warnock, line 6: I would have given this a more descriptive name like RPGGame

lib/game.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def result(p1, p2 = random)
+    winner = { "rock" => "scissors", "paper" => "rock", "scissors" => "paper" }
+    if winner[p1] == p2
+      'WIN!'
+    else
+      (winner[p2] == p1 ? 'LOSE!' : 'DRAW')
+    end
+  end
+
```
bernardleanse, line 17: liking this game logic, very short and to the point

lib/game.rb
```diff
@@ -0,0 +1,18 @@
+class Game
+  def initialize
+  end
+
+  def random
+    ["rock", "paper", "scissors"].sample
+  end
+
+  def result(p1, p2 = random)
```
Matt-Warnock, line 9: I like how you've passed random in as default

lib/game.rb
```diff
@@ -0,0 +1,18 @@
+class Game
+  def initialize
+  end
+
+  def random
+    ["rock", "paper", "scissors"].sample
+  end
```
Matt-Warnock, line 7: This method is behaver that could be described as and 'NPC or computer player', so maybe not the same concern as the game functionality.
You could separate this method to a `NPC` class.

views/layout.erb
```diff
@@ -0,0 +1,14 @@
...
+
+<head>
+  <title>Rock Paper Scissors</title>
+  <link href="/styles.css" rel="stylesheet" type="text/css">
+</head>
+
+<body>
+<div class="content">
+<%= yield %>
+</div>
```
bernardleanse, line 12: Does this apply the backbone of the other views? I was thinking of how to make that.

bernardleanse (commented):
thought i'd do a review as i'm solo today

lib/game.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def result(p1, p2 = random)
+    winner = { "rock" => "scissors", "paper" => "rock", "scissors" => "paper" }
```
Matt-Warnock, line 10: Nice work here! A good way to avoid conditionals.
Maybe winner could be a constant instead. 
that would neaten up this method abit.

lib/game.rb
```diff
@@ -0,0 +1,18 @@
...
+    if winner[p1] == p2
+      'WIN!'
+    else
+      (winner[p2] == p1 ? 'LOSE!' : 'DRAW')
```
Matt-Warnock, line 14: You could do an earlier return at the top of this method for drawing like
`return 'LOSE!' if p1 == p2`
Then you could just use a guard clause  for the win, returning 'LOSE!' by default

spec/game_spec.rb
```diff
@@ -0,0 +1,25 @@
+describe Game do
+  subject(:game) { described_class.new }
+
+  describe '#result' do
+
+    it 'player 1 wins' do
+      p1 = "rock"
+      p2 = "scissors"
```
Matt-Warnock, line 8: very clean and easy to read, nice!

views/layout.erb
```diff
@@ -0,0 +1,14 @@
...
+</body>
+</html>
```
Matt-Warnock, line 14: Nice Layout!

PR #1965, opened 2021-10-10
RPS Challenge Sergei Palmer

s-palmer:
- Single player user story completed


public/css/app.css
```diff
@@ -0,0 +1,64 @@
+body {
```
arthurfincham, line 1: Nice work with the CSS 💪

views/game.erb
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
```
arthurfincham, line 1: There's a way to DRY these views up by using `layout.erb` - essentially you have a template for all your views and then you render them with `<%= yield %>` ! Will explain more

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def sign_in
+  visit ('/')
+  fill_in("name", with: "Sergei")
```
arthurfincham, line 3: Haven't seen this before - looks good

arthurfincham (commented):
On the whole it looks great! Only thing I'd do differently is templating the views

views/game.erb
s-palmer: Yeah I saw you using this in your file, looked interesting!

PR #1966, opened 2021-10-10
Bernard's RPS Challenge

bernardleanse:
Hectic weekend, but managed to get it working on a basic level. Did not quite get to implement a scoreboard, but was almost there. Some questions needing answering on capybara dependencies.


README.md
```diff
@@ -1,86 +1,16 @@
-# RPS Challenge
+## Bernard's RPS Challenge
 
-Instructions
--------
+I started by feature testing the ability to enter a name and it be displayed. This was achieved by test driving two new objects
```
blue-luke, line 3: Nice overview of your approach.

README.md
```diff
@@ -1,86 +1,16 @@
...
-  SimpleCov::Formatter::Console,
-  # Want a nice code coverage website? Uncomment this next line!
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
-
-You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
+In the time available I managed to get the thing to work and display the correct result, but when you go back and enter the same name again, the
+outcome is fixed for some reason. Didn't get time to fix this.
```
blue-luke, line 16: Good note

app.rb
```diff
@@ -0,0 +1,40 @@
...
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    erb(:main)
+  end
+
+  post '/login_name' do
+    Game.create_game
```
blue-luke, line 17: This seems like a lot of lines - there might be a way to do all these things in Game's initialise, keeping the controller thinner

lib/game.rb
```diff
@@ -0,0 +1,41 @@
+class Game
+  
+  def initialize
+    @players = []
+    @key_beats_val_hash = {
```
blue-luke, line 5: Nice way of representing the outcomes

lib/player.rb
```diff
@@ -0,0 +1,37 @@
+class Player
+  def initialize(name, bot: false)
```
blue-luke, line 2: This is a nice way of allowing the opponent to be either a human or a bot

spec/player_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  it "can assign itself random choice" do
+    bot.assign_rand 
+    expect(bot.choice).to eq('rock').or eq('paper').or eq('scissors')
+    
+  end
+
+  it 'can have a win added to it' do
+    bernard.add_win
+    expect(bernard.wins).to eq 1
```
blue-luke, line 28: This is nice extra functionality

blue-luke (commented):
I didn't manage to checkout the code and run it, but it looks well put together from the source files!

README.md
JohnForster: Your readme should always cover how to install and run your project.

lib/game.rb
```diff
@@ -0,0 +1,41 @@
...
+      'rock' => 'paper',
+      'paper' => 'rock',
+      'scissors' => 'paper'
+    }
+    @winner = nil
+  end
+
+  def self.create_game
+    @game = Game.new
+  end
```
JohnForster, line 15: `self` here means that this is a class method, but you're using it to set an _instance_ variable. So which instance contains this variable? The answer is quite complex, but it 100% isn't what you should be doing here.

app.rb
```diff
@@ -0,0 +1,40 @@
...
+    Game.instance.add_player(Player.new(params["Name"]))
+    Game.instance.add_player(Player.new("Computer", bot: true))
+    Game.instance
+    redirect "/play"
+  end
```
JohnForster, line 22: Game.instance is the wrong way to go about this. A better way would be to rewrite this as

```rb
game = Game.new
game.add_player(...)
game.add_player(...)
@game = game
redirect "/play"
```

This way you can get rid of the instance method in Game. If you're trying to ensure that there is only ever one instance of the game class at a time. I'd have a look at this article about singletons: https://www.rubyguides.com/2018/05/singleton-pattern-in-ruby/

lib/game.rb
```diff
@@ -0,0 +1,41 @@
+class Game
+  
+  def initialize
+    @players = []
+    @key_beats_val_hash = {
+      'rock' => 'paper',
+      'paper' => 'rock',
+      'scissors' => 'paper'
+    }
```
JohnForster, line 9: This is nice, but isn't individual to each instance. You can instead store it as a constant.

lib/player.rb
```diff
@@ -0,0 +1,37 @@
...
+    @name
+  end
+
+  def choice=(choice)
+    @choice = choice
+  end
+
+  def choice 
+    @choice
+  end
```
JohnForster, line 19: You can use the `attr_accessor` method to add both of these functions in at once.

lib/player.rb
```diff
@@ -0,0 +1,37 @@
...
+  def initialize(name, bot: false)
+    @name = name
+    @choice = nil
+    @bot = bot
+    @wins = 0
+  end
+
+  def name
+    @name
+  end
```
JohnForster, line 11: There is a shorthand `attr_reader` to do exactly this.

PR #1967, opened 2021-10-10
Rps-Challenge 

maxcfry:
Managed to complete most of this weekend challenge.


README.md
```diff
@@ -2,85 +2,34 @@
 
 Instructions
 -------
+To run program - follow these instructions:
```
t-kellett, line 5: very clear README, made reviewing this all very easy

spec/features/enter_names_feature_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'Enter names' do
```
t-kellett, line 1: Nice clear feature tests

app.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  get '/result' do 
+    erb :result
+  end
+
+  post '/result' do
+    @player_1_option = params[:player_1_option]
+    @option = ["Rock", "Paper", "Scissors"]
+    @bot_option = @option.sample
```
t-kellett, line 31: extracting this into a 'Bot' class as part of a model would have made the controller 'skinnier' in this case.

t-kellett (commented):
This is a clean and easy to understand implementation of the user stories. I think to improve generally either implementing some logic to define who 'wins' in the controller, or extracting that logic into a model would make it easier to define a winner and to change in future (say to add the 'lizard' rules).

PR #1968, opened 2021-10-10
Matt Warnock rps weekend challage

Matt-Warnock:
I aimed to have a skinny controller without the need for a presenter class.

With more time I would have done some CSS styling.


alexboyling (commented):
This was great Matt! I haven't left any comments in the code because it was all so well laid out and clear what was happening.

The only thing I'd suggest changing is using a single page rather than multiple pages depending on the win/lose/draw scenario.

lib/computer.rb
```diff
@@ -0,0 +1,13 @@
...
+class Computer
+  def random_gesture
+    %w[rock paper scissors][random_choice]
+  end
+
+  private
+
+  def random_choice
+    rand(0..2)
+  end
```
alicelieutier: This is quite obviously here to help you test but it's not great practice to stub the return of the class under test. You may want to look into `srand` instead, or stub the kernel `rand` method.

spec/computer_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      gesture = subject.random_gesture
+
+      expect(gesture).to eq 'scissors'
+    end
+  end
+
+  def random_to_return(number)
+    allow(subject).to receive(:random_choice).and_return(number)
+  end
```
alicelieutier: See my other comment.

alicelieutier (commented):
Really not much to say here, it's a great attempt, really good clear well organised code!
One comment on testing and randomness below.

PR #1972, opened 2021-10-10
RPS - Emmanuel 

emanfolo:


lib/computer.rb
```diff
@@ -0,0 +1,5 @@
+class Computer
```
dpkerins, line 1: I like that you separated your computer into a separate class -- I think I should have done this to separate out more from the classes managing the actual game .

lib/game.rb
```diff
@@ -0,0 +1,41 @@
...
+  def initialize(player = Player.new, cpu = Computer.new)
+    @player = player
+    @cpu = cpu
+    @cpu_turn = nil
+    @player_score = 0
+    @cpu_score = 0
+    @result = "Take a spin!"
+  end 
+
+  def turn(turn, _player = @player, cpu = @cpu)
```
dpkerins, line 13: having these parameters with defaults is smart, because you probably could have fit in a second player easily without changing much methods etc

styles.css
```diff
@@ -0,0 +1,8 @@
+body {
```
dpkerins, line 1: nice css work

views/index.erb
```diff
@@ -0,0 +1,50 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Document</title>
+  <style>
+  
+  body {
```
dpkerins, line 10: You could move this css into the css sheet you made

README.md
```diff
@@ -1,86 +1,5 @@
...
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
-  SimpleCov::Formatter::Console,
-  # Want a nice code coverage website? Uncomment this next line!
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
-
-You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
+Improvements I could have made would be to set a maximum score which redirects to a winner page when it is reached by the player or the CPU. I could have also included the option for 2 player mode, however I did not due to time constraints.
```
dpkerins, line 5: Love the way you added a point system. downloaded and played the game and it looks really good. I like the way you handled the pages and views so that completing a round doesn't take you from the game page, it just updates.

README.md
```diff
@@ -1,86 +1,5 @@
-# RPS Challenge
+I've complete the challenge of creating a web application in which a user can sign up with their name and play a game of RPS vs a CPU.
```
dpkerins, line 1: Code looks DRY and clear -- not difficult to read or understand. I think condensing your CSS into a style file would help and your Game.turn method could maybe be split into a second method for the result. Otherwise looks great to me.

PR #1977, opened 2021-10-11
RPS-Challenge

dpkerins:


views/play.erb
```diff
@@ -0,0 +1,7 @@
+<h1><%= @player_name %> VS. Computer</h1>
+<br><br>
+<form action="/moves" method="post">
```
emanfolo, line 3: Very interesting use of each loop to include multiple input fields, definitely going to steal this!

views/results.erb
```diff
@@ -0,0 +1,10 @@
+<h1><%= @game.player.name %> played <%= @game.player_move.name%></h1>
+<h1>The Computer played <%= @game.computer_move.name%></h1>
+<% if @output == "tie" %>
```
emanfolo, line 3: I like the way you've included ruby logic in the erb file to account for different outputs based on the result

lib/game.rb
```diff
@@ -0,0 +1,53 @@
...
+    return "The Computer" if computer_win?
+    return @player.name
+  end
+
+  def tie?
+    player_move == computer_move
+  end
+  
+  def computer_win?
+    false unless @player_move.predators.include?(computer_move.name)
```
emanfolo, line 51: I like the way you've detailed the opposition moves that would beat each turn the user makes. Definitely makes it clearer to the reader.

lib/game.rb
```diff
@@ -0,0 +1,53 @@
...
+    @choices.sample
+  end
+
+  def output
+    return "tie" if tie?
+    return "The Computer" if computer_win?
+    return @player.name
+  end
+
+  def tie?
```
emanfolo, line 46: Slim methods that have a clear and singular responsibility, easy to understand

spec/features/play_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'play game' do
+  scenario 'see move choices' do
+    add_name_start
+    expect(page).to have_button 'ROCK'
+    expect(page).to have_button 'PAPER'
+    expect(page).to have_button 'SCISSORS'
```
emanfolo, line 6: Could have split these into multiple tests

emanfolo (commented):
Very well written

PR #1978, opened 2021-10-11
Weekend RPS challenge  - Code review

Kntanos:
Rspec is still a struggle to me.


README.md
```diff
@@ -1,86 +1,21 @@
...
+This app is built using:
+Ruby
+Sinatra
+Rspec
+Capybara
+HTML
+
+I started with a sequence diagram which I refined towards the end to the following:
+
+![](./sequence_diagram.jpg)
```
TTurvey, line 12: Nice diagram - it makes the process easy to understand, well done.

views/result.erb
```diff
@@ -0,0 +1,17 @@
+<h2>You played <%= session[:button] %></h2>
+<br><br>
+<h2>Computer played <%= Game.instance.computer_choice %></h2>
+<br><br>
+
+<h1><%= Game.instance.play %></h1>
```
TTurvey, line 6: It seems by using the 'self' methods within the 'game.rb' file, you've managed to keep the majority of the ruby logic within that 'game.rb' file instead of having to include the if statements within 'results.erb'.
Perhaps this means that if more HTML or CSS was implemented here then it would actually be easier to read.

spec/game_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'game'
+
+describe Game do
+  describe "#play" do
+    let(:player_choice) { "Paper" }
+    
+    it "should declare the correct winner" do
+      # I tried everything.... I give up
```
TTurvey: As discussed the issue here is that the computer's choice is randomised and so you would have to use a mock scenario to fix the computer's choice in order to test a specified outcome. I think this is one for the coaches to help with as I am also unsure how to implement this.

views/result.erb
```diff
@@ -0,0 +1,17 @@
+<h2>You played <%= session[:button] %></h2>
+<br><br>
+<h2>Computer played <%= Game.instance.computer_choice %></h2>
+<br><br>
+
+<h1><%= Game.instance.play %></h1>
+
+Do you want to play again?
```
TTurvey, line 8: This is a nice feature to have

TTurvey (commented):
The logic used within your code is clear which made it easy to understand. I like that it is contained within the ruby files only. That is something I'm taking away from this review for myself. Make sure you follow up on how to implement the mock aspect of the rspec test. This will likely be a useful skill for future challenges. Overall well done!

app.rb
```diff
@@ -0,0 +1,43 @@
...
+    session[:player_name] = params[:player_name]
+    redirect '/play'
+  end
+
+  get '/play' do
+    erb :play
+  end
+  
+  post '/play_round' do
+    session[:button] = params[:button]
```
eoinmakers, line 27: Do you need to store the chosen button in sessions if you're passing it straight into Game.create on the next line?

app.rb
```diff
@@ -0,0 +1,43 @@
...
+    @game = Game.create(session[:button])
+    redirect '/result'
+  end
+
+  post '/play_again' do
+    session[:button] = params[:button]
+    redirect '/play' if session[:button] == "Yes"
```
eoinmakers, line 34: It would be slightly cleaner to just use a button with an action to direct the user to either play or / here.

views/play.erb
```diff
@@ -0,0 +1,14 @@
+<h2><%= session[:player_name] %> what is your choice?</h2>
```
eoinmakers, line 1: I would store this in an instance variable in the route so you can reference it directly instead of calling sessions in the View - this is to keep more of a separation of concerns between the Model, View and Controller.

eoinmakers (commented):
This is a good RPS challenge, well done! 

If you're interested in one further improvement to the class definitions, you can check out the review doc here to see the benefit of drawing out one more object: https://github.com/Kntanos/rps-challenge/blob/main/docs/review.md#not-encapsulating-the-computer-in-a-separate-class

For testing, I would see this being predominately feature tested, but for unit testing in RSpec, the only other recommendation I would have is fully testing every possible outcome, not just one combination of choices.

PR #1979, opened 2021-10-11
RPS Challenge

marazzo:
There are no tests for this and it doesn't work past the play page.
I had it working and then I started refactoring and lost functionality. My mistake was not pushing to GitHub prior to refactoring.


app.rb
```diff
@@ -0,0 +1,44 @@
...
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    $player_1 = Player.new(params[:player_1_name])
```
charlie83xt, line 18: I was wondering whether the use of global variables at this point is strictly necessary or instance variables could have been used instead. Maybe using global ones is a simpler and better approach in this particular case. However, due to the recommendation of not using these on hiring tech test, wondering what would be the moment to use them.

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def computer_choice
+    @pc_selection = @computer.random_option
+  end
+
+  def result
+    @weapon.judge(p1_selection, pc_selection)
+  end
+end
```
charlie83xt, line 34: Easy to understand code. the name of the variables is spot on. makes it easy to follow.

lib/weapon.rb
```diff
@@ -0,0 +1,22 @@
...
+  RULES = { 
+    rock: :scissors, 
+    paper: :rock, 
+    scissors: :paper 
+  }
+
+  def judge(p1_choice, p2_choice)
+    return :Draw if draw(p1_choice, p2_choice)
+    return :Win if beats(p1_choice, p2_choice)
+    return :Lost if !beats(p1_choice, p2_choice)
```
charlie83xt, line 12: What a great approach. I couldn't have imagined this way.

views/index.erb
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE hmtl>
+<html>
+  <body>
+    <div id="header">
+        <h1>Rock, Paper, Scissors</h1>
+        <h1>Are you ready to defy your luck?</h1>
```
charlie83xt, line 6: We can change to any other similar game and still use this intuitive introduction.

charlie83xt (commented):
Short review of an excellent approach and simple game coded by Marianne Azzopardi.

by Charlie.

PR #1985, opened 2021-11-07
Rock, Paper, Scissorsn game

sigii1992:
I covered both user stories and the app is working as it supposed to work. If I had more time I would use CSS to make my app beautiful.


Will-Birdseye:
Working code that meets all the supplied user stories!
Folder structure is appropriate and well organised. Files are named consistently with all relevant frameworks initialised as needed to run tests and use capybara.
Feature tests and unit tests are detailed and cover possible edge cases.
Game class features strings that could Instead, return representative codes, such as :win and :draw from the Game class which can be translated by the presentation layer.
Game logic could be converted to use of hash to map rules to avoid if/else statements.
Use of separate class for Computer implemented successfully with global variables eliminated well!
Choices could be switched over to constants rather than strings.
DRY code principles integrated regularly and successfully!

lib/game.rb
```diff
@@ -0,0 +1,45 @@
+require './lib/player'
+require './lib/computer'
+
+class Game
+  attr_reader :players, :result
+  CHOICES = ['Rock', 'Paper', 'Scissors']
```
Will-Birdseye, line 6: Review rubric suggests switching choices to constants e.g.

lib/game.rb
```diff
@@ -0,0 +1,45 @@
...
+    @players.last
+  end
+
+  def play(p_choice)
+    player.player_choice(p_choice)
+    computer.random_choice
+  end
+
+  def result
+    if (player.choice == computer.choice)
```
Will-Birdseye, line 35: Review rubric suggests: Use a hash to map the rules e.g.
RULES = { rock: :scissors,
          paper: :rock,
          scissors: :paper }

lib/game.rb
```diff
@@ -0,0 +1,45 @@
...
+      @result = "It is a draw!"
+    elsif (player.choice == 'Rock' && computer.choice == 'Scissors') ||
+      (player.choice == 'Scissors' && computer.choice == 'Paper') ||
+      (player.choice == 'Paper' && computer.choice == 'Rock')
+      @result = "You WON!"
+    else
+      @result = "You LOST!"
```
Will-Birdseye, line 42: Check Step 3 of review rubric for advice on removing strings from results for greater separation of concerns.

views/index.erb
```diff
@@ -0,0 +1,18 @@
+<div>
+  <body style="background-color:Tomato;">
```
Will-Birdseye, line 2: Big fan of Tomato!

Will-Birdseye (commented):
Superb code overall! Meets all user stories and is presented in an understandable way. 

Check review rubric for line comments to hopefully make more sense :)

app.rb
```diff
@@ -0,0 +1,40 @@
...
+    computer = Computer.new
+    @game = Game.create(player1, computer)
+    redirect '/play'
+  end
+
+  post '/winner' do
+    @game = Game.instance
+    choice = params[:rps]
+    @player_choice = @game.player.player_choice(choice)
+    @computer_choice = @game.computer.random_choice
```
alicelieutier, line 34: These two lines make me feel like your classes are not entirely encapsulated as you are reaching into the player within and the computer within the game. I think it would be cleaner to have your controller only speak with the game class, and then the game class can delegate to the player and computer class.

lib/computer.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class Computer
+  attr_reader :name, :choice
+
+  def initialize(name='Computer')
+    @name = name
+    @choice = nil
+  end
+
+  def random_choice
```
alicelieutier, line 11: I would maybe change the name of this method. Something like "choose" or "play". This is to convey the idea that the internal state is changed, as opposed to it returning a value that is interesting to the caller. See my article about naming things: https://medium.com/swlh/an-opinionated-guide-to-naming-your-code-aimed-at-new-developers-6f769d60e098

lib/computer.rb
```diff
@@ -0,0 +1,14 @@
...
+    @choice = Game::CHOICES.sample
```
alicelieutier, line 12: Nice use for a class constant 👍

lib/game.rb
```diff
@@ -0,0 +1,45 @@
...
+    @game = Game.new(player, computer)
+  end
+
+  def self.instance
+    @game
+  end
+
+  def initialize(player, computer)
+    @players = [player, computer]
+    @result = ""
```
alicelieutier, line 18: Should result be an instance variable that is part of the state, or should it be a method, that calculates its return based on the rest of the state? By adding this here, you are duplicating information in the state which makes it more likely to introduce bugs.

lib/game.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def player
+    @players.first
+  end
+
+  def computer
+    @players.last
+  end
```
alicelieutier, line 27: Do you need these?

lib/game.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def play(p_choice)
```
alicelieutier, line 29: This method is the method that should be called by the controller instead of the controller calling the player and computer directly.

lib/game.rb
```diff
@@ -0,0 +1,45 @@
...
+    player.player_choice(p_choice)
+    computer.random_choice
+  end
+
+  def result
```
alicelieutier, line 34: This method should just return the result instead of setting an instance variable.

lib/game.rb
alicelieutier: Agreed with this one. You don't have to, but it can be nice to separate the type of result (win, loss, draw), from the presentation to the user, like the string used, what's shown on the page, etc. You could even have different views for different outcomes.

lib/player.rb
```diff
@@ -0,0 +1,15 @@
...
+  attr_reader :name, :choice
+
+  def initialize(name)
+    @name = name
+    @choice = nil
+  end
+
+  def player_choice(choice)
+    # capitalized_choice = choice.downcase.capitalize
+    @choice = choice if Game::CHOICES.include?(choice)
```
alicelieutier, line 13: What happens if it does not include it?

alicelieutier: Maybe throw and error?

spec/computer_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+describe Computer do 
+   let(:computer) { Computer.new }
+
+   it "shows the players name" do
+      expect(computer.name).to eq 'Computer'
+   end
+
+  describe '#random_choice' do
+   it 'should choose an option from the list' do
+     allow(computer.random_choice).to receive(:sample)
```
alicelieutier, line 12: If you are going to stub sample here, you may as well test fo a fixed result.

spec/features/choose_option_spec.rb
```diff
@@ -0,0 +1,19 @@
+feature "Choose option" do
```
alicelieutier, line 1: Nice feature tests

alicelieutier (commented):
Good clean code and separation between MVC. You can improve the design within the model to make your classes even more encapsulated.
Good tests!

PR #1987, opened 2021-11-07
Mark - Rock, Paper, Scissors challenge

Cloudy1986:
Both user stories are complete. I ran out of time to look at the bonus levels. There are some issues with my tests I'm aware of, particularly lack of coverage where the content rendered in views is dependent on a randomly generated variable from the model.


README.md
```diff
@@ -1,86 +1,48 @@
...
-## Bonus level 1: Multiplayer
-
-Change the game so that two marketeers can play against each other ( _yes there are two of them_ ).
-
-## Bonus level 2: Rock, Paper, Scissors, Spock, Lizard
+Test coverage
+-----
 
-Use the _special_ rules ( _you can find them here http://en.wikipedia.org/wiki/Rock-paper-scissors-lizard-Spock_ )
+99.32%
```
Court534, line 44: Nice clear and easy to read/understand README.MD file.

README.md
```diff
@@ -1,86 +1,48 @@
...
-  # Want a nice code coverage website? Uncomment this next line!
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
 
-You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
+Time spent on challenge
+-----
+Sunday 10am to 7:30pm
```
Court534, line 48: Nice clear and easy to read/understand README.MD file. I like that you included the time you spent working on the challenge at the bottom.

app.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  get '/choice-made' do
+    @game = $game
+    @game.computer.random
+    erb(:choice_made)
+  end
+
+  # start the server if ruby file executed directly
+  run! if app_file == $0
+end
```
Court534, line 49: Code layout is very neat and easy to read.

app.rb
```diff
@@ -0,0 +1,49 @@
...
+  enable :sessions
+
+  get '/' do
+    erb(:index)
+  end
+
+  post '/names' do
+    player_1 = Player.new(params["player_1_name"])
+    computer = Computer.new
+    $game = Game.new(player_1, computer)
```
eoinmakers, line 21: Are these instances all needed here? 

If /names only responsibility is to get the players name, maybe a lighter touch approach would store the player name param directly in a session. Computer can be called later alongside Game once you need the game rules to be called, but it seems they don't have a lot of use here.

app.rb
```diff
@@ -0,0 +1,49 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    @game = $game
```
eoinmakers, line 26: Related to the above, if the only information you require from the Game instance is the player name, sessions is a lighter touch way to share this information with this route.

app.rb
```diff
@@ -0,0 +1,49 @@
...
+    erb(:play)
+  end
+
+  get '/choose' do
+    @game = $game
```
eoinmakers, line 31: As above

app.rb
```diff
@@ -0,0 +1,49 @@
...
+  post '/choice' do
+    @game = $game
+    @game.player_1.update_choice(params["choice"])
+    redirect '/choice-made'
+  end
+
+  get '/choice-made' do
+    @game = $game
+    @game.computer.random
+    erb(:choice_made)
```
eoinmakers, line 44: To me, this route is the crux of the use for Game - its where both the players and computers choices would be made visible, alongside the winner, all of which is contained within the Game class (instantiated with the result from the Computer instance). 

Containing your use of Computer and Game to this route makes the implementation slightly safer, as its not exposing more information than it needs to with the declaration of Game in previous routes, and the use of a global variable.

lib/computer.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  attr_reader :name, :move
+
+  def initialize
+    @name = "Computer"
+    @move = ["Rock", "Paper", "Scissors"]
+  end
+
+  def random
+    @move = @move.sample
```
eoinmakers, line 11: You could make this method more self-contained / changeable by writing it as: 

```
def generate_random_move
    ["Rock", "Paper", "Scissors"].sample
end
```

I also renamed the method to a verb / action, as is best practice for method naming. We like to give them actions describing what responsibility it is they're carrying out.

lib/game.rb
```diff
@@ -0,0 +1,25 @@
+class Game
+
+  def initialize(player_1, computer)
+    @players = [player_1, computer]
```
eoinmakers, line 4: I think this instance doesn't necessarily need to know all of the information inside player_1 and computer. 

It needs to know the player's choice, and the computer's choice, but not much else. Additionally, it's only the 'decide_winner' method that needs to know these pieces of information, so some refactoring could be useful in passing the move choices directly into decide_winner as arguments, instead of passing the whole instances in.

eoinmakers (commented):
Good submission, well done. 

I left some notes around refactoring in order to expose less information where possible. 

Additionally, you could unit test each possible outcome combination (for each draw, win, loss, etc). 

Nice work!

PR #1989, opened 2021-11-07
Ginny - RPS Challenge

GinnyThomas:
6 examples, 0 failures
COVERAGE: 100.00% -- 36/36 lines in 6 files

Game is complete in simplest form.  I have 3 TODO notes in one of the view specs.  This allows me to find areas I would want to improve quickly in my compiler.  

If I had more time and was better at it, I would improve the front end design.  
I would also like to add an option after winner is announced to either "Play again" or "Quit"

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,18 @@
+require 'sinatra'
+# require 'sinatra/reloader'
+
+
+class Game < Sinatra::Base
+  configure :test do
+  end
+  # enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    @player_name = params[:player_name]
+    erb :play
+  end
+end
--- a/app.rb
+++ b/app.rb
@@ -1,8 +1,10 @@
 require 'sinatra'
 require 'sinatra/reloader'
+require './lib/rules'
+require './lib/element'
+require './lib/game'
 
-
-class Game < Sinatra::Base
+class RPS < Sinatra::Base
   configure :development do
   end
 
@@ -19,11 +21,13 @@ class Game < Sinatra::Base
 
   get '/start' do
     @player_name = session[:player_name]
+
     erb :name
   end
 
   get '/play' do
     @player_name = session[:player_name]
+    $game = Game.new
     erb :play
   end
 
--- a/app.rb
+++ b/app.rb
@@ -31,5 +31,16 @@ class RPS < Sinatra::Base
     erb :play
   end
 
+  post '/who_is_winner' do
+    session[:selected_option] = params[:element]
+    redirect '/winner'
+  end
+
+  get '/winner' do
+    @player_name = session[:player_name]
+    @selected_option = session[:selected_option]
+    erb :winner
+  end
+
   run! if app_file == $0
 end
--- a/app.rb
+++ b/app.rb
@@ -14,7 +14,12 @@ class Game < Sinatra::Base
 
   post '/name' do
     session[:player_name] = params[:player_name]
-    redirect '/play'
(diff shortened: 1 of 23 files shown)


app.rb
```diff
@@ -0,0 +1,49 @@
...
+  configure :development do
+  end
+
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
```
katemyoung, line 17: Appreciate: Each discrete action of your programme has it's own dedicated route, they are not dual purpose [(code review rubric)](https://github.com/makersacademy/rps-challenge/blob/main/docs/review.md#inconsistent-routing-and-route-naming )

lib/game.rb
```diff
@@ -0,0 +1,24 @@
...
+    @rules = Rules.new
+    computer_move_options = @rules.elements.map { |h| "#{h.name}"}
+    computer_move_options.sample
+  end
+
+  def determine_winner(player_selection, computer_choice)
+    result = 'something is wrong'
+    element = @rules.elements.select { |element| element.name == player_selection}.first #find element whose name matches player selection
+    if element.beats == computer_choice
+      result = "wins"
```
katemyoung, line 16: What do you think of refactoring so that presentation strings are removed from the business logic layer (open/closed principle). [(code review rubric)](https://github.com/makersacademy/rps-challenge/blob/main/docs/review.md#including-presentation-strings-in-business-logic-layer)

katemyoung (commented):
Thanks for the opportunity to review your code, Ginny. I was able to check it out and play it. I ran the tests and they all passed with 100% coverage. 

Your files are well structured and named [(code review rubric)](https://github.com/makersacademy/rps-challenge/blob/main/docs/review.md#inconsistent-folder-layout-sinatra-structure) and your spec_helper.rb is set up correctly.

Have you considered adding instructions in the README.md [(code review rubric)](https://github.com/makersacademy/rps-challenge/blob/main/docs/review.md#instructions-in-readme) For example, I initially tried to run the code with rackup then realised it was not configured in the app.

app.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  get '/start' do
+    @player_name = session[:player_name]
+
+    erb :name
+  end
+
+  get '/play' do
+    @player_name = session[:player_name]
+    $game = Game.new
```
eoinmakers, line 30: Does this need to be called here? Looks like its not being used - nice work in using sessions to store the player name!

app.rb
```diff
@@ -0,0 +1,49 @@
...
+    session[:selected_option] = params[:element]
+    redirect '/winner'
+  end
+
+  get '/winner' do
+    @game = $game
+    @player_name = session[:player_name]
+    @selected_option = session[:selected_option]
+    @computer_option = $game.computer_selection
+    @winner = $game.determine_winner(@selected_option, @computer_option)
```
eoinmakers, line 44: You can maybe pass these values into determine_winner directly instead of storing them in instance values. Additionally, if you declared your instance of Game here instead of in '/play', you could avoid using a global variable.

eoinmakers (commented):
Successful implementation of sessions, nice work. 

Overall I think this is a good RPS challenge attempt, well done. 

I found the logic to be a little overcomplicated in terms of handling the game logic itself - the rules that define whether a player wins, draws or loses. I think this can be handled in less classes in order to make the code a little more easily readable. 

Additionally, one small refactor in your controller class would mean you could avoid the use of global variables. 

For unit testing this project, I mostly see unit testing the game logic outcomes (rock vs paper, who wins, for example) as being comprehensive enough.

PR #1992, opened 2021-11-08
Rock Paper Scissors

ayseldilan:
I was able to make a submit form of players name and use srand to play against AI. Declared a win output.


ayseldilan:
rps

views/index.erb
```diff
@@ -13,6 +13,7 @@ p {
   font-family: courier;
   font-size: 160%;
 }
+
```
Doubleo-o14, line 16: nice use of the font style and size

PR #1998, opened 2021-11-08
Add players name

willjshark:
First functionality requirement added. Players can enter their name before playing.


app.rb
```diff
@@ -0,0 +1,14 @@
+require 'sinatra/base'
+class RPS < Sinatra::Base
+  
+  get '/' do
+    erb(:index)
+  end
+
+  post '/name' do
+    @player = params[:player]
+    erb(:show)
```
RPGrimes, line 10: Good use of an instance variable as a best practice rather than using a global variable

spec/features/enter_name_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'Testing abilty to enter name' do
+
+  scenario 'Able to enter your name on home page' do
+    visit('http://localhost:4567/')
+    fill_in 'player', with: 'Will'
+    click_button 'Submit'
+    expect(page).to have_content 'Will'
```
RPGrimes, line 7: Nice robust test.

app.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
```
RPGrimes, line 12: As a question, what would your next step be if you'd had more free time this weekend?

PR #1999, opened 2021-11-08
RPS Challenge

katemyoung:
I have implemented the first user story which allows a player to register their name.

**To complete the project I would:**
- [ ] Add a method #play to the Game class in model game.rb, containig the logic for which moves beat which other moves.
- [ ] Develop the app.rb to use this logic in the /winner route.
- [ ] Refactor so that an instance of Player could be used by an instance of Game (not yet working)
- [ ] Store the move parameter that is posted from the /play route in an instance variable (not yet working).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,27 @@
-# RPS Challenge
+# Rock Paper Scissors Challenge
 
-Instructions
--------
+A web app that allows a player to play the game Rock, Paper, Scissors. 
+ 
+**Basic Rules**
+- Rock beats Scissors
+- Scissors beats Paper
+- Paper beats Rock
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+## How to play:
+*update this*
+Use git clone to create a local repo. 
+`git clone https://github.com/katemyoung/airport_challenge.git` 
 
-Task
-----
+Use bundler to install the required gems by running the command `bundle` in the project directory.
 
-Knowing how to build web applications is getting us almost there as web developers!
+Navigate to the local repo then run irb.
+`irb -r './lib/dishes.rb'`
 
-The Makers Academy Marketing Array ( **MAMA** ) have asked us to provide a game for them. Their daily grind is pretty tough and they need time to steam a little.
+Run `rspec` from the project directory to run the tests.
 
-Your task is to provide a _Rock, Paper, Scissors_ game for them so they can play on the web with the following user stories:
+**You can use this programme in irb to:**
+
+## User Stories:
 
 ```
 As a marketeer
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -27,60 +52,70 @@ So that I can enjoy myself away from the daily grind
...
+
+- [ ] What does this do in the app.rb: `run! if app_file == $0`
+- [ ] Should feature tests be in the same or separate files?
+- [ ] I could not make this code work: 
+``` ruby
+it 'has clickable button' do
+ expect(page).to have_button('Rock')
+end
+```
+- [ ] What's the good spec guide say about class names, where instance methods are #method_name
```
GinnyThomas, line 121: I like all the notes you have written in the readme file.  It seems like a good way to be able to go back and see where you struggled, see what was successful, and learn from mistakes.

views/winner.erb
```diff
@@ -0,0 +1,4 @@
+<div>
+Your move: <%=$player1.name%> Computer's move: <%=@player2_move%>
```
GinnyThomas, line 2: This may be easier to read when it's displayed if there is some punctuation

views/winner.erb
```diff
@@ -0,0 +1,4 @@
...
+<%=@winner%> wins!
```
GinnyThomas, line 3: where is @winner being generated from?  I don't believe this code will display the name.  Is this part of the functionality you were referring to in the readme that you want to get working?

spec/features/feature_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    register_name
+    expect(page).to have_content('Make your move: Rock or Paper')
+  end
+end
+
+feature 'Winning or losing' do
+  let(:computer_move) { 'Rock' }
+  scenario 'player wins with paper' do
+    register_name
+    click_button 'Rock'
```
GinnyThomas, line 25: should this be 'Paper' ?

models/player.rb
```diff
@@ -0,0 +1,8 @@
+class Player
+  attr_reader :name, :move
+
+  def initialize(name)
+    @name = name 
+    @move = nil
+  end
+end
```
GinnyThomas, line 8: nice, clean code, easy to follow

models/game.rb
```diff
@@ -0,0 +1,8 @@
+class Game
+  attr_reader :player1
+  
+  def initilize(player1)
+    @player1 = player1
+  end
+
+end
```
GinnyThomas, line 8: clean code.  
what is this class doing?  It does not appear to be anywhere else in the code/app.  Was it setup for further functionality down the road?

GinnyThomas (commented):
All gems were correct - I was able to clone the git and run the rspec file and rackup to be able to see the app in motion.

README.md
katemyoung: Thanks, I find it really helpful for learning.

spec/features/feature_spec.rb
katemyoung: Yes! Thanks, you are right, it should.

PR #2003, opened 2021-11-08
Completed first user story in rps challenge

blessingaliu:


spec/features/presented_choices_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'Playing the game' do 
+    scenario 'presented choices: rock, paper or scissors' do
+        visit('/')
```
mo-codes1, line 3: put lines 3, 4 and 5 into their own web_helpers file so you follow the DRY method.

PR #2011, opened 2021-12-05
RPS Challenge Week 3

EviePalaiochorinou:
I have completed both user stories (I believe!)
Definitely, more tests are needed to expand coverage.
I would love test suggestions/guidance.

### DIFF
--- a/spec/game_spec.rb
+++ b/spec/game_spec.rb
@@ -1,0 +1,18 @@
+require 'game'
+
+describe Game do
+
+    before do
+        @game = Game.new("Evie")
+        allow(@game).to receive(:pick_choice).and_return("Rock")
+    end
+
+    it 'checks if the game picks a random choice' do
+        expect(@game.pick_choice).to eq("Rock")
+    end
+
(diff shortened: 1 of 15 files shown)


spec/game_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'game'
+
+describe Game do
+
+    before do
+        @game = Game.new("Evie")
+        allow(@game).to receive(:pick_choice).and_return("Rock")
```
tc1316, line 7: Good use of "allow" to fix the state of the Game choice

spec/game_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+
+    it 'checks if the game picks a random choice' do
+        expect(@game.pick_choice).to eq("Rock")
```
tc1316, line 11: Suggestion: If you are unit testing for random, you could assert that the game.pick_choice should output something that is included within an array of choices.

E.g. is game.pick_choice output in [Rock,Paper, Scissors]?

PR #2012, opened 2021-12-05
Rock Paper Scissors Weekend Challenge

jessgordon:


Public/style.css
```diff
@@ -0,0 +1,75 @@
+body {
+font-family: 'Fruktur', cursive;
+background-color:#dddddd;
+color: peru;
+text-align: center;
+}
+
```
msc49, line 7: fantastic use of CSS. Creative and detailed :)

lib/player.rb
```diff
@@ -0,0 +1,11 @@
+class Player
```
msc49, line 1: Good use of Single responsibility

spec/features/game_page_spec.rb
```diff
@@ -0,0 +1,69 @@
+feature 'user(s) can play a game of rock paper scissors' do
+
+  feature 'user can play a game of rock paper scissors with the computer' do
+
+    before do
```
msc49, line 5: great way of using the before method in TDD. Maybe 'helpers' would also be another nice way

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+    @mode = mode
+  end
+
+  def update_move(move)
+    one_player_update(move) if mode == '1 player'
+    two_player_update(move) if mode == '2 player'
+  end
+
+  def return_winner
+    draw ? "It's a draw!" : player1_wins ? "#{@players[0].name} wins!" : "#{@players[1].name} wins!"
```
msc49, line 15: genuienly, your code is very efficient and nicely written

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def draw
+    @players[0].move == @players[1].move
+  end
+
+  def player1_wins
+    p1 = @players[0].move[0]
+    p2 = @players[1].move[0]
+    (p1 == "R" && p2 == "S") || (p1 == "P" && p2 == "R") || (p1 == "S" && p2 == "P")
```
msc49: is there any particular reason why you shortne the strings? For instance "R" instead of "Rock"

notes.md
```diff
@@ -0,0 +1,13 @@
+- how do i test the generate_move method? or can I assume the sample method works as it's built in to the array?
```
msc49: potwntially srand? ot you can use the include rspec to check

spec/game_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'game'
+
+describe Game do
+  let(:player1) { double('player1', :name => 'player1', :move => 'ROCK') }
```
msc49, line 4: fantastic!

views/layout.erb
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <link rel="stylesheet" href="<%= url('/style.css') %>" type="text/css" />
+    <script src='https://kit.fontawesome.com/a076d05399.js' crossorigin='anonymous'></script>
```
msc49, line 5: genuinely cant find any issues with your rspec files or erb files

msc49 (commented):
crazy impressive work. Thoroughly enjoyed looking at you work

spec/features/game_page_spec.rb
jessgordon: good shout! I'll do that now, thanks :)

lib/game.rb
jessgordon: thank you that's so encouraging 😁

lib/game.rb
jessgordon: To be honest I just wanted it to fit on one line without it being too long - but I think it might actually read better and be easier to understand if I left them as whole strings and just used three lines 😅

views/layout.erb
jessgordon: Thanks so much for your feedback! I've made a few tweaks thanks to your review :)

app.rb
```diff
@@ -0,0 +1,48 @@
...
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    @@game = nil
+    erb(:index)
+  end
+
+  post '/mode' do
+    @mode = params[:mode]
```
leoht, line 17: Minor one — since the `mode` is mainly used as a flag to either present either the 1-player form or the 2-player form (if I'm correct!), it's probably fine to implement it as a `GET` parameter on the index route rather than a `POST` endpoint

lib/game.rb
```diff
@@ -0,0 +1,42 @@
+class Game
+  attr_reader :players, :mode
+
+  def initialize(player1, player2, mode)
+    @players = [player1, player2]
+    @mode = mode
+  end
+
+  def update_move(move)
+    one_player_update(move) if mode == '1 player'
```
leoht, line 10: You could use constants to store the possible `mode` string values, it's usually better to keep enumerated or configuration values, and make sure there are no typos, etc 
```ruby
class Game
  MODE_1_PLAYER = '1 player'

  # ...later
 def update_move(move)
    one_player_update(move) if mode == MODE_1_PLAYER
# ...
```

lib/game.rb
leoht: Here as the method is named `return_winner` (we could simplify it to `winner`), it makes sense to simply return the *object* representing the player (either `@players[0]` or `@players[1]`). This is for two reasons:
 * The caller of `.return_winner` (or `.winner`) can then decide if it wants to print the name, or do something else with it — we're returning the raw data rather than a string
 * In terms of separation of concerns and the MVC pattern, it's better to move any presentation/display code out of `Game` (this class only deals with data of the game and its logic), and make the view/template responsible for it. Because `"It's a draw!"` is a human readable message, it makes sense to move it to the view layer rather than being returned here :) 

The code in the controller and view would only change slightly:
```
# In app.rb (in the result route)
@winner = @game.winner

# In result.erb 
<p><%= @winner.name %> wins!</p>
```

lib/player.rb
```diff
@@ -0,0 +1,11 @@
...
+  attr_reader :name
+  
+  def initialize(name)
+    @name = name
+  end
+
+  def move(move = @move)
```
leoht, line 8: This looks like it might not be intentional — assigning a default value for the `move` argument from the current instance variable `@move`?

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,32 @@
+require 'game'
+
```
leoht, line 2: Great use of helpers to simplify and DRY the code here :) 
A minor improvement (that's really a detail) could be to add arguments to them, so the helpers can be explicitly called with, for example, the player name:

```ruby
sign_in('Patos')
expect(page).to have_content 'Patos'

# is a bit more explicit that:
sign_in
expect(page).to have_content 'Patos'
```

spec/game_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  let(:player2) { double('player2', :name => 'player2', :move => 'PAPER') }
+  let(:player3) { double('player3', :name => 'player3', :move => 'PAPER') }
```
leoht, line 6: Good use of mocking with doubles here 👍

views/index.erb
```diff
@@ -0,0 +1,32 @@
...
+    </form>
+  <% elsif @mode == "2_player" %>
+    <i class='fas fa-ghost' style='font-size:24px'></i> 
+    <i class='fas fa-ghost' style='font-size:24px'></i> 
+    <form action='/names' method='post'>
+      <input type='text' name='player1' placeholder="Player1 name">
+      <input type='text' name='player2' placeholder="Player2 name">
+      <input type="submit" value="START">
+    </form>       
+  <% else %>
```
leoht, line 19: A minor nice UX improvement — maybe select a default mode to use from the first? And then allow to change it with a link/button :)

app.rb
```diff
@@ -0,0 +1,48 @@
...
+require_relative './lib/player'
+require_relative './lib/game'
+
+class RockPaperScissors < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    @@game = nil
```
leoht, line 12: I'd be tempted to create a method `@@game.reset` (that would reset the state in the instance) and explicitly call it, rather than setting it to `nil`, but I think that's more a matter of preference here!

lib/game.rb
```diff
@@ -0,0 +1,42 @@
...
+    two_player_update(move) if mode == '2 player'
+  end
+
+  def return_winner
+    draw ? "It's a draw!" : player1_wins ? "#{@players[0].name} wins!" : "#{@players[1].name} wins!"
+  end
+    
+  def generate_move
+    ['ROCK', 'PAPER', 'SCISSORS'].sample
```
leoht, line 19: Also a minor one — the random move logic could be moved to a class `Computer` that would behave (implement the same methods) as the `Player` class. Expect that when calling `.move`, it would set a random choice. This way you can have the same interface and call the same method on both players, no matter if they're human or not

leoht (commented):
Great job! I only had a few minor comments on improvement, mainly on separating concerns a bit more (the `winner` method not returning a human readable message to let this job to the view, and maybe moving the random logic out of `Game`). Otherwise the tests are well-written, thorough and make good use of mocking :) 
(and then a few other minor things)
If you have any question on the feedback, let me know!

lib/game.rb
jessgordon: this makes SO much sense, thanks for spotting and explaining in such detail :)

app.rb
jessgordon: ah I see - that would clean up the routes and keep it DRY,  I agree 👍

lib/player.rb
jessgordon: I used this method to initially save the move as an instance variable and later to only return the instance variable already saved i.e.

User chooses move:
```
player.move('rock')
```
>> move method saves `@move` to `'rock'`


Model needs to know what move the user chose:
```
player.move
``` 
``` 
def move(move = 'rock')
    @move = 'rock'
end
``` 
>> move method  returns the pre-saved `@move` instance variable. 
  
I used this instead of two methods (assign and return saved move):
  ```
def assign_move(move)
  @move = move 
end
  
def move
  @move
end
```
  
  But perhaps this way would keep it more clear and readable?

spec/features/web_helpers.rb
jessgordon: yessss another nice addition thank you😁

app.rb
jessgordon: I agree actually, just wouldn't have thought of it :)

lib/game.rb
jessgordon: love it! Will trial this over xmas break thank you

PR #2015, opened 2021-12-06
RPS challange

Alaka-K:
I wasn't able to do the bonus levels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,11 @@
 # RPS Challenge
 
-Instructions
--------
+comments
+---------
+* I was only able to do the user stories and not the Bonus levels
+* On Rubocop it sas the condition size for outcome is too high. could I please get some guidence on this. Thank you.
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Task
-----
-
-Knowing how to build web applications is getting us almost there as web developers!
-
-The Makers Academy Marketing Array ( **MAMA** ) have asked us to provide a game for them. Their daily grind is pretty tough and they need time to steam a little.
-
-Your task is to provide a _Rock, Paper, Scissors_ game for them so they can play on the web with the following user stories:
-
+User Story
 ```
 As a marketeer
 So that I can see my name in lights
@@ -26,6 +15,7 @@ As a marketeer
 So that I can enjoy myself away from the daily grind
 I would like to be able to play rock/paper/scissors
 ```
+
 
 Hints on functionality
 
@@ -41,15 +31,7 @@ As usual please start by
 * Forking this repo
 * TEST driving development of your app
 
-[You may find this guide to setting up a new Ruby web project helpful.](https://github.com/makersacademy/course/blob/main/pills/ruby_web_project_setup_list.md)
 
-## Bonus level 1: Multiplayer
-
-Change the game so that two marketeers can play against each other ( _yes there are two of them_ ).
-
-## Bonus level 2: Rock, Paper, Scissors, Spock, Lizard
-
-Use the _special_ rules ( _you can find them here http://en.wikipedia.org/wiki/Rock-paper-scissors-lizard-Spock_ )
 
 ## Basic Rules
 
@@ -57,30 +39,3 @@ Use the _special_ rules ( _you can find them here http://en.wikipedia.org/wiki/R
 - Scissors beats Paper
 - Paper beats Rock
 
-In code review we'll be hoping to see:
-
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/main/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
-
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance may make the challenge somewhat easier.  You should be the judge of how much challenge you want this at this moment.
-
-Notes on test coverage
-----------------------
-
-Please ensure you have the following **AT THE TOP** of your spec_helper.rb in order to have test coverage stats generated
-on your pull request:
-
-```ruby
-require 'simplecov'
-require 'simplecov-console'
-
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
-  SimpleCov::Formatter::Console,
-  # Want a nice code coverage website? Uncomment this next line!
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
-
-You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
--- a/app.rb
+++ b/app.rb
@@ -1,5 +1,6 @@
 require 'sinatra/base'
 require 'sinatra/reloader'
+require './lib/computer'
 
 class RockPaperScissors < Sinatra::Base
   configure :development do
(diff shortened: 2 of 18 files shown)


lib/game.rb
```diff
@@ -0,0 +1,20 @@
...
+  def outcome(p_choice, c_choice)
+    if p_choice == c_choice
+      "This round is a draw!"
+    elsif p_choice == "rock" && c_choice == "scissors"
+      "You have won!"
+    elsif p_choice == "paper" && c_choice == "rock"
+      "You have won!"
+    elsif p_choice == "scissors" && c_choice == "paper"
+      "You have won!"
+    elsif c_choice == "paper" && p_choice == "rock"
```
kim-morgan, line 12: I think this can be made a bit more DRYer. After the elsif on line 10, I don't think any of the other elsifs are actually necessary - logically, you can just write else "Computer has won!" without explicitly stating each scenario

spec/game_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    it 'if player choses rock' do
+      expect(subject.outcome("rock", "scissors")).to eq("You have won!")
+    end
+  
+    it 'if player choses paper' do
+      expect(subject.outcome("paper", "rock")).to eq("You have won!")
+    end
+  
+    it 'if player choses scissors' do
+      expect(subject.outcome("scissors", "paper")).to eq("You have won!")
```
kim-morgan, line 19: Nice thorough tests!

app.rb
```diff
@@ -30,6 +31,7 @@ class RockPaperScissors < Sinatra::Base
   get '/selectedchoice' do
     @name = session[:name]
     @choice = session[:choice]
+    @computer_choice = Computer.new.choice
```
kim-morgan, line 35: It's good that you wrote your program to avoid using global variables!

README.md
```diff
@@ -1,22 +1,11 @@
...
-
-Knowing how to build web applications is getting us almost there as web developers!
-
-The Makers Academy Marketing Array ( **MAMA** ) have asked us to provide a game for them. Their daily grind is pretty tough and they need time to steam a little.
-
-Your task is to provide a _Rock, Paper, Scissors_ game for them so they can play on the web with the following user stories:
+comments
+---------
+* I was only able to do the user stories and not the Bonus levels
+* On Rubocop it sas the condition size for outcome is too high. could I please get some guidence on this. Thank you.
```
kim-morgan, line 6: I cloned your repo and had a go at running Rubocop. I think basically it's saying that your method has too many "if" statements. One way to reduce the if statements would be combining the "Computer wins" elsifs into one else - logically that would work since you already exhausted the possible opportunities for Player wins and draw, so the only other possible outcomes are computer wins!

app.rb
```diff
@@ -0,0 +1,40 @@
...
+    session[:choice] = params[:choice] 
+    redirect '/selectedchoice'
+  end
+
+  get '/selectedchoice' do
+    @name = session[:name]
+    @choice = session[:choice]
+    @computer_choice = Computer.new.choice
+    @outcome = Game.new.outcome(@choice, @computer_choice)
+    erb :selectedchoice
```
kim-morgan, line 37: Seems like a simple and easy-to-understand app.rb overall!

spec/computer_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+describe Computer do
+  subject { described_class.new }
+  describe "#choice" do
+    it "computer respond to choice" do
+      expect(subject).to respond_to(:choice)
+    end
+
+    it "computer choice is randomly chosen from array" do
+      choices = ["rock", "paper", "scissors"]
+      expect(choices).to include(subject.choice)
```
kim-morgan, line 12: Not sure if this method really tests whether the choice is randomly chosen... You could use something like srand(1), which basically forces random values to be generated in a certain way https://blog.appsignal.com/2018/07/31/generating-random-numbers-in-ruby.html

views/play.erb
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Rock, Paper, Scissors</title>
+    <link href="<%= url('/styles.css') %>" rel="stylesheet" type="text/css" />
+  </head>
+  <body>
+  <div class='container'>
+    <div class='center'>
+      <h1>Welcome <%=@name%> to a game of Rock, Paper, Sissors</h1>
```
kim-morgan, line 10: Small point but there's a spelling mistake in Scissors

kim-morgan (commented):
Here's my comments! I also cloned your repo and had a go running it in rackup and it all seems to work smoothly. Nicely done :)

lib/computer.rb
```diff
@@ -0,0 +1,6 @@
+class Computer
+
```
leoht, line 2: Good use of another class to separate the random choice from the game logic — that's a good example of separation of concerns :)

lib/game.rb
```diff
@@ -0,0 +1,20 @@
+class Game
+
+  def outcome(p_choice, c_choice)
+    if p_choice == c_choice
+      "This round is a draw!"
```
leoht, line 5: Nice! This could be made even better by replacing the human readable strings (e.g `"This round is a draw!"` by either specific values or symbols, such as `:win`, `:lost`, `:draw`

Why? 
 * For separation of concerns, it's better to let the `outcome` method returns the result, but *the view should decide what message to show* (the view is responsible for everything that is display or presentation to the user). By replacing those strings by "raw" data values, we can then move those human-readable messages to the view `selectedchoice.erb`, and then:
     * `Game`'s job is to return the game outcome
     * The .erb view's job is to *display* the right message depending on the outcome
 * It's easier then to test the result of `.outcome` if it's simpler value than plain English strings
 * If we later want to extend our app, and maybe translate the messages to another language, it's easier to change it with the views, we won't need to change our `Game` class at all :)

spec/computer_spec.rb
leoht: Something you could do as well, to test with enough confidence the value returned is *always* one of these 3, is to use a loop, let's say 20 times, and make sure each of these 20 times you call the method, the result is always part of these 3 possible values:
```ruby
20.times do
  # random choice every time
  expect(choices).to include(subject.choice)
end
```

spec/features/web_helper.rb
```diff
@@ -0,0 +1,20 @@
+def sign_in_and_play
```
leoht, line 1: Good to use a lot of helpers to keep the code DRY :) They can be made even better if we can pass arguments to them for e.g the player name, so it's more explicit when used in tests:
```ruby
sign_in_and_play('kenny')
# rather than
sign_in_and_play
```

spec/features/choice_spec.rb
```diff
@@ -0,0 +1,19 @@
+feature 'choice' do
+  scenario 'player choses paper' do
```
leoht, line 2: These feature tests are well-written, but we're missing one that would be good to have: testing the outcome of the game. For this, we could start with the same structure...
```ruby
 scenario 'player choses rock and computer choses scissors' do
    sign_in_and_play
    
    # mock the computer move here to be 'scissors'

    select_rock_and_submit

    # check the outcome message is correct (player should win)
  end
```

You can use method stubs in feature tests too to force predictable behaviour from the `Computer`'s `choice` method

views/index.erb
```diff
@@ -0,0 +1,22 @@
...
+    <link href="<%= url('/styles.css') %>" rel="stylesheet" type="text/css" />
+  </head>
+  <body>
+    <div class='container'>
+      <div class='center'>
+        <h1 class='title'>Would you like to play a game of Rock, Paper, Scissiors?</h1>
+        <h2 class='title'>Enter your name if you would like to play</h2>
+      </div>
+      <div class='center'>
+        <form action='/name' method='post'>
```
leoht, line 14: Minor one — the convention is usually to always use double quotes for HTML attributes, e.g `<div class="container">` rather than single quotes

leoht (commented):
Great job! :) The code is simple but works, I think you're on the right track to work on it further if you'd like at some point — I left a few comments about some minor things, better separation of concerns for the `Game` class outcome messages, and also how to make the feature tests more thorough by adding one that checks the outcome of the test (you can use method stubs in feature tests to achieve this as well).

Let me know if anything's unclear, or if you'd like to discuss it further

PR #2018, opened 2021-12-06
RPS Challenge

christopherxwebb:
I'm checking in a partial solution. So far, I have not been able to implement the logic that decides the outcome of the game.
I am unsure about how to make the different elements of the MVC speak to each. Some work ok: it's the final page that is the problem. Any thoughts or feedback would be greatly appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Task
 
 Knowing how to build web applications is getting us almost there as web developers!
 
-The Makers Academy Marketing Array ( **MAMA** ) have asked us to provide a game for them. Their daily grind is pretty tough and they need time to steam a little.
+The Makers Academy Marketing Array ( **MAMA** ) have asked us to provide a game for them. Their daily grind is pretty tough and they need time to let off steam a little.
 
 Your task is to provide a _Rock, Paper, Scissors_ game for them so they can play on the web with the following user stories:
 
--- a/README.md
+++ b/README.md
@@ -1,9 +1,6 @@
 # RPS Challenge
 
-I'm checking in a partial solution. So far, I have not been able to implement the logic that decides the outcome of the game.
-I am unsure about how to make the different elements of the MVC speak to each. Some work ok: it's the final page that is the problem. Any thoughts or feedback would be greatly appreciated.
-
-
+Solution complete.
 
 Instructions
 -------
--- a/README.md
+++ b/README.md
@@ -1,5 +1,10 @@
 # RPS Challenge
 
+I'm checking in a partial solution. So far, I have not been able to implement the logic that decides the outcome of the game.
+I am unsure about how to make the different elements of the MVC speak to each. Some work ok: it's the final page that is the problem. Any thoughts or feedback would be greatly appreciated.
+
+
+
 Instructions
 -------
 
--- a/app/rps.rb
+++ b/app/rps.rb
@@ -1,8 +1,11 @@
 require "sinatra/base"
 require "sinatra/reloader"
 require "./lib/player.rb"
+require "./lib/weapon.rb"
+require "./lib/computer.rb"
 
 class RPS < Sinatra::Base
+    enable :sessions
     configure :development do
       register Sinatra::Reloader
     end
@@ -11,17 +14,30 @@ class RPS < Sinatra::Base
         erb :index
     end
 
-    post '/names' do
-        $player_1 = Player.new(params[:player_1])           #This is a problem
-        # $player_2 = Player.new(params[:player_2])
-        redirect '/play'
+    post '/weapons' do
+        $player = params[:player_1]
+        $computer = Computer.new.compmove
+        erb :weapons
     end
     
-    get '/play' do
-        @player_1 = $player_1#.name
-        # @player_2 = $player_2#.name
-        erb :play
+    post '/result' do
+        $weapon = params[:p1move]
+        $computer = Computer.new.compmove
+       # $outcome = Game.new.outcome #This seems to be the problem when I run sinatra.
+        erb :result
     end
+
+    # post '/game' do
+    #     @weapon = params[:weapon]
+    #     p params
+    #     erb :game
+    # end
+
+    # post '/result' do
+    #     @weapon = params[:weapon]
+    #     p params
+    #     erb :result
+    # end
 
     run! if app_file == $0
 end
--- a/app/rps.rb
+++ b/app/rps.rb
@@ -1,8 +1,8 @@
 require "sinatra/base"
 require "sinatra/reloader"
 require "./lib/player.rb"
-require "./lib/weapon.rb"
 require "./lib/computer.rb"
+require "./lib/game.rb"
 
 class RPS < Sinatra::Base
     enable :sessions
@@ -16,28 +16,15 @@ class RPS < Sinatra::Base
 
     post '/weapons' do
         $player = params[:player_1]
-        $computer = Computer.new.compmove
+        $cpu = Computer.new.compmove
         erb :weapons
     end
     
     post '/result' do
         $weapon = params[:p1move]
-        $computer = Computer.new.compmove
-       # $outcome = Game.new.outcome #This seems to be the problem when I run sinatra.
+        $outcome = Game.new.result
         erb :result
     end
-
-    # post '/game' do
-    #     @weapon = params[:weapon]
-    #     p params
-    #     erb :game
-    # end
-
-    # post '/result' do
-    #     @weapon = params[:weapon]
-    #     p params
-    #     erb :result
-    # end
 
     run! if app_file == $0
 end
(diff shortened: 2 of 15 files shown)


app/rps.rb
```diff
@@ -0,0 +1,43 @@
...
+    post '/weapons' do
+        $player = params[:player_1]
+        $computer = Computer.new.compmove
+        erb :weapons
+    end
+    
+    post '/result' do
+        $weapon = params[:p1move]
+        $computer = Computer.new.compmove
+       # $outcome = Game.new.outcome #This seems to be the problem when I run sinatra.
```
sherwingp: You forgot to require game.rb

lib/game.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'computer'
+require './app/rps'
+
+class Game
+    attr_reader :outcome
+
+def compmove
```
sherwingp: I can see what your intention is, but every time you call the method compmove, it is selecting a random choice again because of the .sample and assigning it to @compmove. You should move @compmove to your initialise method and add an attribute reader for it

app/rps.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    post '/weapons' do
+        $player = params[:player_1]
+        $computer = Computer.new.compmove
+        erb :weapons
+    end
+    
+    post '/result' do
+        $weapon = params[:p1move]
+        $computer = Computer.new.compmove
```
sherwingp: This is probably a source of a bug. You are correctly using p1move, that is the weapon you chose. But you are reassigning computer to a **new** instance of Computer. You don't have to do this because you already assigned an instance of it to the global variable $computer in the /weapons route

lib/computer.rb
```diff
@@ -0,0 +1,7 @@
+class Computer
+   attr_reader :compmove
+    def compmove
```
sherwingp, line 3: I noticed this method is also present in game.rb. One of these should be removed

app/rps.rb
```diff
@@ -0,0 +1,43 @@
...
+       # $outcome = Game.new.outcome #This seems to be the problem when I run sinatra.
+        erb :result
+    end
+
+    # post '/game' do
```
sherwingp: Yes, I think these routes are redundant now.

app/views/game.erb
```diff
@@ -0,0 +1,8 @@
+<h1> You chose <%=@weapon%> </h1>
```
sherwingp, line 1: This view is unused

app/views/result.erb
```diff
@@ -0,0 +1,6 @@
+<h1> You chose <%= $weapon %>.<h1>
+<h1> Your opponent chose <%= $computer %>.</h1>
+
+<%# # I want to write: 'You lose/win/draw' by using the global variable of $game. %>
```
sherwingp: Remember to add this implementation once you fix the underlying bug in your model (Game.rb)

app/rps.rb
```diff
@@ -0,0 +1,43 @@
...
+      register Sinatra::Reloader
+    end
+
+    get '/' do
+        erb :index
+    end
+
+    post '/weapons' do
+        $player = params[:player_1]
+        $computer = Computer.new.compmove
```
sherwingp: We assigned the computer move to $computer here. But once you call the outcome method in the /results route, is it using the computer move assigned here?

README.md
```diff
@@ -13,7 +15,7 @@ Task
 
 Knowing how to build web applications is getting us almost there as web developers!
 
-The Makers Academy Marketing Array ( **MAMA** ) have asked us to provide a game for them. Their daily grind is pretty tough and they need time to steam a little.
+The Makers Academy Marketing Array ( **MAMA** ) have asked us to provide a game for them. Their daily grind is pretty tough and they need time to let off steam a little.
```
alicelieutier, line 18: If you make a separate PR with just this change, I'll merge it to the course repo ;)

app/rps.rb
```diff
@@ -0,0 +1,30 @@
+require "sinatra/base"
+require "sinatra/reloader"
+require "./lib/player.rb"
+require "./lib/computer.rb"
+require "./lib/game.rb"
+
+class RPS < Sinatra::Base
+    enable :sessions
```
alicelieutier, line 8: You enable sessions but do not use them afterwards, instead preferring a global variable.
I actually think the code would be nicer using sessions.

app/views/game.erb
alicelieutier: @christopherxwebb you marked this as resolved, but the view is still unused. Maybe delete this file?

app/rps.rb
```diff
@@ -0,0 +1,30 @@
...
+        $player = params[:player_1]
+        $cpu = Computer.new.compmove
+        erb :weapons
+    end
+    
+    post '/result' do
+        $weapon = params[:p1move]
+        $outcome = Game.new.result
+        erb :result
+    end
```
alicelieutier, line 27: You may want to follow the post -> redirect -> get pattern here instead or returning views from the post routes. Check https://github.com/makersacademy/course/blob/main/intro_to_the_web/post_redirect_get_pattern.md

lib/game.rb
```diff
@@ -0,0 +1,17 @@
...
+
+def result
+  if $weapon == "rock" && $cpu == "scissors" || $weapon == 'paper' && $cpu == "rock" || $weapon == 'scissors' && $cpu == "paper"
+    :win
+  elsif $weapon == $cpu
+    :draw
+  else
+    :lose
+    end
+  end
```
alicelieutier, line 16: You are accessing global variables from this class. This is not ideal. Basically, your model only works because it relies on the controller setting all these global variables in the right way, and you classes don't speak with eachother.
When you develop a web app, the model part of it should still very much work on its own, like an irb app.
Then the controller and views are just there to make the interface better for the user.

One way you can do this is developing the model as an irb app, using feature tests (in irb or rspec) and unit tests.

In anycase, you should be passing the player (or player move) and the computer (or computer move) to the game class through arguments, as opposed to the game class going and reading global variables outside itself.

lib/player.rb
```diff
@@ -0,0 +1,3 @@
+# class Player
```
alicelieutier, line 1: Delete unused files.

spec/features/enter_name_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'Enter names' do
+    scenario 'submitting names' do
+        visit ('/')
+        fill_in :player_1, with: 'Player'
+        click_button 'Submit'
+        expect(page).to have_content 'Ok Player choose your weapon:'
+    end
+end
```
alicelieutier, line 8: Good test! It would be nice to see a second one, to check that the screen does adapt to the name being entered, and doesn't always reply `'Ok Player choose your weapon:'` even if I enter 'Alice' for example.
You could also test an error scenario, for example there could be an error if the user leaves the player field empty.

spec/features/pick_weapon_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'Picking Weapon' do
+    scenario 'to choose weapon against CPU' do
+        visit ('/')
+        fill_in :player_1, with: 'Player'
+        click_button 'Submit'
+        click_button 'rock'
+        expect(page).to have_content 'You chose rock'
+    end
+end
```
alicelieutier, line 9: Same comment as before. This is a good test, but you should add more tests for each feature.

alicelieutier: Also, by stubbing randomness (check `srand`) you could test the full game, including the result page.

alicelieutier (commented):
It would be nice to see some unit tests for the game class, as it does some heavy logic with calculating the results.
You class organisation in the model can be improved to function more as an independant part, and be less coupled to your controller. See my comments below.

README.md
christopherxwebb: haha thanks Alice. I was being pedantic here!

app/rps.rb
christopherxwebb: Ok great. This has helped because I was unsure about sessions, what they did. I understand they temporarily store form data but need to do a bit more research on how they actually work.

app/rps.rb
christopherxwebb: Great, thanks Alice, I'll take a look now!

lib/player.rb
christopherxwebb: Really bad at (not) doing this. I always like to keep things open but as you rightly suggest makes things way more complicated.

spec/features/enter_name_spec.rb
christopherxwebb: yes great idea!

spec/features/pick_weapon_spec.rb
christopherxwebb: Ah, yes, good point! I'll look into stubbing later on today. Thank you for all these Alice, really helpful!

PR #2019, opened 2021-12-06
Solution for weekend challenge

kim-morgan:
Basic challenges should all be complete


spec/features/game_spec.rb
```diff
@@ -0,0 +1,24 @@
+feature "should have options to click" do
+
+  scenario "can select rock" do
+    visit('/')
+    fill_in('name', with: 'Kim')
+    click_button('Submit')
+    expect(page.has_button?('Rock')).to be_truthy
```
Alaka-K: awesome! didn't know we could use this type of syntax

Alaka-K: awesome! didn't know we could us this syntax

public/styles.css
```diff
@@ -0,0 +1,91 @@
+html {
```
Alaka-K, line 1: A Lot of detail went into your styling of the page. great Job!

spec/features/game_page_spec.rb
```diff
@@ -0,0 +1,18 @@
+feature "should have options to click" do
+
+  scenario "can select rock" do
+    sign_in_and_play
+    expect(page.has_button?('Rock')).to be_truthy
```
Alaka-K, line 5: awesome! Didn't know we could us this type of syntax

views/layout.erb
```diff
@@ -0,0 +1,14 @@
...
+<html>
+  <head>
+    <link href="<%= url('/styles.css') %>" rel="stylesheet" type="text/css" />
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link href="https://fonts.googleapis.com/css2?family=Sora:wght@300&display=swap" rel="stylesheet">  
+  </head>
+  <body>
+  <div class="content">
+        <%= yield %>
```
Alaka-K, line 11: what does this do?

lib/game.rb
```diff
@@ -0,0 +1,26 @@
+require_relative "./player"
```
Alaka-K, line 1: Good use of class interaction.

lib/game.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class Game
+
+  attr_reader :player1, :player2
+
+  BEATS = { "Rock" => "Scissors",
```
Alaka-K, line 7: nice! never would have thought of using a hash

views/layout.erb
kim-morgan: Oh, so basically Jess told me this trick when we were pairing on Friday! If you add a file called layout.erb, then Sinatra will automatically check for it and if this file exists it will apply that layout to every single page. It means that you don't have to link the css stylesheet for every page. 

The content of each page will be displayed after  <%= yield %>. Everything before that will be applied automatically.

Alaka-K (commented):
I don't have any constructive feedback .Really well done. Wish my could be up to this standard.

views/layout.erb
Alaka-K: ah awesome. should note that for next time

PR #2022, opened 2021-12-06
RPS Weekend challenge

zidehen:
Most of the user stories completed, however I got stuck on the last one so was unable to add ability to declare winner


app.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  post '/name' do
+    # session[:marketeer_name] = params[:marketeer_name]
+    $marketeer = Player.new(params[:marketeer_name])
+    redirect '/play'
+  end
+
+  get '/play' do
+    # @marketeer_name = session[:marketeer_name]
+    @marketeer_name = $marketeer
```
leoht, line 26: Looks like it should maybe either be
```ruby
@marketeer_name = $marketeer.name
```
or
```ruby
@marketeer = $marketeer
```
? By naming it `@marketeer` it makes it clear the full object is stored in this variable

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def register_and_play
```
leoht, line 1: Nice use of helpers here to make the code DRY :)

views/game.erb
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+<body>
+<h2>You chose: <%= @play_option %>
+<br>
+<br>
+Computer chose: <%= @computer %></h2>
+</div>
```
leoht: Looks like you could use `@game.result(@play_option, @computer)` here and it would work? Reading the code, seems that we would get the correct result

app.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  post '/game' do
+    session[:play_option] = params[:play_option]
+    redirect '/game'
+  end
+
+  get '/game' do
+    @play_option = session[:play_option]
+    $computer = Computer.new
+    @computer = $computer.move
```
leoht, line 38: Here again, probably better to name it `@computer_move` to make the code more readable

lib/game.rb
```diff
@@ -0,0 +1,11 @@
+class Game
+
+  def result(player, computer)
+    return 'You Win!' if player == 'Rock' && computer == 'Scissors'
```
leoht, line 4: To have a better separation of concerns, it's better to leave the view (the .erb file) to display the human-readable messages such as "You win!" etc, rather than it being returned by this method (`Game` being part of the Model layer). An easy way to improve this is simply to have the winner of the game returned, or maybe a symbol such as `:won`, `:lost` etc. And then pass this value from the controller to the view, and let the view decide what message to display :)

spec/features/choosing_option_spec.rb
```diff
@@ -0,0 +1,16 @@
+feature "Choosing an option" do
+
+  scenario "choose a play option" do
+    register_and_play
+    click_button 'Rock'
+    expect(page).to have_content 'You chose: Rock'
+	 end
+
+  scenario "computer can choose a random option" do
+    allow_any_instance_of(Computer).to receive(:move).and_return(:Rock)
```
leoht, line 10: Good use of stubbing here to get a predictable result :) 
I'd probably add an extra test doing the whole flow, to verify that the result page is showing the correct game result (e.g the computer choses rock, the player choses scissors, the result page should show 'You lost')

leoht (commented):
Nice work! I couldn't reproduce the bug you mentioned in Slack, and all the RSpec seem to be green — maybe you've fixed/improved it since you've asked for review? In any case, the controller seems to call the model code correctly, and pass it using instance variables to the view. I could suggest to:
 * improve name of variables and methods, to be more explicit about what a variable stores, and make the code more readable
 * make sure the code is correctly indented, as it makes it easier to read and follow, so it's easier for you to understand where bugs might come from, and make sense of it

Let me know in case there's anything you want to discuss further on it, or if you're still having some bugs/issues with it you'd like to discuss

PR #2023, opened 2021-12-06
rps-challenge

Mayo-Theodore:


app/app.rb
```diff
@@ -0,0 +1,36 @@
...
+get '/play' do
+  erb :play
+end
+
+post '/play' do
+  erb :play
+end
+
+post '/result' do
+  @computer_play = ["Rock", "Paper", "Scissors"].sample
```
AlfonsoGhislieri, line 31: Try to keep any logic or calculations in the model and not the controller

app/app.rb
```diff
@@ -0,0 +1,36 @@
...
+  erb :result
+end
+
+  run! if app_file == $0
+end
```
AlfonsoGhislieri, line 36: Do you need this many get/post routes?

app/views/result.erb
```diff
@@ -0,0 +1,11 @@
...
+<h2>You win!</h2>
+
+
+<form action="/play" method="post">
+  <input type="submit" value="Play Again">
+</form>
+
+<form action="/" method="post">
+ <input type="submit" value="End Game">
+</form>
```
AlfonsoGhislieri, line 11: Good idea to give the user the option to end the game! You could count how many wins the user and computer has and display those after!

app/app.rb
Mayo-Theodore: Was experimenting with the Play Again/End Game features. Definitely needs refactoring

PR #2027, opened 2022-01-09
Josh Lim - RPS Challenge

sumo-mailman:


app.rb
```diff
@@ -0,0 +1,64 @@
...
+    redirect '/play'
+  end
+  
+  get '/play' do
+    $game = Game.new($player_1.name, $player_2.name)
+    @name1 = $game.player1
+    @name2 = $game.player2
+    erb(:play)
+  end 
+
```
mat-diak, line 34: In my understanding, the use of global variable on $game, and the fact that Game class stores players' names, would allow you not to have to create players into global variables. Instead, you could retrieve the players from game each time.

app.rb
```diff
@@ -0,0 +1,64 @@
...
+  get '/outcome' do 
+    @move1 = $game.move1
+    @move2 = $game.move2
+    @name1 = $game.player1
+    @name2 = $game.player2
+    @outcome = $game.result(@move1, @move2)
+    erb(:outcome)
+  end
```
mat-diak, line 60: In order to keep your controller skinny, you could just use @game = $game and access the player names and moves already in your views.
` @game.player1.name ` etc.

lib/game.rb
```diff
@@ -0,0 +1,19 @@
...
+    @player2 = name2
+  end
+
+  def result(player, computer)
+    return "#{player1} is the winner!" if player == 'Rock' && computer == 'Scissor'
+    return "#{player1} is the winner!" if player == 'Scissor' && computer == 'Paper'
+    return "#{player1} is the winner!" if player == 'Paper' && computer == 'Rock'
+    return "It's a DRAW!" if player == computer
+    "Sorry, this time #{player2} is the winner."
+  end
```
mat-diak, line 17: The `result` method not only determines a winner but also creates a string. I think it would be better to have 2 methods doing this.

mat-diak, line 17: The rules could be refactored to avoid using so many if statements. The code review rubric suggest using a hash.
` RULES = { rock: :scissors,
                     paper: :rock,
                     scissors: :paper } `
In my understanding, this would allow you have less if statements but also you could add optionality without having to do much work.

spec/game_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  describe '#result' do
+    let(:player1) { double :player }
+    let(:player2) { double :player }
+    subject(:game) { described_class.new(:player1, :player2) }
+
+    it 'recognises a win' do 
+      game.move1 = 'Rock'
+      game.move2 = 'Scissor'
+      expect(game.result(game.move1, game.move2)).to eq("#{game.player1} is the winner!")
+
```
mat-diak, line 29: It seems that you are using an `attr_accessor` only to stub player moves. I am not sure if this is good or bad. You could, however, have players have `@choice` variables and then add `choice` method to your player doubles.

app.rb
```diff
@@ -0,0 +1,64 @@
...
+    erb(:play_two)
+  end 
+
+  post '/moves2' do 
+    params[:second_move]
+    $game.move2 = params[:second_move]
+    redirect '/outcome'
+  end 
+
+  get '/outcome' do
```
mat-diak, line 53: The controller correctly uses the post/redirect/get pattern

mat-diak (commented):
Don't take anything I say for granted.
The code seems easy to read and organised. The structure of the files follows the correct template. I run the code myself and everything worked smoothly with the `new players` and `play again` options. 

I also had an issues with the bundler. I had to manually install `webrick` to run `rackup`.

You seem to be missing feature tests for win, loss, draw scenario.

app.rb
eoinmakers: Agreed! Some devs also make use of sessions to store things like the player name.

app.rb
```diff
@@ -0,0 +1,64 @@
...
+    redirect '/play_two'
+  end 
+
+  get '/play_two' do 
+    @name1 = $game.player1
+    @name2 = $game.player2
+    erb(:play_two)
+  end 
+
+  post '/moves2' do
```
eoinmakers, line 47: Some duplication of efforts in this route - is there a way you could use an if/else in the first /moves route to determine if another player should make a move or if the outcome should be calculated?

app.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  post '/moves' do
+    params[:first_move]
+    $game.move1 = params[:first_move]
+    redirect '/play_two'
+  end 
+
+  get '/play_two' do 
+    @name1 = $game.player1
+    @name2 = $game.player2
```
eoinmakers, line 43: Doesn't meet the specification outlined in the user stories, as player 2 should be the computer that makes a random choice.

lib/player.rb
```diff
@@ -0,0 +1,9 @@
+class Player
```
eoinmakers, line 1: Maybe this class could also store the players choice? In OOP, we like to parcel together state & behaviour: in other words, data and the actions that act on that data. We could store the player's name, and their chosen move - fed into the instance on creation via params in the app.rb.

eoinmakers (commented):
Nice attempt! I left just a few comments - I think one of the main takeaways would be in re-reading the specification / user stories, as this submission doesn't meet the ones outlined for RPS. This is a super important thing to take care of in tech tests - that you can work exactly to a specification, with the simplest implementation possible. 

You've demonstrated the implementation of web app logic from View -> Controller -> Model, nice work. 

Mateusz' feedback is well thought out, and worth using to reflect on your code and seeking opportunities to refactor.

PR #2029, opened 2022-01-09
Raven Pull Request for RPS Challenge

ravensears:
Used this walkthrough to work through a little more than half of the challenge:

https://www.youtube.com/watch?v=GoHKmartBYo


spec/features/play_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'spec_helper'
+
+feature 'playing a game' do
+    before do
```
naima90, line 4: great feature test & easy to understand but to keep code dry, maybe you could've used a helper to avoid constant repetition instead of "before do"

naima90 (commented):
overall, really good code, easy to understand the problem it's trying to solve, keep up 👏

PR #2030, opened 2022-01-10
Victoria - RPS challenge

vblack0:


OrganicAC (commented):
I really like your readme! It explains your process and how you want to improve going forward! Only thing I would add is instructions for the user on how to actually play a game of RPS

app.rb
```diff
@@ -9,33 +9,46 @@ class RPSGame < Sinatra::Base
...
+  end
+
   get '/' do
     erb :index
   end
 
   post '/names' do
+    session[:tally] = { player: 0, computer: 0, draws: 0 }
     session[:player] = params[:player_1_name].capitalize
-    # I promise to never use global variables in tech tests
```
OrganicAC: Good promise!

lib/game.rb
```diff
@@ -1,25 +1,18 @@
 class Game
-  attr_reader :player, :result, :tally
+  attr_reader :player, :result
 
   MOVES = ['Rock', 'Paper', 'Scissors'].freeze
```
OrganicAC, line 4: I like the use of a constant here with the .freeze

OrganicAC:
Would it have been possible to merge the win/lose/draw erb files? As it stands it feels like it might be a lot to track if you make a change, then alter it across all three pages?

OrganicAC:
I like that you challenged yourself to add css to make it look like more than just a plain webpage.

PR #2031, opened 2022-01-10
Kiran's RPS Challenge

KiranKukar:


jmodiale (commented):
Overall, your code looks clean, it's readable and does what it's supposed to do. Great job

README.md
```diff
@@ -1,86 +1,3 @@
...
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
-  SimpleCov::Formatter::Console,
-  # Want a nice code coverage website? Uncomment this next line!
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
-
-You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
+I started by setting up the test environment, rackup and sinatra; so that I can use test driven development.
```
jmodiale: It would be great to add some how-to-play instructions in the readme.

app.rb
```diff
@@ -0,0 +1,54 @@
...
+    @game = Game.instance
+    @game.player_selection(selection)
+    redirect '/result'
+  end
+
+  post '/scissors' do
+    selection = 'Scissors'
+    @game = Game.instance
+    @game.player_selection(selection)
+    redirect '/result'
```
eoinmakers, line 50: There is some duplication of code between these three routes, they're carrying out the same action and the only variation is the string you pass in as an argument. 

Can you think of a way in which you could refactor your code in order to only need one route, with an alternative method of passing in the player choice? Perhaps something like params or sessions could be of use here.

lib/game.rb
```diff
@@ -0,0 +1,57 @@
...
+  def computer
+    choices = ['Rock', 'Paper', 'Scissors']
+    @computer_selection = choices[rand(3)]
+  end
+
+  def player_selection(selection)
+    @selection = selection
+  end
+
+  def result
```
eoinmakers: This method is a little long - it's complex enough logic, but do you think you could make it more readable with the use of the || OR operator?

eoinmakers: Something I would do to try to map out this logic is write it out in pseudocode: https://www.unf.edu/~broggio/cop2221/2221pseu.htm

I find it easier to work on a piece of complex logic by having it in the more readable format of pseudocode - from there, you can identify repetition, move lines around more easily, consider different possibilities.

eoinmakers (commented):
This is a good implementation of Rock Paper Scissors, nice work. There are two areas where refactoring would mean that the code is more easily readable, and there would be less duplication of code. 

There are good examples of feature and unit tests here too. If you wanted to fully unit test this, you would actually test for every possible combination of player & computer choices / every outcome.

PR #2033, opened 2022-01-10
Rock, Paper, Scissors Challenge

hannahdesmond:


README.md
```diff
@@ -51,12 +41,6 @@ Change the game so that two marketeers can play against each other ( _yes there
 
 Use the _special_ rules ( _you can find them here http://en.wikipedia.org/wiki/Rock-paper-scissors-lizard-Spock_ )
 
-## Basic Rules
-
-- Rock beats Scissors
-- Scissors beats Paper
-- Paper beats Rock
-
 In code review we'll be hoping to see:
```
smm3000, line 47: Would be great to include some how-to-run instructions in the readme.

lib/game.rb
```diff
@@ -0,0 +1,32 @@
+class Game
+
+  def initialize(player:, computer:)
+    @player_move = player
+    @computer_move = computer
+  end
+
+  def winner
```
smm3000, line 8: I like the use of 'case' within your game logic. Another possible method that was mentioned on the walkthrough was the use of a 'GAME_RULES' constant that uses a hash to summarise the different scenarios for each move.

public/styles.css
```diff
@@ -0,0 +1,11 @@
...
+  background-color: #AEFEFF;
+  font-family: "Papyrus", fantasy;
+  font-color: 072227;
+}
+h1 {
+  font-size: 50px;
+}
+form {
+  font-size: 30px;
+}
```
smm3000, line 11: I like the styling choices!

app.rb
```diff
@@ -0,0 +1,40 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require_relative './lib/computer'
+require_relative './lib/game'
```
alicelieutier, line 4: Could these be require rather than require relative? Or were you having troubles with that?

lib/game.rb
```diff
@@ -0,0 +1,32 @@
+class Game
```
alicelieutier, line 1: I like the style of methods in this class. Very "functional programming", by which I mean minimal use of instance variable (just what's needed to store information, and methods that returns values rather than change the state.
Also very readable - really good!

spec/computer_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'computer'
+
+describe Computer do
+  describe "#move" do
+    let(:computer) { described_class.new }
+    it 'receives #move' do
+      allow(computer).to receive(:move).and_return('paper')
```
alicelieutier, line 7: Stubbing the class under test is bad practice. you end up just testing the stub here, and not the code. You could remove the move method in the Computer class and this test would still pass.

Instead you should stub things that are outside your class. In this case, what is outside your class is the random generator, that is called by sample under the hood.

One way to stub that is to use `srand` - see [the docs here](https://ruby-doc.org/core-2.4.0/Random.html#method-c-srand). There are other ways where you can stub the sample methods on arrays, but I think that would still not test every case.

alicelieutier: If you end up using `srand`, for good measure I would test that each of the possible computer moves can be returned from the method.

app.rb
```diff
@@ -0,0 +1,40 @@
...
+
+class RockPaperScissors < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  enable :sessions
```
alicelieutier, line 11: Great use of sessions and of the post redirect get pattern in this file.

spec/features/rps_spec.rb
```diff
@@ -0,0 +1,21 @@
+feature 'name submission' do
+  scenario 'player 1 can register name' do
+    sign_in_and_play
```
alicelieutier, line 3: In this specific case, I think it's good to have the sign in explicit describe within the test to make it easy to understand what happens if ever the test fails by only looking at this file. 
You can use the helper method for other tests where the name of the player doesn't get tested.

DRY is less important in tests than simplicity of understanding what's going on.

spec/features/rps_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    expect(page).to have_content("Hannah – it's time to play Rock, Paper, Scissors!")
+  end
```
alicelieutier, line 5: It's always nice to have several tests for each behaviour, otherwise, who's to say your code actually uses the user provided name and doesn't just hard-code "Hannah" as the player name?

As a bonus, you could imagine that if the player enters and empty string, the submit doesn't work maybe?

spec/features/rps_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    expect(page).to have_content("Choose rock, paper or scissors:")
+  end
+  scenario 'player loses' do
+    play_and_lose
+    expect(page).to have_content("Computer chose paper – you lose!")
+  end
+  scenario 'player wins' do
+    play_and_win
+    expect(page).to have_content("Computer chose rock – you win!")
+  end
```
alicelieutier, line 20: Same comments as above concerning use of helper method (make the tests more explicit here) and stubbing - you can stub randomness rather than the computer... although here it's less of a problem because you are testing the system and not just the computer.

Also - it would be nice to have a test for draws!

spec/game_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'game'
+
+describe Game do
+  let(:game) { described_class.new(computer: 'rock', player: 'scissors') } 
+  it 'calculates the winner' do
+    expect(game.winner).to eq('Computer chose rock – you lose!')
+  end
```
alicelieutier, line 7: Maybe test different outcomes?

views/index.erb
```diff
@@ -0,0 +1,16 @@
+<head>
+  <title>Rock, Paper, Scissors</title>
+  <link href="<%= url('/styles.css') %>" rel="stylesheet" type="text/css" />
+</head>
```
alicelieutier, line 4: Take a look at partials so you don't have to rewrite the head for every page: http://sinatrarb.com/faq.html#partials

alicelieutier (commented):
Really good challenge overall.
Your code is clean and easy to read, demonstrating good SRP and MVC understanding.
You can improve your tests, especially the computer one, and add some more tests cases.
See my comments below.

PR #2034, opened 2022-01-10
RPS Challenge Attempt

smm3000:


app.rb
```diff
@@ -0,0 +1,31 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+
```
hannahdesmond, line 5: An extra line here should help you run your app.

app.rb
```diff
@@ -0,0 +1,31 @@
...
+        erb :index
+    end 
+
+    post '/names' do 
+        session[:player_name] = params[:player_1_name]
+        redirect '/play'
+    end 
+
+    get '/play' do
+        @turn = Turn.new(session)
```
hannahdesmond, line 24: When I printed your `session` variable here, I got this: <Rack::Session::Abstract::PersistedSecure::SecureSessionHash:0x00007faf7e0fb858>
It looks like you're passing in the whole hash, but I am not sure it has the information that you need, so maybe you could assign session[:player_name] to a variable and pass that into your new instance of Turn?

lib/Turn.rb
```diff
@@ -0,0 +1,35 @@
+class Turn
+    GAME_RULES = {
```
hannahdesmond, line 2: I like how you've calculated who wins – it's really clever.

spec/feature/play_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        click_button 'Paper'
+        expect(page).to have_content 'You lose'
+    end 
+
+    scenario 'I draw' do
+        click_button 'Scissors'
+        expect(page).to have_content 'Draw'
+    end 
+
+    def possible_messages
```
hannahdesmond, line 57: I really like this helper method! Your feature tests seem really comprehensive.

hannahdesmond (commented):
You’ve done some really clever things in your code. I think it highlights your Ruby skills. There are a couple of `require` lines missing seemingly that is stopping your code from running but you should be able to debug those easily! Your classes have a single responsibility, which is great. I have a couple of lines that I wasn't sure what they were doing (through my lack of knowledge) that I will ask you about later! Overall, your code looks great!

PR #2036, opened 2022-01-10
RPS Challenge - Fran

fpmrqs:


cmkap:
I love how you use SRP--I hate extracting classes. Your classes are small and readable. Loved your match logic and the use of private methods. Again, your code reads well and looks tidy to me. I'm having trouble running it on my side with bundle install and rackup, but it covers all the things we were taught last week. One MAJOR :smile: thing I would suggest is that you go away from global variables. They taught us it, but as they said it is coding sin. Session can be used instead.  Your specs are clean. The use of doubles  :+1:.  Also the addition of the web helper to stop redundancy is a good touch.

nice work

app.rb
```diff
@@ -15,15 +16,27 @@ class RPS < Sinatra::Base
   end
 
   get '/game' do
-    @player_1_name = $game.player_1.name
+    @player_one_name = $game.player_one.name
     erb :game
   end
 
+  get '/results' do
+    $match.match_decider
```
cmkap: Use session instead of global variable

views/game.erb
```diff
@@ -3,6 +3,15 @@
 
 <body>
 <div class='tittle'>
```
cmkap, line 5: good use of SRP

app.rb
```diff
@@ -15,15 +16,27 @@ class RPS < Sinatra::Base
   end
 
   get '/game' do
-    @player_1_name = $game.player_1.name
```
cmkap: Avoid Global variable

lib/game.rb
```diff
@@ -1,8 +1,8 @@
 require_relative 'player'
```
cmkap, line 1: Good SRP

lib/match.rb
```diff
@@ -1,9 +1,32 @@
 class Match
```
cmkap, line 1: The class reads very well

lib/match.rb
```diff
@@ -1,9 +1,32 @@
...
+    if rock_vs_scissors?
+      @message = 'Rock beats Scissors. ' + winner
+    end
+  end
+
+  def show_message
+    @message
+  end
+
+  private
```
cmkap, line 18: Nice use of private methods

spec/features/match_results_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'Match can show results' do
+  let(:match) { double(:match, player_one_move => 'rock', player_two_move => 'scissors') }
```
cmkap: Good use of doubles

spec/features/web_helper.rb
```diff
@@ -1,5 +1,5 @@
 def fill_name_and_submit
```
cmkap, line 1: Great addition to limit redundency in the specs

cmkap (commented):
I love how you use SRP--I hate extracting classes. Your classes are small and readable. Loved your match logic and the use of private methods. Again, your code reads well and looks tidy to me. I'm having trouble running it on my side with bundle install and rackup, but it covers all the things we were taught last week. One MAJOR :smile: thing I would suggest is that you go away from global variables. They taught us it, but as they said it is coding sin. Session can be used instead.  Your specs are clean. The use of doubles  :+1:.  Also the addition of the web helper to stop redundancy is a good touch to keep everything DRY

app.rb
```diff
@@ -0,0 +1,41 @@
...
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  get '/game' do
+    @player_one_name = $game.player_one.name
```
alicelieutier, line 19: It would be nice to use sessions instead of global variables.

alicelieutier: Here's the problem with globals:
* open a browser tab to you home page (localhost:9292)
* enter a name, eg "Sarah" and click submit.
* the page says "sarah, pick your move"
* Now open another browser tab (or could be a different browser) to the home page again
* enter another name, eg "bob" and click submit
* the page says "bob, pick your move"
* go back to the first browser tab you opened, and refresh the page.
* the page now says  "bob, pick your move" instead of "sarah, pick your move" !!!

This is because a global is... global. The server only stores one name variable. Using sessions prevents this bug.

spec/features/match_results_spec.rb
```diff
@@ -0,0 +1,18 @@
+feature 'Match can show results' do
+  before { fill_name_and_submit }
+
+  scenario 'player picks rock and sees result' do
+    click_on 'Rock'
+    expect(page).to have_content @message
```
alicelieutier, line 6: What is `@message` here?
Tests are always best when they are easy to understand right away, and tests don't need to change as much as code. So DRY doesn't apply in tests as much. 

If you have a code base with 2000 tests and 1 is failing, you want to understand the full problem by only looking at that one test. Here, looking at this test, I don't know what `@message` is, so I don't know what the test is doing.

spec/features/player_names_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature 'Player names' do
+  scenario 'players can input their names' do
+    fill_name_and_submit
+    expect(page).to have_content('Pabllo, pick your move')
+  end
```
alicelieutier, line 5: For this specific test, I wouldn't use the `fill_name_and_submit`, because then reading just this file I don't know what your test does.

Also, you can make these tests more robust by testing with different names (after all, this test would still be green if you hard coded the name in the app to be "Pabllo"). 
You could also have a case where the user didn't enter any name. What should happen then? Should there be an error?

spec/match_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'match'
+
+describe Match do
+  context 'Decides winner' do
+
+    it 'rock wins' do
+      match = Match.new('Rock', 'Scissors')
+      match.decider
+      expect(match.show_message).to eq 'Rock beats Scissors. You win!'
```
alicelieutier, line 9: I think it would be a better design to have the decider and message in one method. Maybe you call `match.winner`, and it does the deciding + returns the message.
Otherwise, you methods depend on the order in which they are called, and "show_message" is only valid if called just after decider, which makes the code more breakable / harder to use.

lib/game.rb
```diff
@@ -0,0 +1,8 @@
+require_relative 'player'
+
+class Game
+  attr_reader :player_one
+  def initialize(player_one = Player.new)
+    @player_one = player_one
+  end
+end
```
alicelieutier, line 8: This class doesn't seem necessary at the moment as it just stores the player... which is already stored in player. You could remove it and gain in simplicity.

alicelieutier (commented):
Good work overall - plenty of tests, and good SRP, especially MVC separation.
You could improve a few things in the design, see my comments below.

PR #2045, opened 2022-02-11
Rock, Paper, Scissor - Bromley

nyahehnagi:
Initial commit


app.rb
```diff
@@ -0,0 +1,89 @@
+require 'sinatra'
+require "sinatra/reloader" if development?
+
+require File.join(File.dirname(__FILE__), 'lib', 'player.rb')
```
eoinmakers, line 4: A neater alternative is require 'file_name.rb', for files that are in the expected place (lib folder0.

app.rb
```diff
@@ -0,0 +1,89 @@
...
+    redirect('/enter_names')
+  end
+
+  get '/enter_names' do
+    @game_type = session[:game_type]
+    erb(:names)
+  end
+
+  post '/register_name' do
+    redirect('/enter_names') unless name_params?
```
eoinmakers, line 50: Could you push the logic of name_params into the model?

For example, if you were to save the player's name on the Player object, you could check if @player.name is empty. If it is, redirect to the /enter_names era, if not, redirect to player_one_select. 

Your implementation of this is not untidy as you've deferred it to helper methods.

app.rb
```diff
@@ -0,0 +1,89 @@
...
+    erb(:player_two_select)
+  end
+
+  post '/player_one_plays' do
+    redirect('/player_one_select') if params.empty?
+    session[:player_one_choice] = params[:selection].to_sym
+    session[:game_type] == :computer ? redirect('/result') : redirect('/player_two_select') 
+  end
+
+  post '/player_two_plays' do
```
eoinmakers, line 72: There are separate route definitions for player one and player two selecting and playing the game: can you think of any way you could reduce this duplication of code so that the routes would work more generically, no matter who the player is? What else could you change to enable this?

lib/game.rb
```diff
@@ -0,0 +1,26 @@
...
+    @player_two = player_two
+  end
+
+  def play_game
+    rock_paper_scissor(@player_one.weapon, @player_two.weapon)
+  end
+
+  private
+
+  def rock_paper_scissor(wep_one, wep_two)
```
eoinmakers, line 17: Full names for variables is considered better practice - eg. weapon_one, weapon_two.

spec/botplayer_spec.rb
```diff
@@ -0,0 +1,16 @@
+require_relative '../lib/botplayer'
+
+describe BotPlayer do
+
+  subject(:bot_player) { described_class.new }
+
+  it "has a weapon" do
```
eoinmakers, line 7: This test is just testing that your stub is working, so I wouldn't include it. You could test that the result returned from bot_player.weapon is within a range, but you don't need to test that one specifically comes back (testing the logic of the game for each combination of weapons covers this).

spec/game_spec.rb
```diff
@@ -0,0 +1,85 @@
+require_relative '../lib/game'
+
+describe Game do
```
eoinmakers, line 3: Very thorough unit testing here.

views/player_two_select.erb
```diff
@@ -0,0 +1,9 @@
+
+<%=@player%>, are you ready to play?
+<br><br>
+
+<form action="/player_two_plays" method="post">
+<%=erb(:weapon_input)%>
+</form>
+
+
```
eoinmakers, line 9: Small note: neater to remove the extra whitespace at the end & beginning of this erb file.

eoinmakers (commented):
Overall a great RPS implementation. I think this week's work on class methods may provide opportunities to revisit the structure of the interactions between your models and routes, but a refactor isn't necessary here - just reflection. 

Nice work.

app.rb
nyahehnagi: I did think about doing that.. I was going to put the logic if/else style in the view and have it work it out, I decided against it.. almost expecting the feedback :-) Is it Ok then to have some logic in the views like that?

lib/game.rb
nyahehnagi: Fair point :-) noted!

spec/botplayer_spec.rb
nyahehnagi: Yes, I realised this at the time. Good spot. I'll have a look at the range suggestion, it's a good idea

views/player_two_select.erb
nyahehnagi: Thanks. Will do.

app.rb
nyahehnagi: Good idea. I'll do that moving forward.

app.rb
nyahehnagi: will do. I think I did it this way due to the week challenge. I'll be mindful moving forward

spec/game_spec.rb
nyahehnagi: Cheers, I went to town on the testing. :-)

PR #2050, opened 2022-02-13
Francesco's attempt

now-what:
* I started by developing the interactive side: entering player's name and move, and storing the data in the same session and in instance variables.
* I then extracted some logic and created a `Player` class which accepts two arguments (`name` and `move`).
* I created a separate `Game` class to run the logic of the game, and that's when things started to get a bit confusing.
* The `if` statement that runs the logic of the game always returns `nil`, so the unit test fails.
* Had I got this to work, I wouldn't know how to render this logic on the `.erb` files.
* With hindsight I might have been better off with just one class and simpler logic.
* Because of time constraint, I polished a little the front-end, and submitted unfinished.


Gemfile
```diff
@@ -3,6 +3,8 @@ source 'https://rubygems.org'
 ruby '3.0.2'
```
trushnarabadia, line 3: I have run Rspec, we have 1 failure, test was expecting Frankie to win, however a draw was returned. Also Rubocop had some minor formatting errors. I have also run rackup, the interface looks great! Absolutely love the layout. Two main things pick up from the game, it may be a good idea to use buttons instead of textboxes, to disable the user from other inputs. Also the game doesn't return the outcome, but I can see you've left notes in regards to this.

README.md
```diff
@@ -1,86 +1,26 @@
...
-  # Want a nice code coverage website? Uncomment this next line!
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
+Process
+-------
 
-You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
+* I've set up the Sinatra project as per Week 3's learnings.
```
trushnarabadia, line 19: Great explanation and breakdown of process, has definitely explained to me how you reached to where you got, as well as how you tried to proceed and encountered issues.

Gemfile
```diff
@@ -3,6 +3,8 @@ source 'https://rubygems.org'
...
 
 gem 'sinatra'
+gem 'sinatra-reloader'
```
trushnarabadia, line 6: File naming and set up all correct and matching review document.

lib/player.rb
```diff
@@ -0,0 +1,8 @@
+class Player
+  attr_reader :name, :move
+
+  def initialize(name, move = "not yet picked")
```
trushnarabadia, line 4: Loving the default here, it's a great idea

public/style.css
```diff
@@ -0,0 +1,34 @@
+h1, h2, h3, button, p, ::placeholder {
```
trushnarabadia, line 1: Your design aspect is outstanding, it really does look great!

PR #2052, opened 2022-02-13
RPS weekend challenge - mmguinness pull request

mmguinness:


spec/features/select_option_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+feature 'select option' do
+  scenario 'player has selected one option' do
+    visit('/')
+    fill_in :player_name, with: 'Marie'
+    click_button 'Play'
+    choose(option: 'scissors')
+    click_button 'Play'
+    expect(page).to have_content 'scissors'
+  end
+end
```
EdCope, line 12: Could add a feature test to check the result screen displays the correct result.

lib/result.rb
```diff
@@ -0,0 +1,21 @@
+class Result
+  
+  def initialize
+    @win_map = {
+      'scissors' => 'paper',
+      'paper' => 'rock', 
+      'rock' => 'scissors'
+    }
+  end
```
EdCope, line 9: This is a nice way of handling the win/loss conditions!

lib/result.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def winner(p1, p2)
+    if p1 == p2
+      "It's a draw!"
+    elsif @win_map[p1] == p2
+      "You won!"
+    else
+      "I won!"
```
EdCope, line 17: Statement is getting a bit long perhaps write it in a format like return "its a draw" if same?

Extract the comparisons to a private method.

lib/game.rb
```diff
@@ -0,0 +1,16 @@
+require './lib/player'
+require './lib/result'
+require './lib/random'
+
+class Game
+  attr_accessor :computer
+
+  def initialize
+    @computer = Random.new.random_option
```
EdCope, line 9: Could the random class be passed in as variable on creation. So that you don't need to require it at the top.

Possibly call the class Computer too ?

lib/random.rb
```diff
@@ -0,0 +1,11 @@
+class Random
```
EdCope, line 1: mentioned above but the name of class random to me sounds more like a method of an object.

spec/random_spec.rb
```diff
@@ -0,0 +1,12 @@
+require './lib/random'
+
+describe Random do
+
+  describe '#random_option' do
+    it 'should chose a random option' do
+      round = Random.new
```
EdCope, line 7: Is creating the object like this required for the stub() method below. I haven't used this before

views/start.erb
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
```
EdCope, line 1: The layout is looking great in general :)

EdCope (commented):
Looks great and having a proper interface really helps it bring it to life.

PR #2056, opened 2022-02-13
RPS Challenge - Camilla

Camilla-Greene:


app.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  get '/signin' do
+    session[:name] = params[:name]
+    erb(:signin)
+    
+  end
+
+  get '/play' do
+    @name = session[:name]
```
Rakshak81, line 25: Could have use sessions here to save the user input

lib/player.rb
```diff
@@ -0,0 +1,9 @@
+class Player
+  attr_accessor :name
```
Rakshak81, line 2: Player Class has not been used

lib/rps.rb
```diff
@@ -0,0 +1,29 @@
+require_relative './player.rb'
+
+# class Rps
+
+#   def result 
+#     if @name == 'rock' && player_2 == 'scissors'
```
Rakshak81, line 6: Use session variable and the OR (||) operator

PR #2072, opened 2022-03-11
Maria's RPS Challenge

ruiined:
# Marija Gromovaja
### Steps Completed

- [x] Registers a player's name
- [x] Player plays versus the computer (random choices)
- [x] Keeps up the score
- [x] Shows the result of each round 

### Readme
- [x] Overview
- [x] Installation
- [x] Screenshot
- [x] Credits


views/game.erb
```diff
@@ -0,0 +1,147 @@
...
+  }
+  i {
+    color: #87CEEB;
+    font-style: normal;
+  }
+  u {
+    border-bottom: 8px solid #87CEEB;
+    text-decoration: none;
+  }
+</style>
```
r94o, line 113: Have no idea what I have just read but it looks great! Nice one!

lib/game.rb
```diff
@@ -0,0 +1,80 @@
+require_relative 'player'
+
+class Game
+  attr_reader :player, :result
+
+  def initialize(player)
+    @player = player
```
r94o, line 7: good to use dependency injection here

lib/game.rb
```diff
@@ -0,0 +1,80 @@
...
+    computer
+    if computer_picks_rock
+      lose
+    elsif computer_picks_paper
+      win
+    elsif computer_picks_scissors
+      tie
+    end
+  end
+
```
r94o, line 47: I think guard conditions here would make this a bit more readable and less lines

lib/player.rb
```diff
@@ -0,0 +1,18 @@
+class Player
+  attr_reader :name, :score
+
+  DEFAULT_POINTS = 10
+
```
r94o, line 5: Good use of constant to avoid 'magic' numbers

r94o (commented):
Looks good! Tests are all isolated and the CSS is really impressive. I feel perhaps some refactoring could be done to make the logic of who wins a bit easier to read (a new class?)

PR #2073, opened 2022-03-13
Nico's RPS Game

NicolaCortese:


app.rb
```diff
@@ -0,0 +1,40 @@
...
+  get '/game' do
+    @player_name = $game.player.name
+    erb(:game)
+  end
+
+  post '/result' do
+    $player_answer = params[:player_answer]
+    redirect 'result'
+  end
+
```
PhilBligh, line 30: I assume people will have an issue with the global variables, which I used too. But I think once player_answer is in params, it can always be accessed from there, so no need for $player_answer

spec/features/rpsgame_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  scenario 'returns the pc answer' do
+    sign_in
+    choose 'Paper'
+    click_on 'Submit'
+    #this stubbing doesn't work
+    allow($game).to receive(:sample).and_return("Rock")
+    expect(page).to have_content "The computer selected"
+  end
+
```
PhilBligh, line 45: These tests are very good and comprehensive, and well done for attempting the stubbing, which I wanted to try but was too confused.

views/game.erb
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
```
PhilBligh, line 5: Wasn't sure what meta charset="utf-8" was. Just curious!

PR #2076, opened 2022-03-13
Gaya's RPS Challenge

gdrmvrm:
I would really appreciate feedback on the following:
- how to mock a feature test
- if I have missed any obvious tests
- if my tests look okay? (as I am still struggling with writing tests)

Thanks,
Gaya


spec/features/declare_winner_spec.rb
```diff
@@ -0,0 +1,17 @@
+# require_relative 'web_helpers'
+
+# #Not sure how to mock a feature test
+
+
+xfeature 'Declare winner' do
+  scenario "Player & Comp choose" do
+    sign_in_and_play
+    click_button 'Rock'
+    allow_any_instance_of(Comp).to receive(:comp_choice).and_return(:rock)
```
leoht, line 10: Normally you wouldn't mock or stub in your feature tests, but as we're dealing with randomness here, it's probably an OK thing to do — this technique should work, however the stubbed return value should be one of the possible values returned by `Comp#comp_choice` - which is a string (e.g `"Rock"`) and not the symbol `:rock`, this is maybe what was causing the problem

leoht: Also, the stubbing should probably happen before you do any user action in the feature test, otherwise you might run into a situation where you run the action that leads to the game result (`click_button 'Rock'`) *before* the result is actually stubbed, so the stubbed value is not taken into account

leoht: Another thing I've missed - it seems you're stubbing the wrong method on `Comp` (it is named `make_choice`, not `comp_choice` - probably a mixed up between the two method names on different classes)

spec/features/declare_winner_spec.rb
```diff
@@ -0,0 +1,17 @@
+# require_relative 'web_helpers'
+
+# #Not sure how to mock a feature test
+
+
+xfeature 'Declare winner' do
+  scenario "Player & Comp choose" do
```
leoht, line 7: It might be good to write at least two other scenarios, maybe one where the player is winning, and one where they loose, as for now you're only testing the "draw" scenario

app.rb
```diff
@@ -0,0 +1,38 @@
...
+    player = Player.new(@player_name)
+    game = Game.new(player)
+    @player_choice = params[:choice]
+    @comp_choice = game.comp_choice
+    @winner = game.winner(@player_choice, @comp_choice)
+    erb :play
+   end
+
+    post '/exit' do
+     erb :index
```
leoht, line 33: Minor thing - it's good to be mindful at the consistency of indentation size, this is something you can easily check & fix by running Rubocop, for example

lib/game.rb
```diff
@@ -0,0 +1,24 @@
...
+    @comp = comp
+  end
+
+  def comp_choice
+    return @comp.make_choice
+  end
+  
+  def winner(player_choice, comp_choice)
+    if (player_choice == "Rock" && comp_choice == "Scissors") || (player_choice == "Scissors" && comp_choice == "Paper") || (player_choice == "Paper" && comp_choice == "Rock")
+      return "You win!"
```
leoht, line 17: I'd probably suggest moving the responsibility of human-readable strings such as `"It's a draw!"` to the view layer (in the .erb file), and let the `winner` method return a simpler value, such as a symbol (e.g `:draw`). This is because we ideally want a clear separation of responsibilities, and having these strings here means that `Game` does the job of computing the winner, but also of displaying the message to the user — and we'd like to leave this to the view. The .erb file could display these different messages depending of the value returned earlier by `Game.winner`, this way if we'd like to change the displayed message (or even translate them, for an international game?), we'd just need to make changes to the view

leoht (commented):
Great work! I've left comments on the points you wanted specific feedback on, plus a few others. Overall the tests are well written, but it'd be good to add a few more testing cases on the feature test, and quickly check the indentation / format, just to make the code a bit easier to read. Happy to answer any other questions you might have :)

PR #2081, opened 2022-03-14
Jonny Abrams RPS

jonnyabrams:


app.rb
```diff
@@ -0,0 +1,39 @@
...
+    erb :play
+  end
+
+  post '/choice' do
+    $player.input_choice(params[:choice])
+    redirect '/result'
+  end
+
+  get '/result' do
+    @game = Game.new($player)
```
adamwoodcock98, line 33: Perhaps `@game` could've been the global variable and initialised on line 18 and done something like:
```ruby
$game = Game.new(Player.new(params[:name]))
```

This is still very readable, and my advice may be poor, but I think with the hierarchy of your different objects, as `Game` stores a `Player` object, I think the global should be the class that sits at the top of the hierarchy. Again, I could be wrong so don't take this as gospel.

lib/game.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'player'
+
+class Game
+  CHOICES = ['rock', 'paper', 'scissors'].freeze
+  OUTCOMES = {
```
adamwoodcock98, line 5: Nice! Very interesting to see an approach entirely different from my own.

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def draw?
+    result == 'draw'
+  end
+
+  def lose?
+    result == 'lose'
+  end
+
+  def result
```
adamwoodcock98, line 33:
```suggestion
private

  def result
```

adamwoodcock98: This is only called from within other methods in the class, so there's no reason for it to not be private. This could still be tested via the returns of `win?` `lose?` or `draw?`.

lib/player.rb
```diff
@@ -0,0 +1,11 @@
...
+  attr_reader :name, :choice
+  
+  def initialize(name)
+    @name = name
+  end
+
+  def input_choice(player_choice)
+    @choice = player_choice
+  end
+end
```
adamwoodcock98, line 11: Perfect!

spec/features/enter_name_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  context 'player enters name and clicks submit' do
+    before do
+      sign_in
+    end
+
+    scenario 'player sees their name returned returned' do
+      expect(page).to have_content 'Hello Jonny'
+    end
+
+    scenario 'it prompts choice of weapon' do
```
adamwoodcock98, line 16: Weapon 😂

spec/features/enter_name_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      expect(page).to have_content 'Choose your weapon'
+    end
+  end
+end
```
adamwoodcock98, line 20: All looks great here, good use of web helpers.

spec/features/play_spec.rb
```diff
@@ -0,0 +1,31 @@
+feature 'Playing a game of rock paper scissors' do
+  context 'player makes choice' do
+    before do
```
adamwoodcock98, line 3: Great touch calling this in a before block, noted for future use myself!

spec/features/play_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    scenario 'make choice' do
+      choose 'rock'
+      click_button 'Submit'
+      expect(page).to have_content 'You chose rock'
+    end
+
+    scenario 'make choice' do
+      choose 'paper'
+      click_button 'Submit'
```
adamwoodcock98, line 21: Love these 3 scenarios, I actually totally spaced on testing that so well done, great coverage!

lib/game.rb
```diff
@@ -0,0 +1,34 @@
...
+    'rock' => { 'rock' => 'draw', 'paper' => 'lose', 'scissors' => 'win' },
+    'paper' => { 'rock' => 'win', 'paper' => 'draw', 'scissors' => 'lose' },
+    'scissors' => { 'rock' => 'lose', 'paper' => 'win', 'scissors' => 'draw' }
+  }.freeze
+
+  def initialize(player)
+    @player = player
+  end
+  
+  def computer_choice
```
adamwoodcock98, line 15: I feel this logic and associated attributes above would be great extracted into a separate `CPU` class, I touch on this in more detail a bit further down.

spec/game_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'game'
+
+describe Game do
+  subject(:game) { described_class.new(player) }
+  let(:player) { double :player }
+
```
adamwoodcock98, line 6: I guess this is where you were struggling a little with mocking as there's a few methods in the `Game` class that are untested. You could try stubbing shorthand syntax:
```ruby
let(:player) { double(:player, :choice => "rock") } #creating a let :player, then a double with the same name that responds to the method choice with rock
```
This could then be injected into the `initialize` method of your `Game` class.

This then raises the question of stubbing the randomness of your CPU choice (we always must stub randomness). If you're CPU logic was in a separate class, you could inject the CPU:

```ruby
class Game
def initialize(player, player2 = CPU.new)
...
end
```
> This also works nicely because there would only be very minimal changes required to make this multiplayer, as the CPU is only default, and another player object could easily be passed in, especially if your choices used polymorphism to be called in the same way across the player and CPU.

You could now inject the CPU stub:
```ruby
let(:cpu_double) { double(:cpu_double, :computer_choice => "scissors") }

adamwoodcock98 (commented):
Very nicely done Jonny, there was one particular element with an additional class being extracted, I don't think you particularly violated SRP, but it would help with testing. Hopefullt this may have given you some help with the mocking issues, but also hope it doesn't seem like I was trying to tell you how to suck eggs.

We did a lot of things very similarly, so it was nice to have some reinforcement of the thought process. 

But all-in-all a very good job, your code is extremely neat and tidy, I really like how you lay things out and your adherence to the DRY principle is absolutely excellent!

PR #2082, opened 2022-03-14
Chris' RPS challenge

clovellbsc:
I completed the main and multiplayer parts of the challenge, however, I know my code is not DRY as I did not have time to go back and refactor my code to work with both the multiplayer code and single player without repetition

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,53 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/player'
+require './lib/rock_paper_scissors'
+
+class BookmarkManager < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get "/" do
+    erb :index
+  end
+
+  get "/solo-game" do
+    erb :solo_name
+  end
+
+  post "/solo-game-name" do
+    player1 = Player.new(params[:player_1_name])
+    player2 = Player.new("Computer")
+    $rps = RPS.new(player1, player2)
+    redirect "/solo-game-choice"
+  end
+
+  get "/solo-game-choice" do
+    @rps = $rps
+    erb :solo_game_choice
+  end
+
+  post "/solo-choice" do
+    @rps = $rps
+    @rps.player1.pick(params[:selection])
+    @rps.random_choice
+    @rps.check_winner_or_draw
+    redirect "/results"
+  end
+
+  get "/results" do
+    @rps = $rps
+    erb :solo_results
+  end
+
+  # get "/multiplayer-game" do
+  #   erb :multiplayer_names
+  # end
+
+
+  
+
+
+  run! if app_file == $0
+end
--- a/app.rb
+++ b/app.rb
@@ -33,20 +33,53 @@ class BookmarkManager < Sinatra::Base
     @rps.player1.pick(params[:selection])
     @rps.random_choice
     @rps.check_winner_or_draw
-    redirect "/results"
+    redirect "/solo-results"
   end
 
-  get "/results" do
+  get "/solo-results" do
     @rps = $rps
     erb :solo_results
   end
 
-  # get "/multiplayer-game" do
-  #   erb :multiplayer_names
-  # end
+  get "/multiplayer-game" do
+    erb :multiplayer_game_names
+  end
 
+  post "/multiplayer-game-names" do
+    player1 = Player.new(params[:player_1_name])
+    player2 = Player.new(params[:player_2_name])
+    $multiplayer_game = RPS.new(player1, player2)
+    redirect "/player1-choice"
+  end
 
-  
+  get "/player1-choice" do
+    @game = $multiplayer_game
+    erb :player1_choice
+  end
+
+  post "/player1" do
+    @game = $multiplayer_game
+    @game.player1.pick(params[:selection])
+    redirect "/player2-choice"
+  end
+
+  get "/player2-choice" do
+    @game = $multiplayer_game
+    erb :player2_choice
+  end
+
+  post "/player2" do
+    @game = $multiplayer_game
+    p @game
+    @game.player2.pick(params[:selection])
+    redirect "/multiplayer-results"
+  end
+
+  get "/multiplayer-results" do
+    @game = $multiplayer_game
+    @game.check_winner_or_draw
+    erb :multiplayer_results
+  end
 
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -3,7 +3,7 @@ require 'sinatra/reloader'
 require './lib/player'
 require './lib/rock_paper_scissors'
 
-class BookmarkManager < Sinatra::Base
+class RockPaperScissorsApp < Sinatra::Base
   configure :development do
     register Sinatra::Reloader
   end
@@ -32,6 +32,7 @@ class BookmarkManager < Sinatra::Base
     @rps = $rps
     @rps.player1.pick(params[:selection])
     @rps.random_choice
+    p @rps.random_choice
     @rps.check_winner_or_draw
     redirect "/solo-results"
   end
@@ -70,7 +71,6 @@ class BookmarkManager < Sinatra::Base
 
   post "/player2" do
     @game = $multiplayer_game
-    p @game
     @game.player2.pick(params[:selection])
     redirect "/multiplayer-results"
   end
--- a/lib/rock_paper_scissors.rb
+++ b/lib/rock_paper_scissors.rb
@@ -1,0 +1,34 @@
+require_relative 'player'
+
+class RPS
+
+  attr_reader :winner_or_draw, :player1, :player2
+
+  def initialize(player1 = Player.new("Chris"), player2 = Player.new("Coding"))
+    @winner_or_draw = nil
+    @player1 = player1
+    @player2 = player2
+  end
+
+  def check_winner_or_draw
+    win_matrix = { 'Rock' => 'Scissors', 'Paper' => 'Rock', 'Scissors' => 'Paper'}
+    if @player1.move == @player2.move
+      @winner_or_draw = 'Draw'
+    elsif win_matrix[@player1.move] == @player2.move
+      @winner_or_draw = "#{@player1.name} wins!"
+    else
+      @winner_or_draw = "#{@player2.name} wins!"
+    end
+  end
+
+  def random_choice
+    @player2.pick(randomise)
+  end
+
+  private
+
+  def randomise 
+    ['Rock', 'Paper', 'Scissors'].sample
+  end
+
+end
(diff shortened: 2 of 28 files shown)


app.rb
```diff
@@ -0,0 +1,86 @@
...
+  get "/solo-game-choice" do
+    @rps = $rps
+    erb :solo_game_choice
+  end
+
+  post "/solo-choice" do
+    @rps = $rps
+    @rps.player1.pick(params[:selection])
+    @rps.random_choice
+    @rps.check_winner_or_draw
```
jonnyabrams, line 35: Is this logic that could be banished to the model layer? (I don't know, I just like the sound of that question)

app.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  get "/solo-results" do
+    @rps = $rps
+    erb :solo_results
+  end
+
+  get "/multiplayer-game" do
+    erb :multiplayer_game_names
+  end
```
jonnyabrams, line 47: Very impressed you got it multiplayering. My brain wouldn't let me

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,34 @@
...
+
+class RPS
+
+  attr_reader :winner_or_draw, :player1, :player2
+
+  def initialize(player1 = Player.new("Chris"), player2 = Player.new("Coding"))
+    @winner_or_draw = nil
+    @player1 = player1
+    @player2 = player2
+  end
```
jonnyabrams, line 11: This Coding guy sounds like a loose cannon

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def check_winner_or_draw
+    win_matrix = { 'Rock' => 'Scissors', 'Paper' => 'Rock', 'Scissors' => 'Paper'}
+    if @player1.move == @player2.move
+      @winner_or_draw = 'Draw'
+    elsif win_matrix[@player1.move] == @player2.move
+      @winner_or_draw = "#{@player1.name} wins!"
+    else
+      @winner_or_draw = "#{@player2.name} wins!"
+    end
```
jonnyabrams, line 21: I ended up making a 2D hash with all the possible permutations, including draw and lose. Your way is much better.

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def random_choice
+    @player2.pick(randomise)
+  end
+
+  private
+
+  def randomise 
+    ['Rock', 'Paper', 'Scissors'].sample
+  end
```
jonnyabrams, line 32: Making a separate randomise method for repeated use is a nice touch. Might nick this.

spec/features/multiplayer_results_spec.rb
```diff
@@ -0,0 +1,27 @@
+feature "Playing through to the end of the game" do
+  scenario "shows the randomly generated move of computer" do
+    visit("/")
+    click_button "Pass and Play (2 player mode)"
+    fill_in :player_1_name, with: "Chris"
```
jonnyabrams: Could you have done a helper for these steps?

spec/features/player1_choice_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  scenario "Player 1 choice page displays Player 1 name and displays their turn" do
+      visit("/")
+      click_button "Pass and Play (2 player mode)"
+      fill_in :player_1_name, with: "Chris"
+      fill_in :player_2_name, with: "Coding"
+      click_button 'Submit'
+
+      expect(page).to have_content "Chris it's your turn. Coding, no peeking!"
+  end
+end
```
jonnyabrams, line 15: Lovely readable code and personable interface. Liquid coding

spec/player_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'player'
+
+describe Player do
+
+  subject(:player) { Player.new("Chris") }
+
+  it 'will be initialised with one argument' do
+    expect { Player.new }.to raise_error 
+    expect { player }.to_not raise_error
```
jonnyabrams, line 9: Not seen this in a test before. I like it. (Also nicking this.)

spec/player_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  it 'player stores it a players name' do
+    expect(player.name).to eq("Chris")
+  end
+
+  it 'move intitialises as nil' do
+    expect(player.move).to eq nil
+  end
```
jonnyabrams, line 18: I think it's convention for the 'it' lines to read like a real sentence as far as possible, although I guess this can sometimes make it complicated

jonnyabrams: So maybe it could be something like "it 'initialises the move as nil'" instead

spec/rock_paper_scissors_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rock_paper_scissors'
+
+describe RPS do
+  
+  subject(:game) { RPS.new }
+  let(:mocked_game) { RPS.new(chris, coding) }
+  let(:chris) { double(:chris, name: "Chris", move: 'Paper') }
+  let(:coding) { double(:coding, name: "Coding", move: 'Rock') }
```
jonnyabrams, line 10: Textbook mocking and doubling. I'd stand and applaud if my cat wasn't sitting on me

app.rb
```diff
@@ -0,0 +1,86 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/player'
+require './lib/rock_paper_scissors'
+
+class BookmarkManager < Sinatra::Base
```
leoht: Seems like the app class could be renamed to maybe `RockPaperScissorsApp`?

app.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  get "/player2-choice" do
+    @game = $multiplayer_game
+    erb :player2_choice
+  end
+
+  post "/player2" do
+    @game = $multiplayer_game
+    p @game
```
leoht: Your controllers are definitely "skinny" - they deal only with calling the model layer, getting params from the request, and tying it all together. You could probably remove this `p` line here, as this was used for debugging I guess?

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'player'
+
+class RPS
+
+  attr_reader :winner_or_draw, :player1, :player2
```
leoht, line 5: Minor one - I've noticed you have an interesting pattern of naming variables by referring to the possible values they contain, e.g `@winner_or_draw` or `rock_paper_or_scissors` — in Ruby, we can't really enforce the type of a variable or argument to be restricted to a few values, like in other languages. However we can use guard conditions and checks to verify the value is "allowed" - it would be usually more conventional to name such a variable `@result` rather than use the possible values it can contain, although this can be quite helpful to know which values would be stored in this variable

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,34 @@
...
+  def initialize(player1 = Player.new("Chris"), player2 = Player.new("Coding"))
+    @winner_or_draw = nil
+    @player1 = player1
+    @player2 = player2
+  end
+
+  def check_winner_or_draw
+    win_matrix = { 'Rock' => 'Scissors', 'Paper' => 'Rock', 'Scissors' => 'Paper'}
+    if @player1.move == @player2.move
+      @winner_or_draw = 'Draw'
```
leoht: I'd suggest returning "simple" values (such as symbols) rather than human-readable messages (such as "Chris wins!"), this for different reasons:
 * This we can we can enforce the SRP better (it applies also to the MVC pattern) - the only layer that is responsible for displaying and presenting info to the user should be the "view" layer (the ERB templates), so any human-readable messages or info should belong there
 * It makes it easier for the caller to use the result of `check_winner_or_draw` and then delegates to the view what to display - if we later choose to create an international game of RPS, and translate it in other languages, at least we know that we only need to update the view layer to add these translations

spec/features/solo_results_spec.rb
```diff
@@ -0,0 +1,15 @@
+feature "Playing through to the end of the game" do
+  scenario "shows the randomly generated move of computer" do
+    solo_submit_name_select_rock
```
leoht: In this case I'd suggest two things:
 * Add arguments to the helper method `solo_submit_name_select_rock` so we can explicitly pass the player name, etc, to it: `solo_submit_name_select_rock("Chris")` for example - otherwise we need to jump to the helper method definition, see the value that is given, and come back to the test to verify it all adds up - which adds some cognitive load
 * We usually don't mock or stub things in feature tests, but in this case we're dealing with randomness, so it is ok to stub the random move from the program to assert the different scenarios - e.g if the player choose rock and you stub the random choice to be paper, you'll get a "lose" result

views/solo_results.erb
```diff
@@ -0,0 +1,14 @@
...
+<strong> The computer </strong> chose <em> <%= @rps.player2.move %> </em>
+<br> <br>
+<h4><strong> <%= @rps.check_winner_or_draw %> </strong> </h4>
+
+
+<form action="/solo-game-choice" method="get">
+  <input type="submit" value="Play another solo game">
+</form>
+
+<form action="/" method="get">
```
leoht, line 18: Since the form's method is GET and you don't have any parameters, a link would be a simpler way to get the same behaviour (it would make the browser send a GET request to this path - and display the response's HTML):
```
<a href="/">Home</a>
```

spec/rock_paper_scissors_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  
+  it 'should initialise with a choice winner of nil' do
+    expect(game.winner_or_draw).to be_nil
+  end
+
+  it 'should be able to calculate the winner of the game' do
+    expect { mocked_game.check_winner_or_draw }.to change { mocked_game.winner_or_draw }.from(nil).to("#{chris.name} wins!")
```
leoht: Looks like this test is checking the internal attribute `winner_or_draw` (because the `change` matcher "watches" the value of `mocked_game.winner_or_draw`, and assert the original and final values) - in other words, you're testing the internal state rather than the behaviour. A simple way to improve this and test for "behaviour" instead is to assert on the return value of `mocked_game.check_winner_or_draw`, which would be the result value

spec/rock_paper_scissors_spec.rb
leoht: Great use of doubles to mock the game's behaviour indeed :)

spec/features/solo_results_spec.rb
leoht: (Btw, the first point is also true for other helper methods!)

leoht (commented):
Great work, your code is well written, tested and organised in the different layers of MVC - I just had a few comments on a few ways to improve and do some minor touches, but overall well done :) Let me know in case you have any further questions!

Re: having the controller routes a bit more DRY (between solo game and multiplayer game), it's fine to have it this way first, it's more explicit and simpler to understand the code. Once you start to see some common patterns or ways to refactor some parts the code, you can have a go and making it a bit more DRY - the good news is that since our code is tested, we can do this without too much worry!)

PR #2085, opened 2022-03-14
Patryk Okpara rps-challenge

Lorn888:
Completed all user stories but didnt start bonus challenges.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,14 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+
+class BookmarkManager < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    'Hello World'
+  end
+
+  run! if app_file == $0
+end
--- a/app.rb
+++ b/app.rb
@@ -7,7 +7,13 @@ class BookmarkManager < Sinatra::Base
   end
 
   get '/' do
-    'Hello World'
+    erb(:form)
+  end
+
+  post '/name' do
+    @player_name = params[:player_name]
+    p params
+    erb(:name)
   end
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -1,5 +1,6 @@
 require 'sinatra/base'
 require 'sinatra/reloader'
+require './lib/game'
 
 class BookmarkManager < Sinatra::Base
   configure :development do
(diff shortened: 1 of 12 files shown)


app.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  post '/game' do
+    @player_pick = params[:button]
+    pick = Game.new
+    @computer_pick = pick.computer_pick
+    
+    @result = pick.decision(@player_pick, @computer_pick)
+    
+    erb(:game)
+  end
```
PKilgarriff, line 28: Putting a reference to our discussion on the call - this would be the sort of thing that it might be good to extract away from the Controller and View, and instead into the decision method of the Game class.

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+  def computer_pick
+    number = rand(1..3)
+    if number == 1
+      'Rock'
+    elsif number == 2
+      'Paper'
+    elsif number == 3
+      'Scisors'
+    end
+  end
```
PKilgarriff, line 12: The 'sample' method of the Array class could be useful for refactoring this method

lib/game.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def decision (player_pick, computer_pick)
+    if player_pick == computer_pick
+      'its a tie!'
+    elsif player_pick == 'Rock' && computer_pick == 'Paper'
+      'you lost!'
+    elsif player_pick == 'Rock' && computer_pick == 'Scisors'
+      'you won!'
```
PKilgarriff, line 20: I like that you have explicitly covered all the possible outcomes, but this also might be a good place for a switch statement if you are looking to cut down on lines of code.

app.rb
```diff
@@ -0,0 +1,32 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/game'
+
+class BookmarkManager < Sinatra::Base
```
PKilgarriff, line 5: Your controller class doesn't have to be called BookmarkManager, I realised this after following the set-up guide too!

PKilgarriff (commented):
Nicely done - your classes have clear responsibilities and you've implemented the user stories without extra faff and unnecessary features. As we said on the call - there are a few things you are planning to do when you refactor, but it works as it stands which is a good place to have it!

PR #2086, opened 2022-03-14
Joe Thornton - rps challenge

joe-thornton:
Joe Thornton - rps challenge

Further work needed to mock out classes in unit tests


lib/game.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def change_player
+    @turn_manager.change_turn
+  end
+
+  def reset_turns
+    @turn_manager.reset
+  end
+
+  def determine_winner(player1choice = @players[0].choice, player2choice = @players[1].choice)
```
anushanesargi, line 35: You could consider  the reslut_manager class to determine the winner

lib/game.rb
```diff
@@ -0,0 +1,46 @@
...
+  def new_player(name)
+    @players << Player.new(name) # Need to mock this out at some point
+  end
+
+  def reset_players
+    @players = []
+  end
+
+  def change_player
+    @turn_manager.change_turn
```
anushanesargi, line 28: I feel, you could use a class variables/arrays to store player names rather than creating a class for switching turns. Because every player is getting a chance to select their option in every turn.

app.rb
```diff
@@ -0,0 +1,45 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require_relative './lib/game'
+require_relative './lib/player'
+
```
anushanesargi, line 5: - Overall, it is a nice piece of code.
- Clean and quite a lot of SRP
- No global variables.

app.rb
```diff
@@ -0,0 +1,45 @@
...
+    @game.new_player(params[:player2_name])
+    redirect '/play'
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post '/player-choice' do
+    @game.players[@game.turn_manager.turn].assign_choice(params[:radio_button])
```
anushanesargi, line 28: For me this line took the most time to understand. I cannot find a more intelligent way of solving the problem other than assigning some to a variable.

PR #2091, opened 2022-03-14
Challenge

felixtaljaard:


app.rb
```diff
@@ -0,0 +1,32 @@
...
+      redirect '/play'
+    end
+
+    get '/play' do
+      @turn = Turn.new(session)
+      erb :play
+    end
+
+    post '/play' do
+      session[:player_shape] = params[:shape].downcase.to_sym
```
HarryClenyg, line 26: I think that storing the player's move as a 'session' means it can only remember one move per session. If you were to then want to make a best out of 3 version it wouldn't then work.

lib/turn.rb
```diff
@@ -0,0 +1,26 @@
+class Turn
+  GAME_RULES = {
```
HarryClenyg, line 2: This is so simple and clean.

spec/features/play_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    scenario "opponent chooses a random option" do
+      srand(221563)
+      click_button 'Rock'
+      expect(page).to have_content "Opponent chose Scissors!"
+    end
+
+    scenario "I win" do
+      click_button 'Rock'
+      expect(page).to have_content
```
HarryClenyg, line 35: is this supposed to have no specific content?

spec/features/enter_names_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+feature "entering names" do
+  scenario "enter names and see my name" do
+    visit('/')
+    fill_in :player_name, with: "Felix"
+    click_button "Submit"
```
HarryClenyg, line 7: these 3 lines could've been added to a web helper

PR #2101, opened 2022-04-11
Ruzeka's RPS

R552-beep:


app.rb
```diff
@@ -0,0 +1,38 @@
...
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    session[:player_name] = params[:player_name]
```
JordanManu, line 16: great!

app.rb
```diff
@@ -0,0 +1,38 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    @player_name = session[:player_name]
```
JordanManu, line 21: Really good

app.rb
```diff
@@ -0,0 +1,38 @@
...
+    erb :play
+  end
+
+  post '/choose' do
+    session[:option] = params[:option]
+    redirect to '/result'
```
JordanManu, line 27: Good use of get & post methods

spec/features/choice_spec.rb
```diff
@@ -0,0 +1,10 @@
+require_relative '../../app'
+
+feature "#choices" do
+  scenario "the player can select rock, paper or scissors" do
+    sign_in_and_play
```
JordanManu, line 5: good use of sign in and play

spec/features/player_spec.rb
```diff
@@ -0,0 +1,9 @@
+require_relative '../../lib/player'
+
+describe Player do
+  it "returns the player name" do
+    player = Player.new()
```
JordanManu, line 5: If you're using subject below you can probably delete this line

views/result.erb
```diff
@@ -0,0 +1 @@
+You selected <%= @choice %>!
```
JordanManu, line 1: Overall I think your code is really good! It would be great if there was a computer option also so that when the user selects their weapon it gives them a result.
But with the code that you have done, I think you've done a great job. Very good readability

PR #2105, opened 2022-05-07
Amfibiya PR

amfibiya17:
[rock-paper-scissors](https://rock---paper---scissors.herokuapp.com/) game


views/index.erb
```diff
@@ -0,0 +1,33 @@
...
+    <br />
+    <br />
+    <br />
+    <form action="/name" method="post">
+      <div class="input">
+        <input type="text" name="player" placeholder="Your name..." />
+      </div>
+      <br />
+      <div class="button1 action_btn1">
+        <button name="submit" class="action_btn submit" type="submit" value="Submit">Submit</button>
```
AKCDNG, line 22: The code at the moment allows you to submit without adding a name. Can be fixed using `required` in the `button` section of the HTML.

app.rb
```diff
@@ -0,0 +1,40 @@
...
+
+class Game < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+  
+  post '/name' do
+    p params
```
AKCDNG, line 15: May not need to have `p params` in the code if used just for troubleshooting.

AKCDNG (commented):
Loved the project. Incredible presentation, lovely CSS & the code is so clean, easy to understand and very organised.

app.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  get '/game' do
+    @play = session[:play]
+    erb :game
+  end
+
+  post '/game' do
+    @play = session[:play]
+    @play.player.player_choice = params[:player_choice]
```
z-rasool, line 29: I can understand using `play` as the name of your subject, but it may cause some confusion / readibility e.g. `play.player`. Perhaps how I would approach it would be to consider what the "objects" are, rather than the "actions" if that makes sense, but if not let me know and I'll explain better!

views/index.erb
z-rasool: Hmm I didn't have this issue when I tried playing it, it did seem like it is required.

spec/computer_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  subject(:computer) { Computer.new }
+
+  describe "#random figure" do
+    it "defines random figure" do
+      expect(computer).to respond_to(:computer_choice)
+    end
+
+    it "returns random figure" do
+      expect(["rock", "paper", "scissors"]).to include(computer.random_choice)
```
z-rasool, line 13: I think you might want to check this test again - it seems to be checking that an array of multiple options includes the computer choice, when what you want is to check that only one choice is returned out of a set of options?

z-rasool (commented):
Great UI and good testing that covers most cases, will be good to check some of the tests to see they are testing what is intended, and that code is readable. Very impressed with test coverage and completedness of the task, plus loving the UI. Well done.

PR #2106, opened 2022-05-08
Luiza's rps-challenge

LGretzk:


spec/game_spec.rb
```diff
@@ -0,0 +1,41 @@
+require './lib/game'
```
K-Carty, line 1: I love to see how comprehensive your testing is. Do you think that perhaps some of these tests should be in a separate unit-tests folder?

spec/game_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  subject(:game) { Game.new }
+  
+  describe '#computer_picked' do
+    it 'has three options' do
+      expect(Game::OPTIONS).to eq ['rock', 'paper', 'scissors']
+    end
+
+    it 'randomly picks either rock, paper or scissors' do
+      expect(game.computer_picked).to eq('rock').or eq('scissors').or eq('paper')
```
K-Carty, line 13: your code is easy to understand and read

lib/game.rb
```diff
@@ -0,0 +1,32 @@
...
+    @computer_picked = random_pick
+  end
+
+  def play(choice)
+    p1 = choice
+    p2 = computer_picked
+
+    if (p1 == "scissors" && p2 == "paper") || (p1 == "rock" && p2 == "scissors") || 
+      (p1 == "paper" && p2 == "rock")
+        @outcome = "You have won!"
```
K-Carty, line 15: As per the review.md: Your Game class (or similar) should not return presentation strings like "Congratulations - you won!". This is a presentation concern and should be handled in another layer of code (separation of concerns). Instead, return representative codes, such as :win and :draw from the Game class which can be translated by the presentation layer.

lib/game.rb
```diff
@@ -0,0 +1,32 @@
...
+      (p1 == "paper" && p2 == "scissors")
+        @outcome = "You have lost!"
+    else 
+        @outcome = "It's a draw"
+    end
+  end
+
+  private
+
+  OPTIONS = ['rock', 'paper', 'scissors']
```
K-Carty, line 26: By creating a Computer class, you can take advantage of duck-typing in the game class. The game does not need to know if it's comparing two players or one player vs a computer or even two computers!

K-Carty (commented):
You've done an amazing job, especially with the extra touches. you should be proud.

PR #2107, opened 2022-05-08
Tim's Rock, Paper, Scissors

timbo-graham:
Would like to have refactored more and completed the bonus challenges, but didn't have as much time as I would've liked this weekend.

Per the README, the final feature test is flaky, as I wasn't able to work out how to stub Capybara rspec tests.


spec/rps_spec.rb
```diff
@@ -0,0 +1,71 @@
+require_relative '../lib/rps'
+
+describe RpsGame do
+  
+  context "A user can play a game of rock, paper, scissors" do
+    
+    let(:rps) { described_class.new }
+
+    it "the Rps class exists" do
```
ErlantzR, line 9: Not sure about the convention behind this but I guess this test is redundant and could be removed?

spec/rps_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      
+      expect(rps).to be_an_instance_of(RpsGame)
+
+    end
+
+    it "the computer_move method randomly chooses a move" do
+
```
ErlantzR, line 16: I think this might not be testing anything due to the stubbing

app.rb
```diff
@@ -0,0 +1,30 @@
+require 'sinatra/base'
```
ErlantzR, line 1: I think all the code is nice, simple and tidy. Easy to understand what it is trying to do all the time

views/outcome.erb
```diff
@@ -0,0 +1,5 @@
+<%= @outcome %><br>
```
ErlantzR, line 1: Only thing to comment is: Does it show the computer choice to the user?

app.rb
```diff
@@ -0,0 +1,30 @@
...
+require_relative './lib/rps'
+
+class Rps < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    
+    erb :index
```
eoinmakers, line 12: Whitespace above and below this line can be removed to adhere to Ruby styling: https://rubystyle.guide

app.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  post '/names' do
+    @player_name = params[:player_name]
+    erb :play
+  end
+
+  post '/result' do
+    @game = RpsGame.new
+    @game.play(params[:move])
+    @outcome = @game.outcome
```
eoinmakers, line 24: Really concise implementation, avoiding the use of global variables! Nice work.

lib/rps.rb
```diff
@@ -0,0 +1,55 @@
+class RpsGame
+
+  attr_reader :outcome
+
+  def initialize
+
+    @win_matrix = {
```
eoinmakers, line 7: Similar points around extra whitespace here.

lib/rps.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  end
+
+  def tie
+
+    @outcome = "It's a draw"
+
+  end
+
+  def lose
```
eoinmakers, line 49: Perhaps one method with a switch statement to return the outcome string will be a more concise implementation?

spec/rps_spec.rb
eoinmakers: Erlantz is correct here - this test is testing a stub. 

For the computer move, you can test that it returns one of three responses (one for each move), and then stub it later to calculate how it compares with the player moves to generate each outcome.

eoinmakers (commented):
Good implementation! You've clearly demonstrated the learning objectives of web week. Well done. 

Be careful with extra whitespace - this is as important is good indentation to keep code tidy and readable. 

Unit tests here could do with a refactor as they're currently testing stubs - you can use those stubs instead to generate and test outcomes for each combination of player/computer move instead.

PR #2108, opened 2022-05-08
Erlantz's rps-challenge

ErlantzR:


app.rb
```diff
@@ -0,0 +1,44 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/player'
+require './lib/judge'
+
+class Game < Sinatra::Base
+
+  enable :sessions
+
+  # configure :development do
```
timbo-graham, line 10: How come you commented this out? Did you use rackup instead?

app.rb
```diff
@@ -0,0 +1,44 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    @name = $player.name
+    erb :play
+  end
+
+  post '/player_choice' do
+    $player.player_choice(params[:choice].downcase)
```
timbo-graham, line 30: Might have been better to use a non-text field for choice, to reduce the chance of errors/mistypes

timbo-graham (commented):
Not too much to say from me really! It's a very clean and clear implementation, with good separation of concerns and nice flourishes :)

README.md
```diff
@@ -1,86 +1,18 @@
 # RPS Challenge
```
timbo-graham, line 1: I guess the README could be a bit more descriptive about your thinking in putting the programme together, how various parts of it interact, etc - to make it easier for other devs working with it (the same is definitely true of mine)

PR #2109, opened 2022-05-08
Farzan's RPS game (2 user stories complete)

Farzan-I:
Farzan Imanzadeh

Completed the below user stories:
```
As a marketeer
So that I can see my name in lights
I would like to register my name before playing an online game

As a marketeer
So that I can enjoy myself away from the daily grind
I would like to be able to play rock/paper/scissors
```

Did not complete the bonus parts (multiplayer and RPSLS)


lib/game.rb
```diff
@@ -0,0 +1,61 @@
+class Game
+  
+  attr_reader :player, :win, :message
+
+  def initialize(player)
+    @player = player
+    @win = nil
+    @message = ""
+  end 
+
```
S-Spiegl, line 10: this feels like neat logic - very easy to read

lib/game.rb
```diff
@@ -0,0 +1,61 @@
...
+        @message = "You've won! Well done!"
+      when false 
+        @message = "You've lost! Try again!"
+      when nil
+        @message = "It's a tie! Try again!"
+    end
+  end
+
+  private
+
```
S-Spiegl, line 56: You could consider keeping the computer in a separate class

features/enter_names_spec.rb
```diff
@@ -8,4 +8,12 @@
 
     expect(page).to have_content("Farzan Vs. Computer")
   end
+
```
S-Spiegl: Good, specific content requirements throughout the tests - I need to tighten mine up.

app.rb
```diff
@@ -0,0 +1,52 @@
...
+    @game.paper
+    session[:message] = @game.message
+    redirect '/play'
+  end 
+
+  post '/scissors' do
+    @game = $game
+    @game.scissors
+    session[:message] = @game.message
+    redirect '/play'
```
eoinmakers, line 47: Currently, you define a route for each player move option, and each of these routes uses mostly the same code. This is a duplication of code that could be refactored. 

Another method of approach this logic is to pass in the player move as a parameter, calling the same route, and passing that param into the @game instance.

lib/game.rb
```diff
@@ -0,0 +1,61 @@
...
+  def message
+    case @win 
+      when true
+        @message = "You've won! Well done!"
+      when false 
+        @message = "You've lost! Try again!"
+      when nil
+        @message = "It's a tie! Try again!"
+    end
+  end
```
eoinmakers, line 53: This approach can be simplified to make this class more concise. Check out the following approach: https://github.com/makersacademy/rps-challenge/blob/main/docs/review.md#use-of-ifelsif-conditionals-for-business-logic

eoinmakers (commented):
Nice work on this! You've demonstrated the learning objectives of the week well.

My feedback revolves around refactoring to remove duplication of code - let me know if you have any questions. 

You have a good range of unit & feature tests here, inc. stubbing the computer's move. Nice work! Fully comprehensive unit testing for this challenge would include unit tests for every combination of player/computer moves.

PR #2110, opened 2022-05-08
Russells - Rock/Paper/Scissors Challenge

Rmorbey:
First Two User Stories Complete.

If I had more time I would:
- make it look nicer
- add multiplayer

Refactored in newest commit
- Changed how my rule engine works, I initially found a way to do it while watching a youtube video. I decided to change it to a Case statement because I wasn't getting the correct outcomes. I also wanted to refactor it my way.
- Successfully added spock/lizard special rules
- It now outputs what the player and the robot chose on the results page, so you can verify the outcome of the match.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -29,31 +29,57 @@ class RPS < Sinatra::Base
 
   post '/rock' do
     rps = RockPaperScissors.new(guess: 'rock')
+    $player_guess = rps.view_player
+    $robot_guess = rps.view_computer
     $pvr_results = rps.winner_is
     redirect '/pvr_results'
   end
 
   post '/paper' do
     rps = RockPaperScissors.new(guess: 'paper')
+    $player_guess = rps.view_player
+    $robot_guess = rps.view_computer
     $pvr_results = rps.winner_is
     redirect '/pvr_results'
   end
 
   post '/scissors' do
     rps = RockPaperScissors.new(guess: 'scissors')
+    $player_guess = rps.view_player
+    $robot_guess = rps.view_computer
+    $pvr_results = rps.winner_is
+    redirect '/pvr_results'
+  end
+
+  post '/spock' do
+    rps = RockPaperScissors.new(guess: 'spock')
+    $player_guess = rps.view_player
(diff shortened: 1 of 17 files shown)


app.rb
```diff
@@ -0,0 +1,86 @@
...
+    @pvr_results = $pvr_results
+    erb :pvrresults
+  end
+
+  post '/pvr_again' do
+    @player_guess = nil
+    @robot_guess = nil
+    @pvr_results = nil
+    redirect '/pvr'
+  end
```
eoinmakers, line 83: You trigger a different route depending on what move the player chooses, and each of these routes runs the same five lines of code - there's an opportunity here to re-use existing code, keeping your implementation DRY. 

An alternative approach would pass the player's chosen move as a param instead, then passing this into sessions or an instance of Player, for example.

lib/rock_paper_scissors.rb
```diff
@@ -0,0 +1,49 @@
...
+        @computer_guess == :paper && @guess == :rock ||
+        @computer_guess == :paper && @guess == :spock ||
+        @computer_guess == :scissors && @guess == :paper ||
+        @computer_guess == :scissors && @guess == :lizard ||
+        @computer_guess == :lizard && @guess == :paper ||
+        @computer_guess == :lizard && @guess == :spock ||
+        @computer_guess == :spock && @guess == :scissors ||
+        @computer_guess == :spock && @guess == :rock
+      'Robot wins'
+    end
```
eoinmakers, line 39: There are a few approaches to the this game logic, and although your logic is easy to read, an alternative approach could be more concise: https://github.com/makersacademy/rps-challenge/blob/main/docs/review.md#use-of-ifelsif-conditionals-for-business-logic

eoinmakers (commented):
Nice work on this! You've achieved the user stories and learning objectives of the week. There's a good range of both feature and unit tests. 

Fully comprehensive unit testing would cover every possible combination of the game rules -  which sounds long, but would provide 100% test coverage. 

There's an opportunity to do some refactoring to reduce duplication of code - see inline comments.

PR #2114, opened 2022-05-09
Kieran's RPS Challenge

K-Carty:
I managed to get my app working but need to add more feature and unit tests although I'm not sure on the rspec syntax how to write them. Also, I think the app could have better functionality by allowing user to click on buttons rather than type to submit their move. 

I need help with testing and more knowledge on html to improve I think.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -11,15 +11,16 @@ class RPS < Sinatra::Base
   end
 
   post '/name' do
-    @name = params[:name1]
+    @name1 = params[:name1]
     erb :game_choices
   end
 
   post '/game' do
+    @player_move = params[:choice]
     game = Game.new 
-    @player_move = params[:choice] 
     @computer_move = game.computer_choice 
-    p game.result(@player_move, @computer_move)
+    @result = game.result(@player_move, @computer_move)
+    "#{@result}"
   end
 
   run! if app_file == $0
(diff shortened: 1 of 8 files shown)


app.rb
```diff
@@ -0,0 +1,64 @@
...
+    @computer_move = game.computer_choice 
+    @result = game.result(@player_move, @computer_move)
+    "#{@result}"
+  end
+
+  run! if app_file == $0
+end
+
+
+class Game
```
LGretzk, line 30: I think the Game class should be separate from the app.rb file

app.rb
```diff
@@ -0,0 +1,64 @@
...
+    cc = (1 + rand(3))
+    case cc
+    when 1
+      choice = "rock"
+    when 2
+      choice = "paper"
+    when 3
+      choice = "scissors"
+    end
+  end
```
LGretzk, line 41: it could be achieved with less code:
[:rock, :paper, :scissors].sample

spec/features/feature_test_spec.rb
```diff
@@ -0,0 +1,36 @@
+# coding: utf-8
+
+require "capybara/rspec"
+require_relative "../../app"
+
+Capybara.app = RPS
```
LGretzk, line 6: I don't think the above is needed in the feature test spec

spec/features/feature_test_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    expect(page).to have_content "Player: #{name1}!"
+  end 
+end 
+
+feature 'Game' do
+  scenario 'can choose one of three choices' do
+    visit("/name")
+    expect(page).to have_button('submit')
+  end
+
```
LGretzk, line 23: the above is not testing for the choice

LGretzk (commented):
Good division into controller and view, although controller needs to be further separated into model (and controller). The app behaves as expected. Some features tests are included, unit tests are missing.

PR #2116, opened 2022-05-09
Luke's RPS Challenge

lukestorey95:
Working 1 player vs computer version
Working 2 player vs player version


app.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  post "/names" do
+    player1 = Player.new(params[:player_1_name])
+    if params[:player_2_name].empty?
+      player2 = Computer.new
+    else
+      player2 = Player.new(params[:player_2_name])
+    end
+    $game = Game.new(player1, player2)
```
jimmy-lyons, line 18: game as global variable?

app.rb
```diff
@@ -0,0 +1,48 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    @game = $game
```
jimmy-lyons, line 23: I'm sure you had your reasons, but it looks to me like you could have avoided a global variable by using a session?

app.rb
```diff
@@ -0,0 +1,48 @@
...
+    erb :play
+  end
+
+  post '/rps' do
+    @game = $game
```
jimmy-lyons, line 28: Do you have to assign this twice? Already assigned on line 23.

lib/game.rb
```diff
@@ -0,0 +1,76 @@
+class Game
+  attr_reader :players, :turn, :round
+
+  def initialize(player_1, player_2, round = Round.new)
+    @players = [player_1, player_2]
+    @turn = players[0]
+    @round = round
+  end
+
```
jimmy-lyons, line 9: are player_1 and player_2 methods necessary? It feels like these can just be instance variables of just stay as items in the players array.

lib/game.rb
```diff
@@ -0,0 +1,76 @@
+class Game
+  attr_reader :players, :turn, :round
+
+  def initialize(player_1, player_2, round = Round.new)
+    @players = [player_1, player_2]
```
jimmy-lyons, line 5: Would it have been cleaner to have two instance variables rather than an array?

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+      round.set_looser(player_2)
+      round.winner.increase_score
+      calculate_outcome_message
+    else
+      round.set_winner(player_2)
+      round.set_looser(player_1)
+      round.winner.increase_score
+      calculate_outcome_message
+    end
+  end
```
jimmy-lyons, line 64: DRY

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+    case round.winner.action
+    when :rock
+      round.set_outcome('smashes')
+    when :paper
+      round.set_outcome('wraps')
+    when :scissors
+      round.set_outcome('cuts')
+    end
+  end
+end
```
jimmy-lyons, line 76: I like these messages

lib/player.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def increase_score
+    @score += 1
+  end
+
+  def computer?
+    self.class == Computer
+  end
+end
+
```
jimmy-lyons, line 29: This is fancy! So you can create an instance of computer and copy in all the methods and values from Player? Love it

app.rb
```diff
@@ -0,0 +1,48 @@
...
+    erb :index
+  end
+
+  post "/names" do
+    player1 = Player.new(params[:player_1_name])
+    if params[:player_2_name].empty?
+      player2 = Computer.new
+    else
+      player2 = Player.new(params[:player_2_name])
+    end
```
eoinmakers, line 17: You mentioned finding it difficult to draw this logic into a Model - firstly, well done on achieving the functionality of the user stories! 

One thing that may make this refactor simpler is thinking about what parts of your program actually need to know about the Player. It's the Game Model that does - in order to know Player 1's move, the Computer's move, or Player 2's move. 

If your Game class was responsible for taking player name params as arguments when initialised, you could use a default argument for player 2 - defaulting to **nil** if no player 2 name is provided. 

Additionally in your Game logic, the code that handles comparing the two player moves to calculate the winner, could then determine if it's going to be using player2's move or generating a computer move if player2 is nil.

What do you think?

app.rb
```diff
@@ -0,0 +1,48 @@
...
+    erb :play
+  end
+
+  post '/rps' do
+    @game = $game
+    @game.new_round
+    @game.turn.throw(params[:throw].to_sym)
+    @game.switch_turn
+    @game.act_for_computer
+    @game.calculate_outcome
```
eoinmakers, line 33: This is a lot of method calls to keep in the route, when perhaps you could push the responsibility of calling each of these methods into the @game instance, so you only need to call something like @game.generate_winner.

lib/game.rb
```diff
@@ -0,0 +1,76 @@
...
+  def calculate_outcome
+    if player_1.thrown_action? && player_2.thrown_action?
+      rps_logic
+    end
+  end
+
+  private
+
+  def rps_logic
+    win_condition = { scissors: :paper, paper: :rock, rock: :scissors }
```
eoinmakers, line 49: Concise handling of game winner logic.

eoinmakers (commented):
You demonstrated the learning objectives of the week - well done. I left a comment around your refactor of controller -> model, let me know what you think. 

Testing of all possible combinations of player / comp moves to test the outcomes would also be a good addition in future projects like this.

PR #2119, opened 2022-05-09
Jimmy Lyons RPS challenge

jimmy-lyons:
Completed the basic functionality (with no frills).

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,5 +1,7 @@
 require 'sinatra/base'
 require 'sinatra/reloader'
+# couldn't get relative path to work for this. need to find cause
+require '/Users/jimmylyons/Documents/Programming/Makers/Projects/week3/rps-challenge/lib/game.rb'
 
 class Online_Game < Sinatra::Base
   configure :development do
@@ -25,6 +27,9 @@ class Online_Game < Sinatra::Base
   end
 
   get '/result' do
+    game = Game.new('$player_choice')
+    @result = game.return_winner
+    $computer_choice
     erb :result
   end
 
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,14 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+
+class Online_Game < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    'Hello World'
+  end
+
+  run! if app_file == $0
+end
--- a/app.rb
+++ b/app.rb
@@ -10,5 +10,14 @@ class Online_Game < Sinatra::Base
     erb :index
   end
 
+  post '/names' do
+    $player_name = params[:name]
+    redirect '/play'
+  end
+
+  get '/play' do
+    erb :play
+  end
+
   run! if app_file == $0
 end
--- a/app.rb
+++ b/app.rb
@@ -19,5 +19,14 @@ class Online_Game < Sinatra::Base
     erb :play
   end
 
+  post '/choice' do
+    $player_choice = params[:choice]
+    redirect '/result'
+  end
+
+  get '/result' do
+    erb :result
+  end
+
   run! if app_file == $0
 end
--- a/app.rb
+++ b/app.rb
@@ -7,7 +7,7 @@ class Online_Game < Sinatra::Base
   end
 
   get '/' do
-    'Hello World'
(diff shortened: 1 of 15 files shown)


spec/features/game_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    expect(game).to be_instance_of Game
+  end
+
+  describe '# random_choice' do 
+    it 'randomly selects an action from rock, paper or scissors' do
+      expect(game.random_choice).to eq('rock').or eq('paper').or eq('scissors')
+    end
+  end
+
+  describe '# return_winner' do
```
lukestorey95, line 17: Add tests for all possible outcomes, as only three are tested here

spec/features/game_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'game'
+
+describe Game do
+
+  subject(:game) { Game.new('rock') }
+
+  it 'creates instances of game' do
```
lukestorey95, line 7: `describe Game` tests that there is a Game class and that it can be initialised so don't think you need this test as well

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,15 @@
+def sign_in_and_submit_name
+  visit '/'
+  fill_in :name, with: 'Jimmy'
+  click_button 'Submit'
+end
+
+def has_expected_output
+  outputs = ['Computer Wins!', 'Player Wins!', 'Draw']
```
lukestorey95, line 8: Nice use of an array for outputs rather than expecting on multiple individual strings - I hadn't though to do that

app.rb
```diff
@@ -0,0 +1,37 @@
...
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/names' do
+    $player_name = params[:name]
```
lukestorey95, line 16: Global variables are apparently evil - player should be split out into separate class to be injected into Game

lib/game.rb
```diff
@@ -0,0 +1,26 @@
+class Game
+  
+  attr_reader :computer_action
+
+  def initialize(choice)
+    @player_action = choice
+    @pairs = { rock: 'paper', paper: 'scissors', scissors: 'rock' }
```
lukestorey95, line 7: Think this is for the return_winner method, so why is it stored as an instance variable for the class?

lib/game.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def random_choice
+    ['rock', 'paper', 'scissors'].sample
```
lukestorey95, line 11: Could store this as a constant as the three options won't change. Also consider using symbols as they are immutable and are an identifier of an action rather than a typed user input.

views/play.erb
```diff
@@ -0,0 +1,11 @@
+<h1>Pick an action <%=$player_name%>:</h1>
+
+<form action='/choice' method='post'>
+  <input type="radio" id="rock" name="choice" value="Rock">
```
lukestorey95, line 4: Why set value to 'Rock' when you then use value.downcase in the app to convert to lowercase? Could refactor that out

app.rb
```diff
@@ -0,0 +1,37 @@
...
+    redirect '/play'
+  end
+
+  get '/play' do
+    erb :play
+  end
+
+  post '/choice' do
+    $player_choice = params[:choice].downcase
```
lukestorey95, line 25: You're using a button input in the form so the choice shouldn't= need to use .downcase

app.rb
```diff
@@ -0,0 +1,37 @@
...
+    redirect '/result'
+  end
+
+  get '/result' do
+    game = Game.new($player_choice)
```
lukestorey95, line 30: Consider initialising the Game when the choice is created so that you can store the choice within a player and within the Game and avoid the evil global variables

lukestorey95 (commented):
A good starting point, but I think if you wanted to move onto some more complex functionality it might be hard considering the current use of a single class model, and global variables. With some refactoring and splitting the Game class that would leave the code in much better condition, and make it easier to add multiplayer functionality etc. 

I have some thoughts on the use of symbols vs strings, but really liked your use of output array in the spec to test multiple outcomes in one test - very neat

PR #2125, opened 2022-06-04
RPS Challenge

joe-mcgarry:
My attempt at the RPS challenge, with a very simple implementation of multiplayer functionality

### DIFF
--- a/Views/result.erb
+++ b/Views/result.erb
@@ -1,2 +1,18 @@
-<%= @game.player_1.name %> chose <%= @game.get_player_choice %><br>
-AI chose <%= @game.comp_choice %>
+<h3><%= @game.player_1.name %> chose <%= @game.get_player_choice %></h3>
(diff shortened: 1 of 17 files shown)


Views/result.erb
```diff
@@ -0,0 +1,18 @@
+<h3><%= @game.player_1.name %> chose <%= @game.get_player_1_choice %></h3>
+<h3><%= @game.player_2.name %> chose <%= @game.get_player_2_choice %></h3>
+
+<% if @game.get_player_1_choice == @game.get_player_2_choice %>
```
Aisha-Yusuff, line 4: This logic in the view could be moved to a class. So, the view doesn't have to do too much work.

lib/game.rb
```diff
@@ -0,0 +1,29 @@
+class Game
```
Aisha-Yusuff, line 1: I really like this! It makes your code easy to follow and keeps the app.rb concise!

PR #2127, opened 2022-06-05
Weekend challenge

michaelcychan:
Completed the basic and multi player functions.


.gitignore
```diff
@@ -3,3 +3,5 @@
 
 # Local cache of Rubocop remote config
 .rubocop-*
+
+capybara-*.html
```
penguat, line 7: I love this, you've saved yourself some work here!

app.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  post '/game' do
+    multi = Multi.new(params[:player_1_name], params[:player_2_name], params[:players_num].to_i)
+    $game = multi.game_creation
+    @game = $game
+    erb(:game)
+  end
+
+  get '/player_2' do
```
penguat, line 28: Could this be broken up? It feels like a lot of logic in the controller here.

Maybe each condition body (between the if/else and end) could be its own method?

spec/show_results_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+require 'game'
+
+describe Game do
+  subject(:object) { described_class }
+  let(:player1) { double :player }
+  let(:player2) { double :player }
+
+  describe '#match' do
+    it 'declares Player 1 wins' do
+      expect(player1).to receive(:choice).and_return(:rock).at_most(3).times
```
penguat, line 11: This is a lovely clear test

penguat (commented):
Excellent work, well done!

.gitignore
abacus2020: Great idea

config.ru
```diff
@@ -0,0 +1,5 @@
+# ./config.ru
+
+require_relative "./app"
+
+run RPS
```
abacus2020, line 8: Minor issue: Add a code syntax checker to your text editor to check syntax issues

app.rb
abacus2020: I agree; this looks like sound logic, but I would suggest separating the two conditions within the IF statement and considering separating the redirect and view statements into separate routes.

lib/player.rb
```diff
@@ -0,0 +1,22 @@
+# lib/player.rb
+
+class Player
+  attr_reader :name
+  attr_accessor :choice
```
abacus2020, line 5: Great use of attr!

lib/player.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def initialize(name)
+    @name = name
+    @choice = nil
+  end
+
+  def throw
+    @choice = [:rock, :scissors, :paper].sample
```
abacus2020, line 13: Point of consideration: Your implementation works, which is excellent and the most important thing; however, you are not using the symbol object here. I would suggest sticking to string objects within an array for this challenge.

lib/multi.rb
```diff
@@ -0,0 +1,30 @@
+# lib/multi.rb
```
abacus2020: Minor point: I suggest renaming this file to be more self explanatory

lib/multi.rb
```diff
@@ -0,0 +1,30 @@
...
+class Multi
+  attr_reader :players_num
+  
+  def initialize(player_1_name, player_2_name, players_num)
+    @players_num = players_num
+    @player_1_name = player_1_name
+    @player_2_name = player_2_name
+  end
+
+  def set_players
```
abacus2020: If you include single line comments before each method. Your implementation is not trivial, and so the complexity needs to be better explained.

lib/game.rb
```diff
@@ -0,0 +1,83 @@
+# lib/game.rb
+
+require './lib/player'
+
+class Game
+  attr_reader :players, :declaration, :players_num, :current_player, :other_player, :turn
+
+  def initialize(player1, player2, players_num)
+    @winmap = {
```
abacus2020, line 9: I like this use of cases to simplify the winning condition

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def player_1
+    @players[0]
+  end
+
+  def player_2
+    @players[1]
+  end
+
```
abacus2020, line 28: winner, player_1 and player_2 can be refactored into `attr` objects

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+    }
+    @players = [player1, player2]
+    @winner = nil
+
+    @turn = 1
+    @current_player = @players[0]
+    @other_player = @players[1]
+
+    @declaration = ""
+    @players_num = players_num
```
abacus2020, line 20: I think this construction could be refactored and simplified.

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def declaration
+    @declaration
```
abacus2020: not necessary as you have attr'd it above

lib/game.rb
```diff
@@ -0,0 +1,83 @@
...
+    elsif @winmap[player_1.choice] == player_2.choice
+      @winner = player_1
+    else
+      @winner = player_2
+    end
+
+    if @winner == nil
+      @declaration = "It is a tie!"
+    elsif players_num == 2
+      @declaration = "#{@winner.name} wins!"
```
abacus2020: I don't think the declaration variable is needed, this can be inputted directly into the view

views/index.erb
```diff
@@ -0,0 +1,22 @@
+<!-- views/index.erb -->
+
+<h1>Welcome to the Ultimate Rock Paper Scissors Game!</h1>
+
+<div class="PlayerDetails">
+  <p>Before we start...</p>
+  <form action="/game" method="post">
+    <p>Please choose number of players:</p>
+    <p><input type="radio" id="1" name="players_num" value="1">
```
abacus2020, line 9: Great use of radio buttons

PR #2128, opened 2022-06-06
Aisha's RPS Challenge

Aisha-Yusuff:
Completed the first group of user stories, need to refactor code and add to README.md

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,28 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+
+class RockPaperScissors < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  enable :sessions
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    session[:username] = params[:names]
+    redirect '/play'
+  end
+
+  get '/play' do 
+    @username = session[:username]
+    @game = $game
+    erb :play
+  end
+
+
+  run! if app_file == $0
+end
--- a/app.rb
+++ b/app.rb
@@ -1,7 +1,8 @@
 require 'sinatra/base'
 require 'sinatra/reloader'
 require './lib/player'
-
+require './lib/computer'
+require './lib/final_result'
 
 class RockPaperScissors < Sinatra::Base
   configure :development do
@@ -41,7 +42,13 @@ class RockPaperScissors < Sinatra::Base
   end
 
   get '/game' do 
+    @username = session[:username]
+    @player_option = session[:player].option
+    session[:computer] = Computer.new(["Rock", "Paper", "Scissors"])
+    @computer_option = session[:computer].option
+
     
+    erb :game
   end 
 
   run! if app_file == $0
--- a/app.rb
+++ b/app.rb
@@ -1,5 +1,7 @@
 require 'sinatra/base'
 require 'sinatra/reloader'
+require './lib/player'
+
 
 class RockPaperScissors < Sinatra::Base
   configure :development do
@@ -23,6 +25,24 @@ class RockPaperScissors < Sinatra::Base
     erb :play
   end
 
+  post '/rock' do
+    session[:player] = Player.new(params[:player_option])
+    redirect '/game'
+  end 
+
+  post '/paper' do
+    session[:player] = Player.new(params[:player_option])
+    redirect '/game'
+  end
+
+  post '/scissors' do 
+    session[:player] = Player.new(params[:player_option])
+    redirect '/game'
+  end
+
+  get '/game' do 
+    
+  end 
 
   run! if app_file == $0
 end
--- a/app.rb
+++ b/app.rb
@@ -47,7 +47,9 @@ class RockPaperScissors < Sinatra::Base
     session[:computer] = Computer.new(["Rock", "Paper", "Scissors"])
     @computer_option = session[:computer].option
 
-    
+    $final_result = FinalResult.new
+    $final_result.calculate(@player_option, @computer_option)
+    @final_result = $final_result.winner
     erb :game
   end 
 
--- a/lib/final_result.rb
+++ b/lib/final_result.rb
@@ -1,0 +1,24 @@
+class FinalResult
+
+  def initialize
+    @rules = {
+      Rock: { Rock: :draw, Paper: :lose, Scissors: :win },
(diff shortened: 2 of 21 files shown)


spec/features/display_final_result_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      allow_any_instance_of(Array).to receive(:sample).and_return "Scissors" 
+      sign_in_and_play
+      click_button 'Rock'
+      expect(page).to have_content "Congratulations! You won, you beat the computer!"
+    end 
+  end
+    
+
+  feature "player has lost against the computer" do 
+    scenario "When player plays paper" do
```
joe-mcgarry, line 14: Tests look good and comprehensive, nice one!

lib/final_result.rb
```diff
@@ -0,0 +1,24 @@
+class FinalResult
+
+  def initialize
+    @rules = {
+      Rock: { Rock: :draw, Paper: :lose, Scissors: :win },
+      Paper: { Rock: :win, Paper: :draw, Scissors: :lose },
```
joe-mcgarry, line 6: I really like this way of going about determining the rules of the game

app.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  post '/name' do
+    session[:username] = params[:names]
+    redirect '/play'
+  end
+
+  get '/play' do 
+    @username = session[:username]
+    @game = $game
```
joe-mcgarry, line 25: Is $game variable used or needed?

app.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  get '/game' do 
+    @username = session[:username]
+    @player_option = session[:player].option
+    session[:computer] = Computer.new(["Rock", "Paper", "Scissors"])
+    @computer_option = session[:computer].option
+
+    $final_result = FinalResult.new
+    $final_result.calculate(@player_option, @computer_option)
+    @final_result = $final_result.winner
```
joe-mcgarry, line 52: Could the @final_result variable go into a separate get route and have /game redirect to it (either as a post or get)? The global $final_result can stay in /game if needed?

lib/final_result.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def calculate(player_option, comp_option)
+    player = player_option.to_sym
+    comp = comp_option.to_sym
+    @result = @rules[player][comp]
+  end 
+
+  def winner
+    return "Congratulations! You won, you beat the computer!" if @result == :win
```
leoht, line 20: Rather than having human-readable messages returned by this class, it might be better to let this responsibility to the view (in the .erb file) to have all the content / text displayed to the user. This way, it becomes easier to change it, if we want to change the text displayed, translate it in other languages, etc. The method `winner` could then return either the result directly (`:win`, `:lose`, etc), and the the message to display would be part of the view layer

app.rb
```diff
@@ -0,0 +1,57 @@
...
+    redirect '/game'
+  end
+
+  get '/game' do 
+    @username = session[:username]
+    @player_option = session[:player].option
+    session[:computer] = Computer.new(["Rock", "Paper", "Scissors"])
+    @computer_option = session[:computer].option
+
+    $final_result = FinalResult.new
```
leoht, line 50: If the object `FinalResult.new` doesn't need to be accessed in a different route (which seems to be the case, as it's used only to calculate the result and chose the winer), it's probably good to avoid making it a global variable (it's generally best to avoid using global variables unless there is a need to):

```ruby
final_result = FinalResult.new
final_result.calculate(@player_option, @computer_option)
@final_result = $final_result.winner
```

Maybe the name of the class could be changed to `FinalResultCalculator` as well, to avoid mixing up that object with the result value itself?

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,5 @@
+def sign_in_and_play
+  visit('/')
```
leoht, line 2: Nice use of a helper method here - a way to make it better could be to pass the player name as an argument, this way it becomes more explicit, when used in tests, what name is going to be used:

```ruby
sign_in_and_play('Aisha')
click_button 'Rock'
expect(page).to have_content "Aisha played Rock"
```

views/play.erb
```diff
@@ -0,0 +1,36 @@
+<html>
+<head>
+</head>
+<style>
```
leoht, line 4: It's usually best to indent the HTML code, so it's easier to read - this becomes really useful when dealing with larger and more complex HTML pages! You can use a VS code extension (or similar) to auto indent the code when editing HTML files

spec/features/display_computer_option_spec.rb
```diff
@@ -0,0 +1,22 @@
+feature "display computer's option" do
+  scenario "when computer plays rock" do 
+    allow_any_instance_of(Array).to receive(:sample).and_return 'Rock'
```
leoht, line 3: Rather than stubbing the return of `Array.sample` (which could be dangerous, as this would modify the behaviour of the Array's `sample` method for _every_ single array used in your programs, even libraries, etc) - prefer stubbing a method of your own class you have more control of. For example, you could create a new method in `Computer`, called `get_random_move`, which would wrap the call to `.sample` - and then stub this method's return value

spec/final_result_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+  subject (:final_result) { FinalResult.new }
+
+  # Player = Rock
+  describe "when the player plays rock" do
+    subject (:player) { Player.new("Rock") }
+
+    context "and has won the game" do
+      it "it returns winning message" do 
+        comp = Computer.new(["Rock", "Paper", "Scissors"])
+        allow(comp).to receive(:option).and_return("Scissors")
```
leoht, line 15: Here, because you're passing directly the value of `option` to the method `.calculate`, it's not needed to stub the return value of `comp.option`, because you could directly give the plain string value to `.calculate`:

```ruby
final_result.calculate(player.option, "Scissors")
```

Stubbing would be useful in the case we'd inject the whole instance `comp` into that class, and then the method would be called by that class later on

leoht (commented):
Really nice work - I left a few comments here and there to improve mainly the testing and stubbing side of things, and some of the code design in a few places. Let me know if you have any questions on the feedback!

PR #2132, opened 2022-06-06
RPS Challenge

abacus2020:


app.rb
```diff
@@ -0,0 +1,37 @@
...
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/name' do
+    session[:namegiven] = params[:username]
+    redirect '/play'
```
michaelcychan, line 16: Reasonable to see the separation of retrieving user input from displaying a new page.
I think it's a good thing.

views/play.erb
```diff
@@ -0,0 +1,10 @@
+<h1>Welcome <%= @name %> to Rock, Paper, Scissors <%= @name_1 %></h1>
```
michaelcychan, line 1: @name_1 probably needs to be deleted

views/result.erb
```diff
@@ -0,0 +1,9 @@
+<h1>You played <%=@tool_selected%></h1>
+<h1>The computer played <%= @comp_selected %></h1>
+<%if @tool_selected == @comp_selected %>
+<h1>It is a draw</h1>
+<%elsif @tool_selected == "paper" && @comp_selected == "rock" || @tool_selected == "rock" && @comp_selected == "scissors" || @tool_selected == "scissors" && @comp_selected == "paper" %>
```
michaelcychan, line 5: I would use a hash in winning conditions, because it's easier to edit later

app.rb
```diff
@@ -0,0 +1,37 @@
+require 'sinatra/base'
```
michaelcychan, line 1: Adding test cases would be easier for testing multiple times.

views/result.erb
```diff
@@ -0,0 +1,9 @@
+<h1>You played <%=@tool_selected%></h1>
+<h1>The computer played <%= @comp_selected %></h1>
```
michaelcychan, line 2: Very nice to show both choices: user and computer. Good user experience, also good for testing.

PR #2133, opened 2022-06-06
Got up to here on weekend challenge: RPS

BenjaminNeustadt:


spec/features/player_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'player'
+
+describe Player do
+  subject(:player) { Player.new('Player') }
+  subject(:computer) { Player.new('Computer') }
+
+  describe '#name' do
```
raphaella-rose: This is a good initial test for the player class

views/name.erb
```diff
@@ -0,0 +1,6 @@
+We will now ask you for your name.
```
raphaella-rose: This is straight forward and easy to understand :)

controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    erb :name
+  end
+
+  get '/names' do
+    'Hello #{@player}'
+    redirect :play
+  end
+
+  post '/assign_name' do
+    @player = Player.new(params[:name])
```
raphaella-rose: You could use a session to store the player name

raphaella-rose (commented):
Using more tests to drive your code may help you know which direction to go in. I thought your code was easy to read. :)

PR #2136, opened 2022-06-06
Sophie gilder RPS

sophiechannon:


app.rb
```diff
@@ -0,0 +1,62 @@
...
+  post "/names_multi" do
+    player_1 = Player.new(params[:name])
+    player_2 = Player.new(params[:name2])
+    $game = Game.new(player_1, player_2)
+    redirect "/play"
+  end
+  
+  get "/play" do
+    @game = $game
+    erb @game.select_type
```
leoht, line 40: Usually it's best to enforce the view name to be something specific, rather than the return of a method that you might not fully control — for example, tomorrow you (or someone else) might modify `select_type` to return a different value, but might also forget to add a new view to cover this case here, which would break the program. Rather than coupling tightly the return value to the view name, it might be better to add some condition and specifically select the correct view — or better, to have a generic single view that is always used, but using an instance variable and some dynamic ERB tags to display a slightly different portion of the HTML, depending on the value

app.rb
```diff
@@ -0,0 +1,62 @@
...
+  post "/weapon_multi" do
+    @game = $game
+    @game.player_1.weapon = params[:weapon]
+    @game.player_2.weapon = params[:weapon2]
+    redirect '/result'
+  end
+  
+  get "/result" do
+    @game = $game
+    erb @game.result
```
leoht, line 58: Same remark than above here — you could set the result in an instance variable and display something different in the view depending on this

docs/rsp_plan.md
```diff
@@ -0,0 +1,158 @@
...
+So that I can see my name in lights
+I would like to register my name before playing an online game
+
+As a marketeer
+So that I can enjoy myself away from the daily grind
+I would like to be able to play rock/paper/scissors
+
+### Routes
+
+````ruby
```
leoht, line 13: It's good you've designed the routes beforehand here — it probably helped a lot your decisions in terms of routing, and how the application would behave, HTTP-wise

lib/computer.rb
```diff
@@ -0,0 +1,8 @@
+class Computer
+  attr_reader :name, :weapon, :type
+
+  def initialize
+    @weapon = Game::WEAPONS.sample
+    @name = :Computer
```
leoht, line 6: Minor one - usually symbol names are always lowercase

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,19 @@
+def enter_and_submit_name_single
```
leoht, line 1: Nice use of helper methods here — to improve it, something good could be to add arguments (e.g to pass the player name) so, when calling the helper in your tests, the names the tests should expect become more explicit:

```ruby
# Example

enter_and_submit_name_single('Rosie')
click_button "Scissors"
expect(page).to have_text "Rosie selected Scissors"
# ...
```

spec/game_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    it "returns weapons constant" do
+      expect(Game::WEAPONS).to eq [:rock, :paper, :scissors]
+    end
+
+    it "returns rules constant" do
+      expect(Game::RULES).to eq ({rock: :scissors, scissors: :paper, paper: :rock})
+    end
+
+    it "returns true for single player" do
```
leoht, line 22: It might be good to have a test as well that checks it returns `false` when it is _not_ a single player

leoht (commented):
Great work, well done! Sorry for the delay in reviewing this - not much to add aside a few small suggestions here and there. Feel free to ask me if you have more questions on that feedback :)

PR #2137, opened 2022-06-06
rps challenge - raphaella

raphaella-rose:


spec/features/single_player_game_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'single player game' do
+  xscenario 'player selects rock' do
+    visit('/')
+    fill_in :player_name, with: 'Raphaella'
```
BenjaminNeustadt, line 4: this test is comprehensive, and understandable. I just wonder whether you could use the web_helper 'sign in and play' here instead of fill_in[...]?

views/play.erb
```diff
@@ -1 +1,4 @@
-<div>Player: <%= @player_name %></div>
+<div>Player: <%= @player.name %></div>
+<form action='/weapon_selection' method='post'>
+<input type='submit' value='Rock' name='weapon'>
```
BenjaminNeustadt, line 3: this is really nice, I am going to try implement something in my own code.

BenjaminNeustadt (commented):
The tests are good, I personally like to write the controller first as it feels easier that way to know whether the get requests and post requests are working correctly. Very nice work

